PR #33, opened 2023-02-14
Switch pip-tools/pip_compile function to Pipfile/pipenv for requirements construction

atheo89:
This PR refactors the notebooks repository to use `Pipfile` to construct the `Pipfile.lock` instead of the `pip_compile` function (pip-tools). 

Now, the dev has the flexibility to create, refresh or change the python packages of a notebook by running 
 `cd {NOTEBOOK_PIPFILE_DIR} && pipenv lock`

Upstream Issue: https://github.com/opendatahub-io/notebooks/issues/34

## Files that changed:

- Makefile (Deleted the pip_compile function and all its references inside the code)
- Refactored all the Dockerfiles to install python dependencies from Pipfile.lock instead requirments.txt
- Created Pipfiles and Pipfile.locks for all notebooks  
- Deleted requirements.in & requirements.txt files
- Refreshed contributing.md file with pipenv instructions

## How Has This Been Tested?

1. To build/push an image you should run the corresponding:
 `make ${NOTEBOOK_NAME} -e IMAGE_REGISTRY=quay.io/${YOUR_USER}/notebooks` 
For example `make jupyter-minimal-ubi9-python-3.9 -e IMAGE_REGISTRY=quay.io/${YOUR_USER}/notebooks`
2. Run the image locally `podman run --rm -it -p8888:8888 quay.io/${YOUR_USER}/notebooks:{NOTEBOOK_NAME}`

## Merge criteria:


jupyter/datascience/ubi8-python-3.8/Pipfile
```diff
@@ -0,0 +1,38 @@
...
+matplotlib = "~=3.6.3"
+numpy = "~=1.24.1"
+pandas = "~=1.5.3"
+plotly = "~=5.13.0"
+scikit-learn = "~=1.2.1"
+scipy = "~=1.10.0"
+jupyterlab-lsp = "~=3.10.2"
+jupyterlab-widgets = "~=3.0.5"
+jupyter-resource-usage = "~=0.6.0"
+jupyterlab-s3-browser = "~=0.10.1"
```
harshad16, line 23: I see lot of packages are updated in here with respect to requirements.in 
however some packages are downgraded, like jupyterlab-s3-browser from 0.12.0 to 0.10.x 
any particular reason for this? or was it due to dependency resolution. ?

harshad16 (commented):
Great work :+1: 
some question regarding changes

jupyter/datascience/ubi8-python-3.8/Pipfile
atheo89: Exactly, the downgrade on jupyterlab-s3-browser package happened due to dependency resolution.

base/ubi9-python-3.9/Pipfile
```diff
@@ -0,0 +1,13 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+thamos = "~=1.29.1"
+wheel = "~=0.38.4"
```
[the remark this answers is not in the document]
atheo89: Great notice! The criterion for choosing the wheel package here was to have the current latest version. The ubi base images (ubi8,ubi9) that we are using have the latest tag. Do you think that may this cause any incompatibilities? Also, if we define here the pip version can we overwrite one on the base images?

jupyter/datascience/ubi8-python-3.8/Pipfile
guimou: At some point we should question the inclusion of jupyterlab-s3-browser. It has not been updated since July 2022, the main contributor seems to have switched to something else.  I fear that at some point it will slow/block us, requiring packages that are totally obsolete. @atheo89 solved the issue by downgrading a lot, which removed bad requirements,  I just got rid of it in my contrib images.

base/ubi9-python-3.9/Pipfile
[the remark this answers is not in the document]
harshad16: With layer, any package version of the base image can be overwritten.
I feel the wheel package should be discussed going forward for now it looks good to me.

jupyter/minimal/ubi8-python-3.8/Pipfile
```diff
@@ -0,0 +1,26 @@
...
+jupyter-server = "~=2.1.0"
+jupyter-server-proxy = "~=3.2.2"
+jupyter-server-terminals = "~=0.4.4"
+jupyterlab-git = "~=0.41.0"
+nbdime = "~=3.1.1"
+nbgitpuller = "~=1.1.1"
+
+# Parent image requirements to maintain cohesion
+thamos = "~=1.29.1"
+wheel = "~=0.38.4"
```
harshad16, line 23: in hindsight, 
if we are installing these bits here, we can remove the [base](./base) layer completely. 
Right? it would make one  less image.

harshad16 (changes requested):
on suggestion
/remove lgtm

jupyter/minimal/ubi8-python-3.8/Pipfile
guimou: Base may be required for the runtime images, to use in the pipelines. Those images will be identical to the ootb notebooks, but without all the Jupyter stuff. We cannot use directly the Jupyter images as runtimes because of some incompatibilities with the Elyra process, when its bootstrap parses the notebooks.

atheo89: Right, moreover we use the base to build the CUDA layer.

harshad16 (commented):
/lgtm
thanks :+1:

guimou:
> I pushed a last commit to include the `tf2onnx` package under TensorFlow

Great! It's funny because I just looked at the PR after remembering this morning that we wanted to include that to be able to work with model serving!

harshad16 (commented):
/lgtm 
:+1:

atheo89:
> > I pushed a last commit to include the `tf2onnx` package under TensorFlow
> 
> Great! It's funny because I just looked at the PR after remembering this morning that we wanted to include that to be able to work with model serving!

Yeah, I passed through the list  with candidate packages and I noticed it there

PR #44, opened 2023-02-28
Add piplock-renewal workflow to main

VaishnaviHire:
## Description
This PR adds the workflow to renew Pipfile.lock for `main` branch

## Merge criteria:


atheo89:
Doing this implementation on the main upstream branch means that the corresponding downstream main will have Pull Requests via the wei plugin. And consequently, the openshift-ci pipeline is going to build and push the new images every week, but we don't have to do this for the main branch but only for the release branches. 

If we want this workflow on main you should define on the openshift-ci to not trigger new builds.

In any case, the main branch will have the updated piplock file when a new image is going to be integrated or apply a new change.

PR #194, opened 2023-08-31
docs(readme): fix typo in image names

rkpattnaik780:
This PR fixes a typo in the README

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

Issue #195, opened 2023-09-01
The data science anaconda notebook fails to build on openshift CI


atheo89:
When attempting to build the Data Science Anaconda Notebook on Openshift CI as part of the test run, the build process encounters an error. 
The failure occurs in the Dockerfile at [this line](https://github.com/opendatahub-io/notebooks/blob/main/jupyter/datascience/anaconda-python-3.8/Dockerfile#L31). The error message received is "DockerBuildFailed: Dockerfile build strategy has failed."
That build error has been observed recently, on this link are the detailed logs [here](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/opendatahub-io_notebooks/188/pull-ci-opendatahub-io-notebooks-main-jupyter-datascience-anaconda-python-3-8-pr-image-mirror/1696795994106105856).

jstourac:
Just providing some data since the originally linked run isn't available anymore.

Reason for this issue are conflicting packages found during the following command:

```
/opt/anaconda3/bin/conda env update -p "${APP_ROOT}" --file environment.yml
```

```
STEP 11/19: RUN echo "Installing softwares and packages" &&     /opt/anaconda3/bin/conda env update -p "${APP_ROOT}" --file environment.yml &&     rm -f ./environment.yml &&     mkdir /opt/app-root/runtimes &&     cp utils/pipeline-flow.svg /opt/app-root/lib/python3.8/site-packages/elyra/static/icons/kubeflow.svg &&     sed -i "s/Kubeflow Pipelines/Data Science/g" /opt/app-root/lib/python3.8/site-packages/elyra/pipeline/runtime_type.py &&     sed -i "s/Kubeflow Pipelines/Data Science Pipelines/g" /opt/app-root/lib/python3.8/site-packages/elyra/metadata/schemas/kfp.json &&     sed -i "s/\"default\": \"Argo\",/\"default\": \"Tekton\",/g" /opt/app-root/lib/python3.8/site-packages/elyra/metadata/schemas/kfp.json &&     sed -i -e "s/Python.*/$(python --version)\",/" /opt/app-root/share/jupyter/kernels/python3/kernel.json &&     rm /opt/app-root/share/jupyter/metadata/runtime-images/*.json &&     chmod -R g+w /opt/app-root/lib/python3.8/site-packages &&     fix-permissions /opt/app-root -P
Installing softwares and packages
[...]

PR #199, opened 2023-09-01
Fix UBI version used on particular places with regards the Pyton version

jstourac:
This PR contains fixes to narrow down appropriate UBI version with the Python version in use.

## Description

Two of the changes are mainly trivial because it's part of the image labels only.

One thing is a little bit more important, I think, since it has a direct impact on the way how the `jupyter-pytorch-ubi8-python-3.8` image is built (dependency in Makefile used).

## How Has This Been Tested?

Not tested at all right, now - if you tell me what to do, I'm happy to do some testing.

## Merge criteria:

Note - this actually follows up on the changes in PR #194, @rkpattnaik780 JFYI so you don't duplicate effort eventually :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ cuda-jupyter-tensorflow-ubi8-python-3.8: cuda-jupyter-datascience-ubi8-python-3.
 
 # Build and push jupyter-pytorch-ubi8-python-3.8 image to the registry
 .PHONY: jupyter-pytorch-ubi8-python-3.8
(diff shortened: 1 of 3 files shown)


rkpattnaik780 (approved):
LGTM!

Nice catch @jstourac . I wasn't aware of these yet. 
To test it, I did a `make jupyter-pytorch-ubi8-python-3.8`. That should suffice.

jstourac:
Thank you @rkpattnaik780.

Looking at the results, I think that the failed tests are actually caused by the #195 issue.

harshad16 (approved):
/lgtm
/approve

Good catch
Thanks for fixing this :100:

PR #200, opened 2023-09-01
Use RHEL9 repository for UBI9 and CentOs Stream 9 Cuda images

jstourac:
This updates the used Nvidia repository for CUDA packages in our UBI9 and CoreOs 9 Stream images.

## Description

Nvidia provides RHEL9 repository of their packages. We use RHEL8 for all of our images at the moment (UBI8, UBI9, CoreOs 9 Stream). I think we should use RHEL9 repository for UBI9 and CoreOS 9 Stream, since I believe these are closer to the RHEL9.

I haven't checked content and difference of the Nvidia `rhel8` vs `rhel9` repositories yet, I can do it if you ask. It's possible that most of the packages there will be identical, not sure how they build.

Please, let me know whether this change is welcomed actually. Thank you.

## How Has This Been Tested?

!!!! This hasn't been tested and since this is quite a big change we should definitely test before we merge this !!!!

## Merge criteria:

### DIFF
--- a/cuda/c9s-python-3.9/cuda.repo-x86_64
+++ b/cuda/c9s-python-3.9/cuda.repo-x86_64
@@ -1,6 +1,6 @@
 [cuda]
 name=cuda
-baseurl=https://developer.download.nvidia.com/compute/cuda/repos/rhel8/x86_64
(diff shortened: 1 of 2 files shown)


harshad16 (approved):
/lgtm
/approve

Good catch 
Thanks :100:

PR #217, opened 2023-09-21
Include patch to version pipeline runs with numbers instead of timestamp

### LINKED ISSUE #206: Introduce pipeline version with numbers(v1,etc) instead of datastamp
Currently when pipelines are created using elyra from the Notebooks, each additional iteration of the pipeline run, and create a new pipeline version, the pipeline version names are defined in the following manner: `<pipeline-name>+<datastamp>`.

UX feels it would be more convenient for the users to have the pipeline version names defined with numbers instead of data stamps. like `<pipeline-name>+<number>` example: `test-v1`

**Acceptance criteria**

- [ ] Extend the pipeline version name to include the number in place of the timestamp.
https://github.com/elyra-ai/elyra/blob/3c27ada25a27b719529e88268bed38d135e40805/elyra/pipeline/kfp/processor_kfp.py#L312C58-L312C70
   - There is a [GET](https://www.kubeflow.org/docs/components/pipelines/v1/reference/api/kubeflow-pipeline-api-spec/#operation--apis-v1beta1-pipeline_versions-get) call for pipeline_version, to gather the count of the pipeline_version.

- [ ] Open the PR upstream on elyra, if it taking time to merge.
Open PR to patch it in the notebook for the time being. 
https://github.com/opendatahub-io/notebooks/blob/main/jupyter/datascience/ubi9-python-3.9/utils/processor_kfp.patch
https://github.com/opendatahub-io/notebooks/blob/57b82522d53b41252ecc53c8d2a7be9303ae7c2b/jupyter/datascience/ubi9-python-3.9/Dockerfile#L41C68-L41C71

rkpattnaik780:
Patch to enable elyra use version names for pipeline runs instead of timestamp to make it more convenient for user.

## Description
Currently, when pipelines are created using elyra from the Notebooks, each additional iteration of the pipeline run creates a new pipeline version identified by timestamp.

The PR will allow user to switch from timestamp identifiers to versions numbers (v1, etc) using an environment variable `KFP_SUFFIX_USE_VERSION`. This will make the pipeline runs easily distinguishable.

Fixes #206 

## Merge criteria:

### DIFF
--- a/jupyter/datascience/ubi9-python-3.9/utils/processor_kfp.patch
+++ b/jupyter/datascience/ubi9-python-3.9/utils/processor_kfp.patch
@@ -8,7 +8,25 @@
                  )
              else:
                  client = ArgoClient(
-@@ -416,7 +417,7 @@
+@@ -309,8 +310,16 @@
+ 
+                 # Create an instance id that will be used to store
+                 # the pipelines' dependencies, if applicable
+-                pipeline_instance_id = f"{pipeline_name}-{timestamp}"
++                pipeline_instance_id = ""
+ 
++                # Use environment variable to set version as suffix instead of timestamp
++                if os.environ.get("KFP_SUFFIX_USE_VERSION") == "true" :
++                    # Version is determined by the count of existing pipeline versions
++                    version = client.list_pipeline_versions(pipeline_id=pipeline_id).total_size
++                    pipeline_instance_id = f"{pipeline_name}-v{version}"
++                else :
++                    pipeline_instance_id = f"{pipeline_name}-{timestamp}"
++
+                 # Generate Python DSL from workflow
+                 pipeline_dsl = self._generate_pipeline_dsl(
+                     pipeline=pipeline,
+@@ -416,7 +425,7 @@
  
                  # create pipeline run (or specified pipeline version)
                  run = client.run_pipeline(
@@ -17,7 +35,7 @@
                  )
  
              except Exception as ex:
-@@ -435,7 +436,7 @@
+@@ -435,7 +444,7 @@
  
              self.log_pipeline_info(
                  pipeline_name,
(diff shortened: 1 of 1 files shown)


jupyter/datascience/ubi9-python-3.9/utils/processor_kfp.patch
```diff
@@ -8,7 +8,25 @@
...
+ 
+                 # Create an instance id that will be used to store
+                 # the pipelines' dependencies, if applicable
+-                pipeline_instance_id = f"{pipeline_name}-{timestamp}"
++                pipeline_instance_id = ""
+ 
++                # Use environment variable to set version as suffix instead of timestamp
++                if os.environ.get("KFP_SUFFIX_USE_VERSION") == "true" :
++                    # Version is determined by the count of existing pipeline versions
++                    version = client.list_pipeline_versions(pipeline_id=pipeline_id).total_size
```
harshad16, line 21: We need to check this pattern,
the scenario, which i would like to test is following:
- once we have few pipeline. ex: 10 , the version=10
- a user delete the 6th pipeline, the total_size=9 
- the next call of pipeline, would try to create the version=10 
again , perhaps. WDYT?

rkpattnaik780: I hadn't thought of deletion case, the current patch will fail to handle a deletion scenario. Let me search the docs a bit more and try to find a better approach.

harshad16:
/hold

Based on the sync with the Dashboard team,
seems like it would be best to talk with the UX team in detail about this behavior.
Hold till we have further instructions.

harshad16:
Thank you for working on this.
as the deletion case would cause the issue.
Closing the PR.
https://github.com/opendatahub-io/notebooks/issues/206#issuecomment-1756072052

Issue #34, opened 2023-02-14
Switch pip_compile functionality to Pipenv 


atheo89:
The pip_compile method on the Makefile uses pip-tools for the generation of the requirements.txt file introducing some limitations as below:

 -  Incompatibility with Elyra 
 -  Slow debug time

We should switch to Pipfile and Pipfile.lock

PR #226, opened 2023-09-27
Set $scheme correctly on vscode and rstudio on nginx configuration

atheo89:
Set $scheme correctly on vscode and rstudio on nginx configuration

## Description
In the connection flow, there is a redirect to an http session (before being sent back to https) that makes thing impossible to work on networks filtering http traffic (the first call, normally answered with a redirect, never makes it to OpenShift).

Redirections to other locations were made to http, whatever the original scheme.
Route would take care of redirecting back to https, but this did not work when other proxies between browser and OpenShift would filter out any http traffic.
The patches make all redirections follow the original scheme, http or https.

Original issue:[ #34](https://github.com/opendatahub-io-contrib/workbench-images/issues/34 ) 
Fixes: https://github.com/opendatahub-io/notebooks/issues/227

## Merge criteria:

### DIFF
--- a/codeserver/c9s-python-3.9/nginx/httpconf/http.conf
+++ b/codeserver/c9s-python-3.9/nginx/httpconf/http.conf
@@ -32,3 +32,8 @@ log_format json escape=json '[{'
     '"execution_state":"busy",'
     '"connections": 1'
     '}]';
+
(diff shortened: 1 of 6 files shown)


harshad16 (approved):
/lgtm
/approve

Thanks for bring this in.

PR #258, opened 2023-10-13
[tests] Allow to overwrite repository for testing script download

jstourac:
This unifies the path prefix from where the testing scripts are downloaded into one environment property that can be easily overriden - this may be useful during the debugging of the tests execution or implementation.

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ DATE 			 ?= $(shell date +'%Y%m%d')
 IMAGE_TAG		 ?= $(RELEASE)_$(DATE)
 KUBECTL_BIN      ?= bin/kubectl
 KUBECTL_VERSION  ?= v1.23.11
+NOTEBOOK_REPO_BRANCH_BASE ?= https://raw.githubusercontent.com/opendatahub-io/notebooks/main
 REQUIRED_RUNTIME_IMAGE_COMMANDS="curl python3"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,6 +5,7 @@ DATE 			 ?= $(shell date +'%Y%m%d')
 IMAGE_TAG		 ?= $(RELEASE)_$(DATE)
 KUBECTL_BIN      ?= bin/kubectl
 KUBECTL_VERSION  ?= v1.23.11
+NOTEBOOK_REPO_BRANCH_BASE="https://raw.githubusercontent.com/opendatahub-io/notebooks/main"
```
harshad16, line 8:
```suggestion
NOTEBOOK_REPO_BRANCH_BASE ?= "https://raw.githubusercontent.com/opendatahub-io/notebooks/main"
```

lets make it generic, so other can run the test from their fork , if they want.

harshad16 (commented):
lgtm
might need a small change
https://www.gnu.org/software/make/manual/html_node/Setting.html

Makefile
jstourac: Oh, yeah, good catch! I wrote it the `bash` way :upside_down_face: . But it worked as expected too :smile: 

Anyway, I'll update. Thank you :slightly_smiling_face:

atheo89 (approved):
Good catch Jan! Thank you :slightly_smiling_face:  
/lgtm

Issue #292, opened 2023-10-27
[Test] Check Dockerfile labels for images


jstourac:
Let's review and check the content of the labels in Dockerfiles for runtimes (maybe also others?) images here in this repository !and also in downstream repo! . Namely these labels should be checked for content on our branches:

* `authoritative-source-url`
* `io.openshift.build.commit.ref`
* `io.openshift.build.source-location`
* `io.openshift.build.image`

https://issues.redhat.com/browse/RHOAIENG-229

jiridanek:
This issue was written when OpenShift-CI was the build system. The four labels it asks to validate (`authoritative-source-url`, `io.openshift.build.commit.ref`, `io.openshift.build.source-location`, `io.openshift.build.image`) are injected by OpenShift-CI at build time — they don't appear in our Dockerfiles.

Production builds now come from Konflux, which injects its own set of labels (e.g., `vcs-ref` instead of `io.openshift.build.commit.id`). The existing test infrastructure already handles this:

- `tests/containers/params_env_validation_test.py` detects the build system and validates the appropriate labels (`vcs-ref` for Konflux, `io.openshift.build.commit.ref` for OpenShift-CI)
- `ci/expected-image-metadata.yaml` provides expected values for `name`, `commitref`, `size_mb`, etc.

This issue is effectively obsolete. Proposing to close as no longer applicable.

jiridanek:
@coderabbitai assess and provide recommendations

jiridanek:
Closure evidence: Jira [RHAIENG-5140](https://redhat.atlassian.net/browse/RHAIENG-5140) (https://redhat.atlassian.net/browse/RHAIENG-5140) is closed as Won't Do because the original OpenShift-CI label requirement is obsolete under Konflux; replacement validation exists where applicable. Closing on that basis.

[RHAIENG-5140]: https://redhat.atlassian.net/browse/RHAIENG-5140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #309, opened 2023-11-03
[Fix/enhancement] dockerimage labels check and attempted to deduplicate

jstourac:
Relates to: https://github.com/opendatahub-io/notebooks/issues/292

Docker image labels reviewed

## Description

This change tries to deduplicate some strings in our docker image labels with ability to propagate some common values directly from the Makefile so we don't need to change these values with every release branch or for every newly introduced docker image.

There were some minor fixes in the labels introduced too in this commit.

---

## Notable points for discussion:

1. I used value of the `RELEASE` variable for the build-ref at the moment - I understand that this may not be ideal. We may introduce separate value for this in Makefile or we can try to compute the value (either ref hash or branch name) dynamically in Makefile too. But in general - the value of the `RELEASE` should match what we've used so far.
2. Also, some labels (`authoritative-source-url` and `build commit ref` at least) could be propagated directly from Makefile - in this case I mean by using the `--label` parameter. There are two drawbacks I can think of:
   1. We could lost them in case of an individual build because this way these won't be enforced by Dockerfile itself
   2. In [this Dockerfile](https://github.com/opendatahub-io/notebooks/blob/main/jupyter/datascience/anaconda-python-3.8/Dockerfile) we have a slightly different naming for some labels which includes `s2i` infix - we would have to assure we don't change it wrongly.
      * Actually - do we want these `s2i` infixes?
3. I didn't touch the Anaconda Base Dockerfile - looks like the labels there should deserve review too.

## How Has This Been Tested?

I did a brief build of one image so far to check the idea behind the changes works. Will do more once the conversation here is finalized:

```
make codeserver-ubi9-python-3.9 -e IMAGE_REGISTRY=quay.io/jstourac/workbench-images
```

Result images can be seen in https://quay.io/repository/jstourac/workbench-images?tab=tags&tag=latest

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ DATE 			 ?= $(shell date +'%Y%m%d')
 IMAGE_TAG		 ?= $(RELEASE)_$(DATE)
 KUBECTL_BIN      ?= bin/kubectl
 KUBECTL_VERSION  ?= v1.23.11
+AUTHORITATIVE_SOURCE_URL ?= https://github.com/opendatahub-io/notebooks
 NOTEBOOK_REPO_BRANCH_BASE ?= https://raw.githubusercontent.com/opendatahub-io/notebooks/main
 REQUIRED_RUNTIME_IMAGE_COMMANDS="curl python3"
 REQUIRED_CODE_SERVER_IMAGE_COMMANDS="curl python oc code-server"
@@ -18,12 +19,13 @@ REQUIRED_R_STUDIO_IMAGE_COMMANDS="curl python oc /usr/lib/rstudio-server/bin/rse
 define build_image
 	$(eval IMAGE_NAME := $(IMAGE_REGISTRY):$(1)-$(IMAGE_TAG))
 	$(info # Building $(IMAGE_NAME) image...)
+	$(eval BUILD_ARGS := --build-arg BUILD_COMMIT_REF=$(RELEASE) --build-arg IMAGE_REGISTRY=$(IMAGE_REGISTRY) \
+		--build-arg AUTHORITATIVE_SOURCE_URL=$(AUTHORITATIVE_SOURCE_URL))
 	$(if $(3),
 		$(eval BASE_IMAGE_NAME := $(IMAGE_REGISTRY):$(3)-$(IMAGE_TAG))
-		$(eval BUILD_ARGS := --build-arg BASE_IMAGE=$(BASE_IMAGE_NAME)),
-		$(eval BUILD_ARGS :=)
+		$(eval BUILD_ARGS := $(BUILD_ARGS) --build-arg BASE_IMAGE=$(BASE_IMAGE_NAME)),
 	)
-	$(CONTAINER_ENGINE) build --no-cache  -t $(IMAGE_NAME) $(BUILD_ARGS) $(2)
+	$(CONTAINER_ENGINE) build --no-cache --tag $(IMAGE_NAME) $(BUILD_ARGS) $(2)
 endef
 
 # Push function for the notebok image:
--- a/base/c9s-python-3.9/Dockerfile
+++ b/base/c9s-python-3.9/Dockerfile
@@ -1,14 +1,18 @@
 FROM quay.io/sclorg/python-39-c9s:c9s
 
+ARG AUTHORITATIVE_SOURCE_URL
+ARG BUILD_COMMIT_REF
+ARG IMAGE_REGISTRY
+
 LABEL name="odh-notebook-base-centos-stream9-python-3.9" \
       summary="Python 3.9 CentOS Stream 9 base image for ODH notebooks" \
       description="Base Python 3.9 builder image based on CentOS Stream 9 for ODH notebooks" \
       io.k8s.display-name="Python 3.9 c9s base image for ODH notebooks" \
       io.k8s.description="Base Python 3.9 builder image based on C9S for ODH notebooks" \
-      authoritative-source-url="https://github.com/opendatahub-io/notebooks" \
-      io.openshift.build.commit.ref="main" \
-      io.openshift.build.source-location="https://github.com/opendatahub-io/notebooks/tree/main/base/c9s-python-3.9" \
-      io.openshift.build.image="quay.io/opendatahub/workbench-images:base-c9s-python-3.9"
+      authoritative-source-url="${AUTHORITATIVE_SOURCE_URL}" \
+      io.openshift.build.commit.ref="${BUILD_COMMIT_REF}" \
+      io.openshift.build.source-location="${AUTHORITATIVE_SOURCE_URL}/tree/${BUILD_COMMIT_REF}/base/c9s-python-3.9" \
+      io.openshift.build.image="${IMAGE_REGISTRY}:base-c9s-python-3.9"
 
 WORKDIR /opt/app-root/bin
 
--- a/base/ubi8-python-3.8/Dockerfile
+++ b/base/ubi8-python-3.8/Dockerfile
@@ -1,14 +1,18 @@
 FROM registry.access.redhat.com/ubi8/python-38:latest
 
+ARG AUTHORITATIVE_SOURCE_URL
+ARG BUILD_COMMIT_REF
+ARG IMAGE_REGISTRY
+
 LABEL name="odh-notebook-base-ubi8-python-3.8" \
       summary="Python 3.8 base image for ODH notebooks" \
       description="Base Python 3.8 builder image based on UBI8 for ODH notebooks" \
       io.k8s.display-name="Python 3.8 base image for ODH notebooks" \
       io.k8s.description="Base Python 3.8 builder image based on UBI8 for ODH notebooks" \
-      authoritative-source-url="https://github.com/opendatahub-io/notebooks" \
-      io.openshift.build.commit.ref="main" \
-      io.openshift.build.source-location="https://github.com/opendatahub-io/notebooks/tree/main/base/ubi8-python-3.8" \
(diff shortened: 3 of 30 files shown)


Makefile
```diff
@@ -18,12 +18,12 @@ REQUIRED_R_STUDIO_IMAGE_COMMANDS="curl python oc /usr/lib/rstudio-server/bin/rse
 define build_image
 	$(eval IMAGE_NAME := $(IMAGE_REGISTRY):$(1)-$(IMAGE_TAG))
 	$(info # Building $(IMAGE_NAME) image...)
+	$(eval BUILD_ARGS := --build-arg BUILD_COMMIT_REF=$(RELEASE) --build-arg IMAGE_REGISTRY=$(IMAGE_REGISTRY))
 	$(if $(3),
 		$(eval BASE_IMAGE_NAME := $(IMAGE_REGISTRY):$(3)-$(IMAGE_TAG))
-		$(eval BUILD_ARGS := --build-arg BASE_IMAGE=$(BASE_IMAGE_NAME)),
-		$(eval BUILD_ARGS :=)
```
atheo89, line 24: Why this line deleted?

base/c9s-python-3.9/Dockerfile
```diff
@@ -1,14 +1,19 @@
 FROM quay.io/sclorg/python-39-c9s:c9s
 
+ARG BUILD_COMMIT_REF
+ARG IMAGE_REGISTRY
+
+ARG AUTHORITATIVE_SOURCE_URL="https://github.com/opendatahub-io/notebooks"
```
atheo89, line 6: What if we move also this one outside from the Dockerfile? on the Makefile for example as we do with the other --build-args?

atheo89:
Hey Jan! This changes looks good to me  :slightly_smiling_face:  I added a small comment on the proposal!

base/c9s-python-3.9/Dockerfile
jstourac: yes, agree with you - I pointed out in the point 2 in the description in section `Notable points for discussion` - since you are also favor of it, I'll update my PR this way :slightly_smiling_face:

atheo89: Right! Tnx :)

jstourac: Changed per your proposal :slightly_smiling_face:  ; Also, I removed relevant point from the description since it's irrelevant now.

jstourac:
Okay, I rebased and updated with the change so that the `AUTHORITATIVE_SOURCE_URL` is now propagated also from `Makefile`. And I also used the `BUILD_COMMIT_REF` value in the `io.openshift.build.source-location` value.

jstourac:
Regarding the Anaconda - there are two `LABEL` definitions in the Dockerfile:

https://github.com/opendatahub-io/notebooks/blob/main/base/anaconda-python-3.8/Dockerfile#L25C1-L34C48
```
LABEL summary="$SUMMARY" \
      description="$DESCRIPTION" \
      io.k8s.description="$DESCRIPTION" \
      io.k8s.display-name="Anaconda Python 3.8" \
      io.openshift.expose-services="8080:http" \
      io.openshift.tags="anaconda,builder,python,python38,python-38,anaconda-python38" \
      name="anaconda-python38" \
      version="1" \
      usage="s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.8/test/setup-test-app/ ubi8/python-38 python-sample-app" \
      maintainer="spryor@redhat.com for now..."
```

and

https://github.com/opendatahub-io/notebooks/blob/main/base/anaconda-python-3.8/Dockerfile#L87C1-L91C98
```
LABEL name="odh-notebook-base-ubi8-anaconda-python-3.8" \
      summary="Anaconda based Python 3.8 base image for ODH notebooks" \
      description="Anaconda Base Python 3.8 builder image based on UBI8 for ODH notebooks" \
      io.k8s.display-name="Anaconda Python 3.8 base image for ODH notebooks" \
      io.k8s.description="Anaconda Base Python 3.8 builder image based on UBI8 for ODH notebooks"
```

Is there anything what we want to do with it? Do we want to update the second section to similar format as is e.g. in:

https://github.com/opendatahub-io/notebooks/blob/main/base/ubi9-python-3.9/Dockerfile#L3C1-L11C91
```
LABEL name="odh-notebook-base-ubi9-python-3.9" \
      summary="Python 3.9 base image for ODH notebooks" \
      description="Base Python 3.9 builder image based on UBI9 for ODH notebooks" \
      io.k8s.display-name="Python 3.9 base image for ODH notebooks" \
      io.k8s.description="Base Python 3.9 builder image based on UBI9 for ODH notebooks" \
      authoritative-source-url="https://github.com/opendatahub-io/notebooks" \
      io.openshift.build.commit.ref="main" \
      io.openshift.build.source-location="https://github.com/opendatahub-io/notebooks/tree/main/base/ubi9-python-3.9" \
      io.openshift.build.image="quay.io/opendatahub/workbench-images:base-ubi9-python-3.9"
```

harshad16 (commented):
initial thought which made me hesitate were that, we would need to do this separately in downstream, however as we pretty much follow the similar methodology.
This would indeed help with both streams.

@jstourac can you rebase and update the PR.

harshad16:
> Regarding the Anaconda - there are two `LABEL` definitions in the Dockerfile:
> ...
> Is there anything what we want to do with it? Do we want to update the second section to similar format as is e.g. in:
> 
> https://github.com/opendatahub-io/notebooks/blob/main/base/ubi9-python-3.9/Dockerfile#L3C1-L11C91
> 
> ```
> LABEL name="odh-notebook-base-ubi9-python-3.9" \
>       summary="Python 3.9 base image for ODH notebooks" \
>       description="Base Python 3.9 builder image based on UBI9 for ODH notebooks" \
>       io.k8s.display-name="Python 3.9 base image for ODH notebooks" \
>       io.k8s.description="Base Python 3.9 builder image based on UBI9 for ODH notebooks" \
>       authoritative-source-url="https://github.com/opendatahub-io/notebooks" \
>       io.openshift.build.commit.ref="main" \
>       io.openshift.build.source-location="https://github.com/opendatahub-io/notebooks/tree/main/base/ubi9-python-3.9" \
>       io.openshift.build.image="quay.io/opendatahub/workbench-images:base-ubi9-python-3.9"
> ```

Updating the 2nd part exclusively the jupyter/anaconda-x part , same as other would be prefered.

jstourac:
Thank you, @harshad16, for your review. I'll try to find some time for this this week :slightly_smiling_face:

base/anaconda-python-3.8/Dockerfile
```diff
@@ -88,7 +92,11 @@ LABEL name="odh-notebook-base-ubi8-anaconda-python-3.8" \
       summary="Anaconda based Python 3.8 base image for ODH notebooks" \
       description="Anaconda Base Python 3.8 builder image based on UBI8 for ODH notebooks" \
       io.k8s.display-name="Anaconda Python 3.8 base image for ODH notebooks" \
-      io.k8s.description="Anaconda Base Python 3.8 builder image based on UBI8 for ODH notebooks"
+      io.k8s.description="Anaconda Base Python 3.8 builder image based on UBI8 for ODH notebooks" \
+      authoritative-source-url="${AUTHORITATIVE_SOURCE_URL}" \
+      io.openshift.build.commit.ref="${BUILD_COMMIT_REF}" \
+      io.openshift.build.source-location="${AUTHORITATIVE_SOURCE_URL}/tree/${BUILD_COMMIT_REF}/base/anaconda-python-3.8" \
+      io.openshift.build.image="${IMAGE_REGISTRY}:base-c9s-python-3.9" TODO ?????
```
jstourac, line 99: I'm not quite sure what value should be put on this place. I wasn't able to find the image in the https://quay.io/repository/opendatahub/workbench-images - did I miss it somehow? :thinking:

harshad16: This particular is not built on the ODH, 
it is exclusively only built downstream.
we should probably clean a few of these.

jstourac: I see. So, do you prefer me to remove just this `io.openshift.build.image` label or all four in this case?

The first three are still valid and could be reused for the downstream part. Although, if there is no image build in upstream, I understand that it doesn't make much sense to have them here either.

harshad16: yeah true,
we can remove all of these:
```
 authoritative-source-url
 io.openshift.build.commit.ref
 io.openshift.build.source-location
 io.openshift.build.image
```
as this more for anyone , who would like to build it, though we wouldn't be building it in upstream.

jstourac: Thank you, Harshad! Updated :slightly_smiling_face:

harshad16:
The PR looks great, my major concern is this would break the current CI build from openshift CI 
and that doesn't completely rely on makefile. 
https://github.com/openshift/release/blob/358594a0d50341051cdd9eb1137757a734fab063/ci-operator/config/opendatahub-io/notebooks/opendatahub-io-notebooks-main.yaml#L8
the CI , doesn't take in build args, which make cause this issue.
Result image of this PR needs to be tested.

harshad16:
> The PR looks great, my major concern is this would break the current CI build from openshift CI and that doesn't completely rely on makefile. https://github.com/openshift/release/blob/358594a0d50341051cdd9eb1137757a734fab063/ci-operator/config/opendatahub-io/notebooks/opendatahub-io-notebooks-main.yaml#L8 the CI , doesn't take in build args, which make cause this issue. Result image of this PR needs to be tested.

Due to this, we are changing this PR to the draft.

Issue #336, opened 2023-11-14
 Add some extensions without the need to go and install them all over again on vscode workbench


atheo89:
**Why you need this feature:**

It would be nice to have on pre-installed extensions on vscode workbench

**Describe the solution you'd like:**

 Add some extensions without the need to go and install them all over again on vscode. Like Python/Jupyter/etc.

**Anything else you would like to add:**

There is a CLI on VS Code able to do this on a docker image. Ref [link](https://code.visualstudio.com/docs/editor/command-line#_working-with-extensions)

JIRA: https://issues.redhat.com/browse/RHOAIENG-231

harshad16:
This issue is completed with the PR #347 
Thank you for working on this.

PR #347, opened 2023-11-15
Pre-install some extensions on code-server notebook

atheo89:
Related to: https://github.com/opendatahub-io/notebooks/issues/336

## Description
By default, code-server does not come with any installed extensions. This pull request aims to streamline the user experience by pre-installing several useful extensions on behalf of the end user. The extensions that we added are:
- Jupyter (ms-toolsai.jupyter)
- python (ms-python.python)

Test image: `quay.io/rh_ee_atheodor/workbench-images@sha256:ac16959bffb660397247100666729b314eb78a8b0ec387131b878697d3cb1e48`

## How Has This Been Tested?
1. Spin up ODH 
2. Edit the `code-server-notebook` imagestream by adding the following tag, Builds > ImageStreams
```
    - name: 'new'
      annotations:
        opendatahub.io/notebook-python-dependencies: '[{"name":"code-sever","version":"4.11"}]'
        opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"}]'
        openshift.io/imported-from: quay.io/opendatahub/workbench-images
      from:
        kind: DockerImage
        name: >-
         quay.io/rh_ee_atheodor/workbench-images@sha256:ac16959bffb660397247100666729b314eb78a8b0ec387131b878697d3cb1e48
      importPolicy: {}
      referencePolicy:
        type: Source
```
3. Enter on DSP and trigger a vs code using the `new` tag that you created above.
4. Inspect on the extension tab, and check if there are already installed extensions.
You should see the following:
image

5. Or by a terminal should type: `code-server --list-extensions`
You should see the following:
image

6. To locally test the notebook in air-gap env disconnect your internet connection and trigger by using
`podman run --network=host --name validation-container quay.io/rh_ee_atheodor/workbench-images@sha256:ac16959bffb660397247100666729b314eb78a8b0ec387131b878697d3cb1e48`

## Merge criteria:


codeserver/ubi9-python-3.9/run-code-server.sh
```diff
@@ -16,6 +16,12 @@ fi
 # Initilize access logs for culling
 echo '[{"id":"code-server","name":"code-server","last_activity":"'$(date -Iseconds)'","execution_state":"running","connections":1}]' > /var/log/nginx/vscode.access.log
 
+# Install extensions if not present
+if [ ! -f "/opt/app-root/src/.local/share/code-server" ]; then
+   code-server --install-extension ms-python.python
+   code-server --install-extension ms-toolsai.jupyter
+fi
```
harshad16, line 23: Instead of doing this on run time, we should probably do this in build time.
as this would fail in disconnected environment.

```
RUN code-server --install-extension ms-python.python && \
   code-server --install-extension ms-toolsai.jupyter
```
in the dockerfile.
https://github.com/opendatahub-io/notebooks/blob/271208a7754c5ed29e02aeaac981fc80d68c08de/codeserver/ubi9-python-3.9/Dockerfile#L89

thoughts?

harshad16 (commented):
great work,
just had one suggestion.

codeserver/ubi9-python-3.9/run-code-server.sh
atheo89: I tried to install them on build run, but when you lunch the image on ODH, the extensions are not there, probably because the user's PVC is mounted at /opt/app-root/src, and it's "masked" and replaced by the volume that was mounted. Maybe, I should investigate the case to download the extension's files.

atheo89: Used .vsix packages for ms-python and ms-toolsai to be compatible with disconnected environments. I guess we will be safe now!

harshad16: Thanks for the work.
:100:

codeserver/ubi9-python-3.9/run-code-server.sh
```diff
@@ -16,6 +16,31 @@ fi
...
+
+        # Install specific extensions
+        code-server --install-extension ${SCRIPT_DIR}/utils/ms-python.python-2023.14.0.vsix
+        code-server --install-extension ${SCRIPT_DIR}/utils/ms-toolsai.jupyter-2023.3.100.vsix
+    else
+        # No internet connection
+        echo "No internet connection. Installing all extensions."
+
+        # Install all extensions
+        code-server --install-extension ${SCRIPT_DIR}/utils/ms-python.python-2023.14.0.vsix
```
shalberd, line 35: would `code-server --install-extension ${SCRIPT_DIR}/utils/*.vsix` also work?
For those users that modify the default set.
I am also thinking: would it be possible to put the list of binary extensions in a snippets folder location?

harshad16: Thanks for this suggestion, we can make it general.

harshad16 (approved):
/lgtm
/approve

thanks for the work :100:

codeserver/ubi9-python-3.9/run-code-server.sh
shalberd: good to know, thank you.

codeserver/ubi9-python-3.9/utils/ms-toolsai.jupyter-2023.3.100.vsix
shalberd, line 1: @atheo89 is there a particular reason you selected Version 2023.3.100 for the Jupyter extension? when I go to [marketplace visualstudio, I get 2024 versions](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter&ssr=false#version-history) and there are a bunch of changelogs for e.g. [February 2024 milestone](https://github.com/microsoft/vscode-jupyter/issues?page=2&q=is%3Aclosed+milestone%3A%22February+2024%22)

codeserver/ubi9-python-3.9/run-code-server.sh
```diff
@@ -16,6 +16,31 @@ fi
...
+        code-server --install-extension ${SCRIPT_DIR}/utils/ms-toolsai.jupyter-2023.3.100.vsix
```
shalberd, line 36: Hi: The documentation says: The Jupyter Extension will automatically install the following extensions by default to provide enhanced Jupyter notebook experiences in VS Code.
    [Jupyter Keymap](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter-keymap) - to provide Jupyter-consistent keymaps
    [Jupyter Notebook Renderers](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter-renderers) - to provide renderers for MIME types such as latex, plotly, vega, etc.
    [Jupyter Cell Tags](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.vscode-jupyter-cell-tags) and [Jupyter Slide Show](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.vscode-jupyter-slideshow) - to provide the ability to tag cells in notebooks and support for presentations
    
I guess this means that in ms.toolsai.jupyter extension vsix offline install mode, that still needs explicit one by one install of those extra 4 extensions like here, correct?

codeserver/ubi9-python-3.9/utils/ms-toolsai.jupyter-2023.3.100.vsix
atheo89: Hi @shalberd , the reason we specifically selected these versions of the extension is to ensure compatibility with code-server version 4.16.1.

codeserver/ubi9-python-3.9/run-code-server.sh
atheo89: Exactlym that's why we also provide them for use in disconnected environments [here](https://github.com/opendatahub-io/notebooks/blob/main/codeserver/ubi9-python-3.9/run-code-server.sh#L30)

codeserver/ubi9-python-3.9/utils/ms-toolsai.jupyter-2023.3.100.vsix
shalberd: Hi @atheo89  ok, from[ the vsx marketplace](https://open-vsx.org/extension/ms-toolsai/jupyter), or from the Github pages of the components, how can I see which code-server version or VS Code version the extension is for? I am looking for a kind of compatibility info. I do understand that
Codeserver 4.16.1 is VS Code 1.80.2, in essence.

codeserver/ubi9-python-3.9/run-code-server.sh
shalberd: yup, thank you for the clarification on what "automatically install" means in disconnected environments.

codeserver/ubi9-python-3.9/utils/ms-toolsai.jupyter-2023.3.100.vsix
atheo89: When installing an extension without specifying the version, code-server automatically selects the most compatible one. You can initially install them without specifying a version, then determine compatibility by running code --show-versions.

shalberd: @atheo89 found it, though with some research on stackoverflow, it is the package.json engine entry that shows the required version of code server / VS Code. https://github.com/microsoft/vscode-python/blob/main/package.json#L47 sorry for my long asking around, maybe this also helps others in future. Case closed.

atheo89: We downloaded the packages from vsx marketplace

shalberd: I am mostly focused on offline install, that is why I was looking around for this info so much.

shalberd: > We downloaded the packages from vsx marketplace

Yes, but very un-user-friendly, the marketplace does not show the required code server / VS Code version. One always has to go to the engines section of package.json ...

shalberd: @atheo89 @harshad16 there is a problem with this approach and current Notebooks when installing more than just these few extensions at runtime. The [defaults for liveness and readiness probe times come from odh-dashboard, I think, when assembling the notebook CR.](https://github.com/shalberd/odh-dashboard/blob/c0dca01fa32691e45cef37cb7c84471950e193c8/frontend/src/api/k8s/notebooks.ts#L150)


```
          name: code-server-vs
          livenessProbe:
            failureThreshold: 3
            httpGet:
              path: /notebook/dstst-analytix-fraud/code-server-vs/api
              port: notebook-port
              scheme: HTTP
            initialDelaySeconds: 10
            periodSeconds: 5
            successThreshold: 1
            timeoutSeconds: 1
```

Because installing extensions takes some time at the code-server-vs/api health check endpoint is not up and running until after all extensions install and the nginx http server is running, the container keeps re-starting without more than 5 or 6 extensions ever installed.

I solved this by patching the notebook CR to have a large initialDelaySeconds of 100, so then code server started.

What do you think?

guimou:
This is definitely an issue if the probes can't keep up. Several solutions:
- Can we start code-server and then only install the extensions?
- Right now the probe calls are forwarded by NGINX to code-server. We could
modify this to make NGINX answer directly. But it defeats the purpose of
the probes...

Honestly, I would not install any extensions. As they are persisted anyway,
I would let the users install what they want, when they want. They have to
do it only once, so... Plus the extensions installed now are related mostly
to Jupyter. Who says I want to use notebooks in VSCode? Personally I don't.
If I create a VSCode workbench, it's to do standard Python or JavaScript
coding, so...

Le lun. 19 févr. 2024 à 17:28, Sven Thoms ***@***.***> a
écrit :

shalberd:
about not installing any extensions and about the whole Jupyter stuff in general in VS Code:

yes, though some things like material theme and other GUI things like German language pack are quite useful actually.

about not installing any extensions and letting users install them:

that is a good solution in non-airgapped environments. But in environments like ours, with a proxy not allowing connections to marketplaces, pypi.org and so on, this command-based install of extensions by providing them in the file system is the only way to go. 

Here's some example ;-)

```
        code-server --install-extension ${SCRIPT_DIR}/utils/steoates.autoimport-1.5.3.vsix
        code-server --install-extension ${SCRIPT_DIR}/utils/njpwerner.autodocstring-0.6.1.vsix
        code-server --install-extension ${SCRIPT_DIR}/utils/aaron-bond.better-comments-3.0.2.vsix
        code-server --install-extension ${SCRIPT_DIR}/utils/alefragnani.Bookmarks-13.0.1.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/Catppuccin.catppuccin-vsc-3.10.1.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/formulahendry.code-runner-0.12.1.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/streetsidesoftware.code-spell-checker-3.0.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/streetsidesoftware.code-spell-checker-german-2.3.1.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/MS-CEINTL.vscode-language-pack-de-1.80.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/davidlday.languagetool-linter-0.20.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/PKief.material-icon-theme-4.33.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/ionutvmi.path-autocomplete-1.25.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/alefragnani.project-manager-12.0.1.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/KevinRose.vsc-python-indent-1.18.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/donjayamanne.python-environment-manager-1.0.4.vsix
        code-server --install-extension ${SCRIPT_DIR}/utils/hbenl.test-adapter-converter-0.1.6.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/hbenl.vscode-test-explorer-2.21.1.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/LittleFoxTeam.vscode-python-test-adapter-0.8.2.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/njqdev.vscode-python-typehint-1.4.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/ms-python.black-formatter-2023.6.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/wmaurer.change-case-1.0.0.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/mgesbert.python-path-0.0.14.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/dongli.python-preview-0.0.4.vsix
	code-server --install-extension ${SCRIPT_DIR}/utils/Semgrep.semgrep-1.6.3.vsix
```

guimou:
Sure. But then it's something that has to be fully customized to a specific
use case/environment. We can't impose all sorts of extensions, even a few,
to everyone just to cover this specific case.
In this scenario, one would only have to create a custom image deriving
from this one by adding the needed extensions in the utils folder.

shalberd:
Yes, this is what I did on contrib workbench images builder. I added a custom ide snippet for my vscode and used the logic here, with all the extensions I need in the utils folder.

However, the current approach to install extensions fails due to the imposed liveness probe limit that I had to manually patch to be waiting 100s initially so the extension install can run through.

Since liveness probe is something that is statically created by odh dashboard, this is not very good.

harshad16:
Thanks @shalberd and @guimou , i will convert this into an issue
so we can have this and work on this as we proceed

Issue #372, opened 2023-11-27
Incorporate on vscode standard database clients for MySQL, PostgreSQL, MSSQL, and MongoDB


atheo89:
**Why you need this feature:**
[Is your feature request related to a problem? Please describe in details]
This can be useful for the end user to have these on the workbench embedded.

**Describe the solution you'd like:**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
Is a follow up to this issue: https://github.com/opendatahub-io/notebooks/issues/345

https://issues.redhat.com/browse/RHOAIENG-140

atheo89:
Do you think it would be more beneficial to incorporate this integration into the base ubi9 image, allowing all notebooks beneath it to inherit these advantages?
Currently, we only introduce them in the datascience notebook, and as a result, all other notebooks inherit these benefits. Given that we need to integrate them into vscode, this implies that we will be repeating the same procedure as was done for the datascience notebook (and possibly later for the RStudio notebook).

harshad16:
Check in about these requirements: @harshad16

harshad16:
Closing this issue in favor of the JIRA issue: https://issues.redhat.com/browse/RHOAIENG-140

Issue #345, opened 2023-11-15
Pre-installed python packages in vscode environment


harshad16:
**Why you need this feature:**
[Is your feature request related to a problem? Please describe in details]

Pre-installed python packages in vscode environment.
- [ ] install minimal required packages. 

**Describe the solution you'd like:**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

https://issues.redhat.com/browse/RHOAIENG-224.

PR #374, opened 2023-11-28
pre-install minimal requirment python packages on vscode

atheo89:
Related to: https://github.com/opendatahub-io/notebooks/issues/345 

## Description

This PR introduce a minimal list of python packages that are required for data sciences and programming workloads.

**NOTE:** Will follow up a separated issue/pr to incorporate the standard database clients https://github.com/opendatahub-io/notebooks/issues/372

## How Has This Been Tested?
1. Spin up the vscode image generated from this PR
2. Open a terminal inside the container and run `pip list` and ensure there is a long list with packages
3. Test some of the installed packages, by creating a .py or .ipynb file, add the following code:
```
import pandas as pd
import numpy as np
import matplotlib.pyplot as plt

# Create a sample DataFrame
data = {'A': np.random.rand(100),
        'B': np.random.randn(100),
        'C': np.linspace(0, 1, 100)}

df = pd.DataFrame(data)

# Perform some operations using NumPy
df['D'] = np.sin(df['A']) + np.cos(df['B'])

# Plotting with Matplotlib
plt.figure(figsize=(10, 6))

plt.subplot(2, 2, 1)
plt.scatter(df['A'], df['B'])
plt.title('Scatter Plot')

plt.subplot(2, 2, 2)
plt.plot(df['C'], df['D'], color='red')
plt.title('Line Plot')

plt.subplot(2, 2, 3)
plt.hist(df['B'], bins=20, color='green', alpha=0.7)
plt.title('Histogram')

plt.subplot(2, 2, 4)
plt.boxplot(df[['A', 'C']])
plt.title('Boxplot')

plt.tight_layout()
plt.show()
```
4. You shall see the following plot:
image

5. In case of issue check which interpreter you are using 

## Merge criteria:

### DIFF
--- a/codeserver/ubi9-python-3.9/Pipfile
+++ b/codeserver/ubi9-python-3.9/Pipfile
@@ -1,0 +1,38 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[[source]]
+name = "pytorch"
+url = "https://download.pytorch.org/whl/cu118"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+# Base packages
+wheel = "~=0.41.2"
+setuptools = "~=68.1.2"
+
+# Datascience packages
+boto3 = "~=1.29.7"
+kafka-python = "~=2.0.2"
+matplotlib = "~=3.8.2"
+numpy = "~=1.26.2"
+pandas = "~=2.1.3"
+plotly = "~=5.18.0"
+scikit-learn = "~=1.3.2"
+scipy = "~=1.11.4"
+skl2onnx = "~=1.15.0"
+ipykernel = "~=6.26.0"
+
+# Some extra usefull packages
+opencensus = "~=0.11.3"
+smart-open = "~=6.4.0"
+virtualenv = "~=20.21.0"
+py-spy = "~=0.3.14"
+prometheus-client = "~=0.19.0"
+
+[requires]
+python_version = "3.9"
--- a/codeserver/ubi9-python-3.9/Dockerfile
+++ b/codeserver/ubi9-python-3.9/Dockerfile
@@ -17,6 +17,20 @@ USER 0
 
 WORKDIR /opt/app-root/bin
 
+# Install usefull packages from Pipfile.lock
+COPY Pipfile.lock ./
+
+# Install packages and cleanup
+RUN echo "Installing softwares and packages" && \
+    micropipenv install && \
+    rm -f ./Pipfile.lock && \
+    # Fix permissions to support pip in Openshift environments \
+    chmod -R g+w /opt/app-root/lib/python3.9/site-packages && \
+    fix-permissions /opt/app-root -P
+
+# Install usefull OS packages
+RUN dnf install -y jq git-lfs libsndfile
+
 # Install Code Server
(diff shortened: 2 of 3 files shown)


rkpattnaik780 (approved):
LGTM! Able to run code snippets that use the added libraries.

/approve

atheo89:
Glad to hear that! :slightly_smiling_face: 

> Also, we have a topic of where to put VSCode extensions and make them available without download ... but that is a different story. 

We are cooking something as well for this! Check out here: https://github.com/opendatahub-io/notebooks/pull/347

> Best regards from snowy CH.

Greetings from Turin on the other side of the Alps!

codeserver/ubi9-python-3.9/Pipfile
```diff
@@ -0,0 +1,47 @@
...
+scipy = "~=1.11.4"
+skl2onnx = "~=1.15.0"
+codeflare-sdk = "~=0.12.1"
+kfp = "~=1.8.22"
+kfp-tekton = "~=1.8.1"
+ipykernel = "~=6.26.0"
+
+# PyTorch packages
+tensorboard = "~=2.15.1"
+torch = {version = "~=2.1.1", index = "pytorch"}
```
harshad16, line 35: any reason for only installing torch and not tensorflow.
on hindsight , i feel we should not install any of these big packages.
and only stick with basic packages like

boto3, matplotlib, pandas, numpy, scipy

harshad16 (changes requested):
Thanks for the excellent work.
After closing checking, feels like we are install more packages then intended

codeserver/ubi9-python-3.9/Pipfile
shalberd: yes, there is a similar issue in @guimou recent changes to the contrib datascience UI pathon packages, they have all sorts of heavy nvidia* libraries in it, regardless of whether the base docker image is for Tensorflow or not. https://github.com/opendatahub-io-contrib/workbench-images/issues/48. Agreed that in a first step here, minimal might be good. And you could get these changes here in this PR into the snippets for codeserver VSCode [at contrib](https://github.com/opendatahub-io-contrib/workbench-images/tree/main), so users can then always add more of their own packages, if they want to, in own builds.

guimou:
Yeah, the issue comes from the codeflare sdk, which has dependencies on
pytorch, then nvidia...
For contrib I wanted to introduce codeflare in all the images, as well as
Elyra/kfp-tekton, but I'm not sure it's a good idea.
I talked to the Codeflare team, they were supposed to review their
dependencies, I did not check yet if they managed to have less heavy
dependencies.
So I guess it will come down to another discussion: which images should
include Codeflare sdk?

codeserver/ubi9-python-3.9/Pipfile
shalberd: answerered in contrib https://github.com/opendatahub-io-contrib/workbench-images/issues/48#issuecomment-1842844803

shalberd:
@guimou to keep that aspect separate, I answered in [odh-contrib](https://github.com/opendatahub-io-contrib/workbench-images/issues/48#issuecomment-1842844803)

codeserver/ubi9-python-3.9/Pipfile
atheo89: Thank you for reviewing this! 😊 I was in a dilemma about whether to include the most advanced packages or not. I agree with you; I'll omit them for now and stick to the basic data science  packages.

harshad16 (approved):
Thanks for the work, lgtm.

/lgtm
/approve

Thanks all

Issue #373, opened 2023-11-27
Recommended image version label for workbench is showed for two versions


jstourac:
This is an upstream issue for the following downstream one https://issues.redhat.com/browse/RHODS-12870.

**What steps did you take and what happened:**
For workbench notebook images, there is a Recommended label showed for the latest version of the notebook image. 

Unfortunately, this seem to be broken during upgrade operation as this label is showed for two versions - the new 2023b and also for version 2023a, which was tagged with this label in previous release. We could see it with the upgrade from 1.33 to 2.4.

You can find more information in https://issues.redhat.com/browse/RHODS-12870

This issue is present only after the upgrade operation (1.33 to 2.3 ATM) and isn't present with the fresh installation.

Root cause of the issue is right in the relevant imagestreams definitions, as I can see following for one of the image:

```
   - name: '1.2'
      annotations:
        opendatahub.io/notebook-python-dependencies: >-
          [{"name":"JupyterLab","version": "3.2"}, {"name":
          "Notebook","version": "6.4"}]
        opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.8"}]'
        openshift.io/imported-from: quay.io/modh/odh-minimal-notebook-container
      from:
        kind: DockerImage
        name: >-
          quay.io/modh/odh-minimal-notebook-container@sha256:39068767eebdf3a127fe8857fbdaca0832cdfef69eed6ec3ff6ed1858029420f
      generation: 4
      importPolicy: {}
      referencePolicy:
        type: Source
    - name: '2023.1'
      annotations:
        opendatahub.io/default-image: 'true'
        opendatahub.io/notebook-python-dependencies: >-
          [{"name":"JupyterLab","version": "3.5"}, {"name":
          "Notebook","version": "6.5"}]
        opendatahub.io/notebook-software: '[{"name":"Python","version":"v3.9"}]'
        opendatahub.io/workbench-image-recommended: 'true'
        openshift.io/imported-from: quay.io/modh/odh-minimal-notebook-container
      from:
        kind: DockerImage
        name: >-
          quay.io/modh/odh-minimal-notebook-container@sha256:54a5af13057df3445f917494a8cc12a0bed081c67204e00c9ed26daeaf9d0079
      generation: 4
      importPolicy: {}
      referencePolicy:
        type: Source
    - name: '2023.2'
      annotations:
        opendatahub.io/default-image: 'true'
        opendatahub.io/notebook-python-dependencies: >-
[...]

PR #376, opened 2023-11-28
hot-fix: set the workbench-image-recommended anno to false for 2023.1

harshad16:
set the workbench-image-recommended anno to false for 2023.1

## Description

Related-to: https://github.com/opendatahub-io/notebooks/issues/373

In ODH 2.2 upgrade to 2.4 + version, the recommend-tag annotation shows up for 2023.2 and 2023.1.
The 2023.1 would not have the recommended tag with the annotation set to false.

## How Has This Been Tested?

1. On a cluster, which was upgrade to 2.4 , mark the recommended annotation to false.
Screenshot from 2023-11-28 12-39-52

## Merge criteria:


harshad16:
>  maybe we should assign the `recommended = false` also for the 1.2 images... but probably not necessary.

as we never had that annotation for that tag, including it would be necessary.
also we hide it now in 2.5 , so more reason not to do it.

PR #395, opened 2024-02-02
Fix the typo and keep code-server case-sensitive

harshad16:
Fix the typo and keep code-server case-sensitive

## Description
Fix the typo and keep code-server case-sensitive
Bring this here as well for consistency , was hot-fixed downstream.


harshad16 (commented):
/approve
thanks for the review :100:

PR #433, opened 2024-02-26
ci: initial implementation of top-level pytest tests

jiridanek:
## Description
This establishes a top level Poetry-managed environment for writing tests for Notebooks in Python. There is currently a single test to check that all `Pipfile`s have the correct python version specified in them.

Documentation

* https://python-poetry.org/docs/basic-usage/
* https://docs.pytest.org/en/7.1.x/getting-started.html

Pytest supports two styles of writing tests, either class-based (`unittest` package style) or function and fixtures-based (`pytest` package style). Here's docs for the unittests approach https://docs.python.org/3/library/unittest.html#basic-example, and for the support in Pytest https://docs.pytest.org/en/7.1.x/how-to/unittest.html.

Main advantage of Pytest is that it can export results as JUnit XML files out of the box, and that it supports assertions using the `assert` keyword, instead of having to use the assert methods.

## Merge criteria:


jiridanek:
Yeah, this, @jstourac just mentioned this PR at a meeting, asking about status

.github/workflows/code-quality.yaml
```diff
@@ -30,6 +30,46 @@ jobs:
...
+        uses: actions/cache/save@v4
+        with:
+          path: ~/.local
+          key: ${{ steps.cache-poetry-restore.outputs.cache-primary-key }}
+
+      - name: Set up Python
+        id: setup-python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
```
jstourac, line 62: I'm wondering whether such new python version won't collide with possible future basic functional tests to test actual python packages? Shouldn't we rather stick to what we have in 2024a branch, that means Python 3.11? :thinking:

jstourac (commented):
I wonder - do we plan to utilize this also for the functional testing of the build images or is this planned just as a set of checks for something like the static code analysis? Maybe a combination of both?

My concern here is that we don't introduce yet another "static code" check which we already have written in bash scripts elsewhere. The ultimate goal should be to replace the current Makefile like testing approach with something more flexible, cleaner and more scalable. Does this apply to this approach?

.github/workflows/code-quality.yaml
jiridanek: Python 3.12 is not a problem. I think that that would be a valid comment for what @paulovmr is doing in

* https://github.com/opendatahub-io/notebooks/pull/610

(and he has that handled, in his PR there's a separate env for every image)

Here, if we wanted to run something inside the images, which is the only meaningful way to do functional tests of packages installed in images, then we would naturally not use the python, package version, or lock file maintained in the top level by poetry, but we'd run pip install the same way as the current makefile tests install papermill, etc.

jiridanek (commented):
> I wonder - do we plan to utilize this also for the functional testing of the build images or is this planned just as a set of checks for something like the static code analysis? Maybe a combination of both?

I want to use https://github.com/testcontainers/testcontainers-python to test the built images.

My concern here is that we don't introduce yet another "static code" check which we already have written in bash scripts elsewhere. The ultimate goal should be to replace the current Makefile like testing approach with something more flexible, cleaner and more scalable. Does this apply to this approach?

Absolutely; replacing static checks with Python would be also nice, I think, but the main goal for me is testing the built images.

jstourac (approved):

PR #462, opened 2024-03-16
[CI] fix the expected metadata of the images

jstourac:
Makes CI check for params.env happy again.

## Merge criteria:

### DIFF
--- a/ci/check-params-env.sh
+++ b/ci/check-params-env.sh
@@ -24,7 +24,7 @@ PARAMS_ENV_PATH="manifests/base/params.env"
 
 # This value needs to be updated everytime we deliberately change number of the
 # images we want to have in the `params.env` file.
-EXPECTED_NUM_RECORDS=24
+EXPECTED_NUM_RECORDS=19
 
 # ---------------------------- DEFINED FUNCTIONS ----------------------------- #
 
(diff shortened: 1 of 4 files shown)


ci/check-params-env.sh
```diff
@@ -69,6 +69,7 @@ function check_image_variable_matches_name_and_commitref() {
     case "${image_variable}" in
         odh-minimal-notebook-image-n)
             expected_name="odh-notebook-jupyter-minimal-ubi9-python-3.9"
+            # expected_commitref="2024a" or main
             expected_commitref="2023b"
```
atheo89, line 73: This shouldn't be 2024a? :thinking:

jstourac: these should be - as is stated in the comment, but right now the images that are used are still compiled with the 2023b - I though that this is an expected temporary state as this was for some time before. Or has the all images been updated already? Can check later for this if you want.

atheo89: Alright, should be as the comment indicates. There have been recent changes resulting in upside-down modifications to the params.env. Below is the manifest params.env file that we will adhere to until the next release.
https://github.com/opendatahub-io/notebooks/blob/main/manifests/base/params.env

jstourac: Thanks, didn't notice it has been updated in the meantime. I rebased this PR and updated. Validation now passes with the expected ref values.

Only exception is habana, where we still have a `main`. I suppose that this image hasn't been updated to 2024a yet - which is also reason that there is still just one version available, correct - https://github.com/opendatahub-io/notebooks/blob/main/manifests/base/params.env#L13 ?

atheo89: Yeah, for habana we don't have any updates

PR #533, opened 2024-05-22
Update of Habana SynapseAI notebook to 1.15.1

Xaenalt:
## Description
Adds the SynapseAI 1.15.1 notebook, which is needed for RHOAI 2.10 compatibility

This required quite a few changes, since the SynapseAI stack requires Python 3.11 for 1.15.1. Uses a similar workflow to Anaconda with a custom base image, this however should be easy to update in the future just by swapping out the FROM in the base image and verifying the python versions going forward.

## How Has This Been Tested?
Builds successfully, could use some functional testing

## Merge criteria:

### DIFF
--- a/base/habana-python-3.10/Dockerfile
+++ b/base/habana-python-3.10/Dockerfile
@@ -1,0 +1,77 @@
+FROM vault.habana.ai/gaudi-docker/1.15.1/rhel9.2/habanalabs/pytorch-installer-2.2.0:1.15.1-15 as base
+
+LABEL name="odh-notebook-habana-jupyter-1.15.1-ubi9-python-3.10" \
+    summary="Jupyter HabanaAI 1.15.1 notebook image for ODH notebooks" \
+    description="Jupyter HabanaAI 1.15.1 notebook image with base Python 3.10 builder image based on ubi9 for ODH notebooks" \
+    io.k8s.display-name="Jupyter HabanaAI 1.15.1 notebook image for ODH notebooks" \
+    io.k8s.description="Jupyter HabanaAI 1.15.1 notebook image with base Python 3.10 builder image based on ubi9 for ODH notebooks" \
+    authoritative-source-url="https://github.com/opendatahub-io/notebooks" \
+    io.openshift.build.commit.ref="main" \
+    io.openshift.build.source-location="https://github.com/opendatahub-io/notebooks/tree/main/habana/1.15.1/ubi9-python-3.10" \
+    io.openshift.build.image="quay.io/opendatahub/workbench-images:habana-jupyter-1.15.1-ubi9-python-3.10"
+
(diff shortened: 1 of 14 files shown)


Xaenalt:
@harshad16 PTAL

Xaenalt:
Yeah, thankfully the same versions of everything exist in 3.10, but yeah let's have a longer discussion about it

Xaenalt:
According to Intel, this would only be necessary until the 1.17 release which will be end of July-ish

jiridanek:
> @Xaenalt , i m not sure if we would like to maintain python 3.10 to be honest

Red Hat ships Python 3.11, can you use that? It is supported until it's retirement date in May 2026. If you base the image on ub9/rhel9 and use Python 3.12, that has retirement date in Apr 2027. https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle#rhel8_application_streams

Xaenalt:
If we're good to use Python 3.11 that'll be excellent, and will be supported in the 1.17 release

I was just doing a few small fixes to the notebook PR, since it's a much easier to use template for the future

Xaenalt:
If possible, I'd like to revisit the timeline until 1.17 (EO July timeframe) at which point we can retire the 1.15/1.16 series which are stuck on Python 3.10. Currently (pre 1.17) the notebook SynapseAI version has to match what's in the operator

jstourac:
@Xaenalt JFYI, I updated the description of the https://issues.redhat.com/browse/RHOAIENG-5404 which is tracking addition of the Gaudi v1.17 into RHOAI as v1.17 seems to be out now.

In the meantime, there is a plan to introduce a Python 3.11 via #659 (just the image definitions but won't be used in the actual RHOAI/ODH builds yet - that will be a next step).

base/habana-python-3.10/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+    fix-permissions ${APP_ROOT} -P && \
+    rpm-file-permissions && \
+    echo "unset BASH_ENV PROMPT_COMMAND ENV" >> ${APP_ROOT}/bin/activate
+
+ENV BASH_ENV="${APP_ROOT}/bin/activate" \
+    ENV="${APP_ROOT}/bin/activate" \
+    PROMPT_COMMAND=". ${APP_ROOT}/bin/activate"
+
+# Squash a warning from the Habana runtime
+ENV TSAN_OPTIONS='ignore_noninstrumented_modules=1'
```
jiridanek, line 49: are we really compiling something with thread sanitizer enabled? this gives you 3 to 10x more memory consumption and about the same amount of performance degradation; tsan is intended for debug builds only

jstourac:
I think that this work will be superseded by the #695.

PR #557, opened 2024-06-12
Setup build images for Pytorch & Tensorflow with base ROCm image

dibryant:
Fixes for https://issues.redhat.com/browse/RHOAIENG-6377
## Merge criteria:


harshad16 (changes requested):
great start @dibryant 
please also include the rest of the files for respective directories
https://github.com/opendatahub-io/notebooks/tree/main/jupyter/pytorch/ubi9-python-3.9
https://github.com/opendatahub-io/notebooks/tree/main/jupyter/tensorflow/ubi9-python-3.9

jupyter/amd/pytorch/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+      authoritative-source-url="https://github.com/opendatahub-io/notebooks" \
+      io.openshift.build.commit.ref="main" \
+      io.openshift.build.source-location="https://github.com/opendatahub-io/notebooks/tree/main/amd/python-3.9" \
+      io.openshift.build.image="quay.io/opendatahub/workbench-images:rocm-python-3.9"
+
+USER 0
+WORKDIR /opt/app-root/bin
+
+ARG ROCM_VERSION=6.1
+ARG AMDGPU_VERSION=6.1
```
jstourac, line 17: Looks like 6.1.2 version is the latest one as of today, don't we want to use that one?
* https://repo.radeon.com/amdgpu/
* https://repo.radeon.com/rocm/el9/

jupyter/amd/pytorch/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,102 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+LABEL name="odh-notebook-rocm-python-3.9" \
+      summary="ROCm Python 3.9 base image for ODH notebooks" \
+      description="ROCm Python 3.9 builder image based on CentOS Stream 9 for ODH notebooks" \
+      io.k8s.display-name="ROCm Python 3.9 base image for ODH notebooks" \
+      io.k8s.description="ROCm Python 3.9 builder image based on C9S for ODH notebooks" \
```
jstourac, line 7: Please update the labels so they match - we don't use CentOS here but UBI9.

jupyter/amd/tensorflow/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,102 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+LABEL name="odh-notebook-rocm-python-3.9" \
+      summary="ROCm Python 3.9 base image for ODH notebooks" \
+      description="ROCm Python 3.9 builder image based on CentOS Stream 9 for ODH notebooks" \
+      io.k8s.display-name="ROCm Python 3.9 base image for ODH notebooks" \
+      io.k8s.description="ROCm Python 3.9 builder image based on C9S for ODH notebooks" \
```
jstourac, line 7: Same as the other file - please update reference from CentOS to UBI9 as we do in other similar files.

jupyter/amd/pytorch/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+# Install the ROCm rpms
+RUN echo "[ROCm]" > /etc/yum.repos.d/rocm.repo && \
+    echo "name=ROCm" >> /etc/yum.repos.d/rocm.repo && \
+    echo "baseurl=https://repo.radeon.com/rocm/el9/$ROCM_VERSION/main" >> /etc/yum.repos.d/rocm.repo && \
+    echo "enabled=1" >> /etc/yum.repos.d/rocm.repo && \
+    echo "gpgcheck=0" >> /etc/yum.repos.d/rocm.repo
+
+RUN echo "[amdgpu]" > /etc/yum.repos.d/amdgpu.repo && \
+    echo "name=amdgpu" >> /etc/yum.repos.d/amdgpu.repo && \
+    echo "baseurl=https://repo.radeon.com/amdgpu/$AMDGPU_VERSION/el/9.2/main/x86_64" >> /etc/yum.repos.d/amdgpu.repo && \
```
jstourac, line 81: Would it be possible to have some ARG variable e.g. `EL_VERSION` defined at the top of the file together with the `AMDGPU_VERSION` and `ROCM_VERSION` so that this is more visible? Something like:
```suggestion
    echo "baseurl=https://repo.radeon.com/amdgpu/$AMDGPU_VERSION/el/$EL_VERSION/main/x86_64" >> /etc/yum.repos.d/amdgpu.repo && \
```

jstourac, line 81: Moreover - we use [this image](https://github.com/opendatahub-io/notebooks/blob/main/base/ubi9-python-3.9/Dockerfile#L1) as the very base for our UBI9 images:
* https://catalog.redhat.com/software/containers/ubi9/python-39/61a61032bfd4a5234d59629e?architecture=amd64&image=666134175ea3c2839df25dad&container-tabs=dockerfile

That image states that it's based on this image:
```
FROM ubi9/s2i-base:rhel9.4.0
```

So instead of the `9.2` version, we should use `9.4`, thus something like this:
```
ARG EL_VERSION=9.4
```

We can either have this hardcoded as above (disadvantage that we need to think about this and update it when necessary), or we can call something like this eventually:
```
cat /etc/os-release | grep "VERSION_ID" | sed 's#.*"\(.*\)"#\1#'
```
Disadvantage in this second case is that we may possibly hit a window when the AMD sources won't be available for the particular UBI version yet (e.g. since it's too fresh - aka we updated our base image too soon).

From my point of view it's a bit better to use the latter approach since this gives as an heads up at least. But I don't insist.

jupyter/amd/tensorflow/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+# Install the ROCm rpms
+RUN echo "[ROCm]" > /etc/yum.repos.d/rocm.repo && \
+    echo "name=ROCm" >> /etc/yum.repos.d/rocm.repo && \
+    echo "baseurl=https://repo.radeon.com/rocm/el9/$ROCM_VERSION/main" >> /etc/yum.repos.d/rocm.repo && \
+    echo "enabled=1" >> /etc/yum.repos.d/rocm.repo && \
+    echo "gpgcheck=0" >> /etc/yum.repos.d/rocm.repo
+
+RUN echo "[amdgpu]" > /etc/yum.repos.d/amdgpu.repo && \
+    echo "name=amdgpu" >> /etc/yum.repos.d/amdgpu.repo && \
+    echo "baseurl=https://repo.radeon.com/amdgpu/$AMDGPU_VERSION/el/9.2/main/x86_64" >> /etc/yum.repos.d/amdgpu.repo && \
```
jstourac, line 81: Same comment as in the other file - let's have an arg for this value and move to 9.4.

jstourac (commented):
Thank you for this. I did very brief check and I've put some comments.

jupyter/amd/pytorch/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,27 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+LABEL name="odh-notebook-jupyter--amd-pytorch-ubi9-python-3.9" \
```
atheo89, line 4:
```suggestion
LABEL name="odh-notebook-jupyter-amd-pytorch-ubi9-python-3.9" \
```

jupyter/amd/pytorch/ubi9-python-3.9/Pipfile
```diff
@@ -0,0 +1,64 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[[source]]
+url = "https://download.pytorch.org/whl/nightly/rocm6.1/"
```
atheo89, line 7: I found the version on this [link](https://pytorch.org/get-started/previous-versions/#linux-and-windows) to be 6.0. Does this update automatically somehow?

```suggestion
url = "https://download.pytorch.org/whl/nightly/rocm6.0/"
```

jupyter/amd/pytorch/ubi9-python-3.9/test/test_notebook.ipynb
```diff
@@ -0,0 +1,163 @@
...
+    "\n",
+    "class TestPytorchNotebook(unittest.TestCase):\n",
+    "    \n",
+    "    def test_python_version(self):\n",
+    "        expected_major_minor = '3.9'  # Set the expected version (x.y)\n",
+    "        actual_major_minor = '.'.join(python_version().split('.')[:2]) \n",
+    "        self.assertEqual(actual_major_minor, expected_major_minor, \"incorrect version\")\n",
+    "    \n",
+    "    def test_torch_version(self):\n",
+    "        expected_major_minor = '2.2'  # Set the expected version (x.y)\n",
```
atheo89, line 27:
```suggestion
    "        expected_major_minor = '2.3'  # Set the expected version (x.y)\n",
```

jupyter/amd/tensorflow/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,27 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+LABEL name="odh-notebook-cuda-jupyter-tensorflow-ubi9-python-3.9" \
```
atheo89, line 4:
```suggestion
LABEL name="odh-notebook-amd-jupyter-tensorflow-ubi9-python-3.9" \
```

atheo89 (changes requested):
Thank you for adding this work. I've added some comments to the files. Could you please take a look at them?

Additionally, could you add the .PHONY recipes to the Makefile? This would allow us to build these images locally as we do for the rest.

jupyter/amd/pytorch/ubi9-python-3.9/Pipfile
jiridanek, line 7:
```suggestion
url = "https://download.pytorch.org/whl/rocm6.0/"
```

I don't like the `/nightly/` part. We should use stable releases, not nightly builds. So maybe like this?

atheo89:
Moreover, something that would be super useful for the PR review, but also as a complementary step to fulfill this work, is to set up the OCP-CI to build the newly created AMD PyTorch and TensorFlow notebooks.

jiridanek:
Whoa, was wondering why the gha is failing, and it looks like it is running out of disk space

```
Total download size: 2.2 G
Installed size: 25 G
Downloading Packages:
```

Is it really necessary to install this much stuff? If yes, I'll get you more disk space tomorrow ;)

jiridanek:
This ci fail is about missing lock file, https://github.com/opendatahub-io/notebooks/actions/runs/9600777904/job/26477839250?pr=557

jiridanek:
> Whoa, was wondering why the gha is failing, and it looks like it is running out of disk space

Here you go

* https://github.com/opendatahub-io/notebooks/pull/577

atheo89:
Adding this here as ref: https://github.com/openshift/release/pull/53309

jupyter/amd/pytorch/ubi9-python-3.9/Dockerfile
harshad16: Marking this a resolved, as wrong Dockerfile was pushed earlier.

harshad16:
This is look good to me 
i will be merging this for now.

The size is definitely as concern, we should take a deeper look at this.
and try to see, if this can be reduced.

Thanks all for the review and great work on this.
/lgtm
/approve

PR #576, opened 2024-06-20
Set Telemetry level off

atheo89:
Related to: https://issues.redhat.com/browse/RHOAIENG-3348

## Description
This PR set Telemetry off on code-server notebook. 

## How Has This Been Tested?
1. Import the below image as custom notebook, with `code-server-telemetry-off` name on dashboard
`quay.io/rh_ee_atheodor/workbench-images@sha256:826646b64a20dfa836e037a143d3996ab6c04c7bc0d730b3463dbcf264d76021`
2. Spin up the workbench that just imported (`code-server-telemetry-off``)
3. Go to File -> Preferences -> settings ->search for `telemetry` -> check that the drop-down listbox is set to `off` the default value was `all` 

## Merge criteria:


atheo89:
Hey @jiridanek, can you reproduce the steps that you did at that time when you found that we had telemetry on, and check if we miss anything behind? So, far the only check that i did is to see if the elementary setting is `off` applied from `.local/share/code-server/User/settings.json`

codeserver/ubi9-python-3.9/run-code-server.sh
```diff
@@ -13,60 +13,69 @@ if [ ! -f "/opt/app-root/src/.bashrc" ]; then
...
+# Define universal settings
+universal_dir="/opt/app-root/src/.local/share/code-server/User/"
+user_settings_filepath="${universal_dir}settings.json"
+universal_json_settings='{
+  "python.defaultInterpreterPath": "/opt/app-root/bin/python3",
+  "telemetry.telemetryLevel": "off",
+  "telemetry.enableTelemetry": false,
+  "telemetry.enableCrashReporter": false,
+  "telemetry.enableErrorTelemetry": false,
+  "workbench.enableExperiments": false
```
jiridanek, line 51: Checking; so far, I see that Jupyter experiments may not be disabled, according to the screenshot below, because that's a separate setting, separate from https://github.com/atheo89/notebooks/commit/0d1ab4d33feb187e747f626e2d18b881ddc0a606#r143328849

image

Jupyter extension docs say that opting out of telemetry opts out of AB testing. Does that mean that experiments are disabled? https://github.com/microsoft/vscode-jupyter/wiki/Experiments

It would make sense to disable all experiments, IMO.

jiridanek:
> Hey @jiridanek, can you reproduce the steps that you did at that time when you found that we had telemetry on, and check if we miss anything behind? So, far the only check that i did is to see if the elementary setting is `off` applied from `.local/share/code-server/User/settings.json`

Checked, I wrote the steps at https://issues.redhat.com/browse/RHOAIENG-3348. Looks good.

jiridanek (approved):
/lgtm

codeserver/ubi9-python-3.9/run-code-server.sh
jiridanek: Anyways, we can do this in a separate issue later.

atheo89: I checked again about this, as far as i understand since we set `workbench.enableExperiments: false` to the settings file it disables the participation in A/B testing and experimental features. This setting stops the editor from enabling experimental features that are being tested with a subset of users. So, I believe we are good  until proven otherwise!

PR #600, opened 2024-06-28
Add daily check for vulnerability issues using Trivy

caponetto:
https://issues.redhat.com/browse/RHOAIENG-8779

## Description

This PR introduces the usage of [Trivy](https://trivy.dev/) to scan the container images and consolidate vulnerability reports. I've configured the `Build Notebooks` workflow to be triggered daily at 2 am UTC to generate the report (the scan runs after each image is built). I didn't see the need to run it on each `push` but we can enable it if it makes sense. Each report is uploaded to the workflow summary as soon as each individual job is completed. Finally, reports follow a markdown template.

Demo

https://github.com/opendatahub-io/notebooks/assets/638737/68f92ecb-40a2-4fa9-8f19-c5037f98458d

Here is an [example of workflow run](https://github.com/caponetto/opendatahub-io-notebooks/actions/runs/9713547448) where you can see real reports.

**Note**: I haven't added the checks for PRs because the scan operation adds an extra ~10 minutes, which would slow our PR jobs down.

## How Has This Been Tested?

I've executed the workflow on my fork.

## Merge criteria:

### DIFF
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -43,6 +43,14 @@ jobs:
 
           df -h
 
+          sudo apt-get update
+          sudo apt-get remove -y '^dotnet-.*'
+          sudo apt-get remove -y '^llvm-.*'
+          sudo apt-get remove -y 'php.*'
+          sudo apt-get remove -y '^mongodb-.*'
+          sudo apt-get autoremove -y
+          sudo apt-get clean
+          sudo rm -rf /usr/local/.ghcup &
           sudo rm -rf /usr/local/lib/android &
           sudo rm -rf /usr/local/share/boost &
           sudo rm -rf /usr/local/lib/node_modules &
@@ -96,13 +104,39 @@ jobs:
           mkdir -p $HOME/.local/share/containers/storage/tmp
 
       # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push
-      - name: "push: make ${{ inputs.target }}"
+      - name: "push|schedule: make ${{ inputs.target }}"
         run: "make ${{ inputs.target }}"
-        if: "${{ fromJson(inputs.github).event_name == 'push' }}"
+        if: ${{ fromJson(inputs.github).event_name == 'push' || fromJson(inputs.github).event_name == 'schedule' }}
         env:
           IMAGE_TAG: "${{ github.ref_name }}_${{ github.sha }}"
           IMAGE_REGISTRY: "ghcr.io/${{ github.repository }}/workbench-images"
           CONTAINER_BUILD_CACHE_ARGS: "--cache-from ${{ env.CACHE }} --cache-to ${{ env.CACHE }}"
+
+      - name: "schedule: run Trivy vulnerability scanner"
+        if: "${{ fromJson(inputs.github).event_name == 'schedule' }}"
+        run: |
+          TRIVY_VERSION=0.52.2
+          REPORT_FOLDER=${{ github.workspace }}/report
+          REPORT_FILE=trivy-report.md
+          REPORT_TEMPLATE=trivy-markdown.tpl
+
+          mkdir -p $REPORT_FOLDER
+          cp ci/$REPORT_TEMPLATE $REPORT_FOLDER
+
+          IMAGE_NAME=ghcr.io/${{ github.repository }}/workbench-images:${{ inputs.target }}-${{ github.ref_name }}_${{ github.sha }}
+          echo "Scanning $IMAGE_NAME"
+
+          podman run --rm \
+              -v $REPORT_FOLDER:/report \
+              docker.io/aquasec/trivy:$TRIVY_VERSION \
+                image \
+                --scanners vuln,secret \
+                --exit-code 0 --timeout 30m \
+                --severity CRITICAL,HIGH \
+                --format template --template "@/report/$REPORT_TEMPLATE" -o /report/$REPORT_FILE \
+                $IMAGE_NAME
+
+          cat $REPORT_FOLDER/$REPORT_FILE >> $GITHUB_STEP_SUMMARY
 
       # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
       - name: "pull_request: make ${{ inputs.target }}"
--- a/ci/trivy-markdown.tpl
+++ b/ci/trivy-markdown.tpl
@@ -1,0 +1,58 @@
+## Vulnerability Report by [Trivy](https://trivy.dev)
+
+<details>
+  {{- if . }}
+    {{- range . }}
+      {{- if or (gt (len .Vulnerabilities) 0) (gt (len .Misconfigurations) 0) }}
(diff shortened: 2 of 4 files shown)


.github/workflows/build-notebooks-TEMPLATE.yaml
```diff
@@ -43,6 +43,14 @@ jobs:
 
           df -h
 
+          sudo apt-get update
+          sudo apt-get remove -y '^dotnet-.*'
+          sudo apt-get remove -y '^llvm-.*'
+          sudo apt-get remove -y 'php.*'
+          sudo apt-get remove -y '^mongodb-.*'
+          sudo apt-get autoremove -y
+          sudo apt-get clean
```
jiridanek, line 52: Is this necessary? it adds more than 30 seconds to every job run, and given how many jobs we have, it adds up. Maybe run single `sudo apt-get remove -y` command that removes all these?

```suggestion
          sudo apt-get update
          sudo apt-get remove -y '^dotnet-.*' '^llvm-.*' 'php.*' '^mongodb-.*'
          sudo apt-get autoremove -y
          sudo apt-get clean
```

caponetto: It came out during my investigation of the full storage. These lines free around 7GB.
I can remove them as they are not part of the solution anymore but I figure it wouldn't hurt to keep them.

ci/trivy-markdown.tpl
```diff
@@ -0,0 +1,58 @@
+## Vulnerability Report by [Trivy](https://trivy.dev)
+
+<details>
```
jiridanek, line 3: Given that GitHub automatically hides the content of each except the first few files (https://github.com/caponetto/opendatahub-io-notebooks/actions/runs/9713547448#summary-26810623673) it may make more sense to not hide this. Now I have to click "expand" twice, instead of once.

caponetto: Given that we don't control how many of them GitHub hides, I preferred to set them all collapsed so that you can expand only the ones you are interested in seeing, avoiding uncontrolled scrolling.

jiridanek (approved):
Just a few small suggestions. Over all, lgtm.

We can't fail the build when vulnerabilities are detected, that way we'd never have a pass. But these inforeports are not very practical to work with either, there is too much info to scan. We'd need to either reduce the number of findings the tool produces, so it is possible to read through it easily.

Currently, this is mainly useful to check that some vulnerability we intended to fix was indeed fixed. But, then I'd have to first merge the PR to get scan on it. (Or run scan on my machine.)

It does not seem very actionable to me right away, but it is a great capability to have around. And, I did not know the github feature https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary, so that's interesting for me too.

caponetto:
@jiridanek thanks for the review!

> We can't fail the build when vulnerabilities are detected, that way we'd never have a pass. But these inforeports are not very practical to work with either, there is too much info to scan. We'd need to either reduce the number of findings the tool produces, so it is possible to read through it easily.

Yeah, I agree. I've reduced the scope to only `HIGH` and `CRITICAL` issues because the list was getting bigger if we allowed all of them. But let's see if it helps otherwise we can revisit it and try to personalize more.

> Currently, this is mainly useful to check that some vulnerability we intended to fix was indeed fixed. But, then I'd have to first merge the PR to get scan on it. (Or run scan on my machine.)

We can't enable it to run for all PRs due to the amount of time it adds but now I'm thinking if we can come up with some strategy like running it if the PR has a certain label or something like that.

.github/workflows/build-notebooks-TEMPLATE.yaml
```diff
@@ -96,14 +104,38 @@ jobs:
...
+        if: "${{ fromJson(inputs.github).event_name == 'schedule' }}"
+        run: |
+          TRIVY_VERSION=0.52.2
+          REPORT_FOLDER=${{ github.workspace }}/report
+          REPORT_FILE=trivy-report.md
+          REPORT_TEMPLATE=trivy-markdown.tpl
+
+          mkdir -p $REPORT_FOLDER
+          cp ci/$REPORT_TEMPLATE $REPORT_FOLDER
+          IMAGE_NAME=$(podman images --format '{{.Repository}}:{{.Tag}}' --no-trunc | head -n 1)
```
jiridanek, line 125: I had another look at this and I suspect that `IMAGE_NAME` may not be the image we intend to scan. It would be best to construct IMAGE_NAME from the variables the step above uses.

```

          IMAGE_TAG: "${{ github.ref_name }}_${{ github.sha }}"
          IMAGE_REGISTRY: "ghcr.io/${{ github.repository }}/workbench-images"
```

Can you add a debug print, so we are sure what we're scanning?

```suggestion
          IMAGE_NAME=$(podman images --format '{{.Repository}}:{{.Tag}}' --no-trunc | head -n 1)
          echo "[INFO] IMAGE_NAME=${IMAGE_NAME}"
```

caponetto: I had a problem when I tried to use the variables but let me double-check.

jiridanek: just the echo would be sufficient for now, just that I can always check what's it actually scanning

caponetto: Just tested. It was scanning the last build-cache created, which had the same ImageId of the image we wanted to check. Anyway, I'll change to:
```bash
IMAGE_NAME=ghcr.io/${{ github.repository }}/workbench-images:${{ inputs.target }}-${{ github.ref_name }}_${{ github.sha }}
echo "Scanning $IMAGE_NAME"
```
Good catch!

caponetto: The only thing is that it becomes too coupled with the logic defined in the Makefile.

jiridanek: github does not have sad emoji, and I can't use similarly looking confused emoji, because I'm not actually confused

😢

caponetto:
Please hold the merge. I thought about some minor improvements to this PR. I will explore them on Monday.

caponetto:
Ok, this PR is ready.

jiridanek (approved):

atheo89 (commented):
Great CVE report tool! I have a question: Do we push the images to ghcr after completing the report? If so, I would suggest not doing it because we will fill up the registry on a daily basis.

jiridanek:
> Do we push the images to ghcr after completing the report? If so, I would suggest not doing it because we will fill up the registry on a daily basis.

we have to push, the makefile is written in such a way that it builds and pushes, without the possibility to skip a push

https://github.com/opendatahub-io/notebooks/blob/3f93529616cf01bd8305a16e6671f6253ee27ba8/Makefile#L61-L65

regarding filling up space, first, there does not seem to be a limit that applies to us, and second, please review this PR that sets up a cleaning job to free up the space daily

* https://github.com/opendatahub-io/notebooks/pull/601

caponetto:
@atheo89 Thanks for raising this concern! Another option is to use the approach we have for `pull_request`, which publishes the image to localhost ([ref](https://github.com/opendatahub-io/notebooks/blob/main/.github/workflows/build-notebooks-TEMPLATE.yaml#L119)) but  I think it won't be an issue anymore after @jiridanek's PR that cleans up the registry periodically. Plus, it could be helpful to have them published somewhere if someone needs to debug the associated image with the report.

atheo89:
>we have to push, the makefile is written in such a way that it builds and pushes, without the possibility to skip a push

That's true; but it's not a big deal to make recipes more independent if that makes our lives easier.  Added a tracker [here](https://issues.redhat.com/browse/RHOAIENG-9349)

PR #558, opened 2024-06-13
Limit PR checks to build only the modified images

jiridanek:
## Description
Followup to

* https://github.com/opendatahub-io/notebooks/pull/543

that filters out the list of images to build during a PR check only to those that have some source files changed.

This will speed up the PR check build.

## How Has This Been Tested?
Created a draft PR that changes one file in a single docker image

* https://github.com/opendatahub-io/notebooks/pull/555

Checked that it correctly builds all images that depend on this one.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #604, opened 2024-06-30
ci: filter GitHub Actions PR build by changed files and `.dockerignore`s

### LINKED ISSUE #581: Can we implement filtering to avoid building images for certain file types, such as *.md or *.yaml? Probably as future enhancement
Can we implement filtering to avoid building images for certain file types, such as *.md or *.yaml? Probably as future enhancement
 :thinking: This idea came to me when I opened this PR:: https://github.com/opendatahub-io/notebooks/pull/573

_Originally posted by @atheo89 in https://github.com/opendatahub-io/notebooks/issues/558#issuecomment-2175775840_

jiridanek:
## Description
Fixes #581

## How Has This Been Tested?

I'm not conviced this is worth doing. Pretty much the only file that we can `.dockerignore` are the `README.md`s under `habana/`. The issue mentions `*.yaml`s and so on, but these files aren't located in the container image directories, so when they are modified, the images won't rebuild simply because of #558.

## Merge criteria:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+# This file is stored in .dockerignore in the root of the repository
+# and is symlinked to every image directory below, so that each .dockerignore has the same content
+
+README.md
--- a/amd/c9s-python-3.9/.dockerignore
+++ b/amd/c9s-python-3.9/.dockerignore
@@ -1,0 +1,1 @@
+../../.dockerignore
--- a/base/anaconda-python-3.8/.dockerignore
+++ b/base/anaconda-python-3.8/.dockerignore
@@ -1,0 +1,1 @@
+../../.dockerignore
--- a/base/c9s-python-3.9/.dockerignore
+++ b/base/c9s-python-3.9/.dockerignore
@@ -1,0 +1,1 @@
+../../.dockerignore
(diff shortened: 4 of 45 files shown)


jiridanek:
Looks like podman will not follow symlink for .dockerignore that leads out of the build context. The images do build in docker just fine, but not in podman.

If I wanted to fix this, I'd add script to ci/ that synces the .dockerignore files and added to static check workflow that all .dockerignores are properly synced.

But, it seems to me that this whole PR is not worth the effort and increased complexity, given how few files there is to be filtered out.

jiridanek:
@atheo89 as the original author of the general idea, do you think this is worth pursuing, or can I close the PR and the issue and not do this, given what I found out in my previous comment?

atheo89:
>@atheo89 as the original author of the general idea, do you think this is worth pursuing, or can I close the PR and the issue and not do this, given what I found out in my previous comment? 

Apologies for the late reply. I appreciate your proposed solution involving syncing .dockerignore files via a script in the CI workflow. Given the minimal number of files that need to be filtered out, I agree that pursuing this change may not justify the added complexity. You can proceed with closing this PR.

PR #607, opened 2024-07-04
RHOAIENG-9473: Optionally enable Trivy for pull requests

caponetto:
https://issues.redhat.com/browse/RHOAIENG-9473

## Description

This PR enables Trivy scan for PRs. It won't run by default for all PRs because it adds extra minutes to the jobs and the report won't be needed for all PRs. In order to execute the scan, the label `trivy-scan` must be added to the PR before its creation. The workflow does not react to label changes, so a new commit must be pushed to run the scan if the `trivy-scan` label is added after the PR is opened.

Also on this PR:
- The `--severity` option includes all of them when the trigger is PR, while the daily report only includes `HIGH` and `CRITICAL` ones. This is because low and medium issues are numerous and could be more interesting to be listed in a PR, rather than the daily report.
- Added the `--ignore-unfixed` option to reduce the size of the report table. An issue that does not have a fixed version is not very useful to be listed.
- Added `concurrency` for the PR workflow that builds notebooks to reduce redundant builds.
- Removed scan for secrets as it takes more time and it's not bringing value to the reports.
- For reviewers: I didn't touch the `pull_request: make ${{ inputs.target }}` step; I only moved it up.

## How Has This Been Tested?

I've executed the tests on my fork.

## Merge criteria:

### DIFF
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -120,37 +120,6 @@ jobs:
           IMAGE_REGISTRY: "ghcr.io/${{ github.repository }}/workbench-images"
           CONTAINER_BUILD_CACHE_ARGS: "--cache-from ${{ env.CACHE }} --cache-to ${{ env.CACHE }}"
 
-      - name: "schedule: run Trivy vulnerability scanner"
-        if: "${{ fromJson(inputs.github).event_name == 'schedule' }}"
-        run: |
-          TRIVY_VERSION=0.53.0
-          REPORT_FOLDER=${{ github.workspace }}/report
-          REPORT_FILE=trivy-report.md
-          REPORT_TEMPLATE=trivy-markdown.tpl
-
-          mkdir -p $REPORT_FOLDER
-          cp ci/$REPORT_TEMPLATE $REPORT_FOLDER
-
-          IMAGE_NAME=ghcr.io/${{ github.repository }}/workbench-images:${{ inputs.target }}-${{ github.ref_name }}_${{ github.sha }}
-          echo "Scanning $IMAGE_NAME"
-
-          # have trivy access podman socket,
-          # https://github.com/aquasecurity/trivy/issues/580#issuecomment-666423279
-          podman run --rm \
-              -v ${PODMAN_SOCK}:/var/run/podman/podman.sock \
-              -v ${REPORT_FOLDER}:/report \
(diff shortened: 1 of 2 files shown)


.github/workflows/build-notebooks-TEMPLATE.yaml
```diff
@@ -143,28 +183,13 @@ jobs:
                 image \
                 --image-src podman \
                 --podman-host /var/run/podman/podman.sock \
-                --scanners vuln,secret \
+                --scanners vuln,secret --ignore-unfixed \
```
jiridanek, line 186: Have we already discussed not running the `secret` scanner for PRs? I think not yet. That's the main part that's taking up time, as trivy itself helpfully points out in logs.

ci/cached-builds/gen_gha_matrix_jobs.py
```diff
@@ -108,6 +108,10 @@ def write_github_workflow_file(tree: dict[str, list[str]], path: pathlib.Path) -
             "workflow_dispatch": {},
             "schedule": [{ "cron": "0 2 * * *"}], # 2am UTC everyday
         },
+        "concurrency": {
+            "group": "${{ format('build-notebooks-{0}', github.sha) }}",
```
jiridanek, line 112: When does it happen that there is more than one run at the same time with the same sha? Sounds unlikely to me.

caponetto: I can think of two scenarios: 1) subsequent `workflow_dispatch` and 2) when a `schedule` starts after a running `workflow_dispatch` or vice-versa.

.github/workflows/build-notebooks-TEMPLATE.yaml
caponetto: I almost removed `secret` on this PR because yeah it takes more time and I haven't seen any issue related to that. But then thought that we could miss issues if we remove it. Anyway, I'm +1 for removing it.

ci/cached-builds/gen_gha_matrix_jobs.py
caponetto: On second thought, these scenarios should be rare and I don't want the `schedule` to be canceled if a `workflow_dispatch` starts right after. I'm gonna revert that.

jiridanek (approved):

jiridanek:
@atheo89 can  you please create a `trivy-scan` label in this project, and give it some nice color?

jstourac (approved):

PR #570, opened 2024-06-18
[2023b] RHOAIENG-8740 Fix spawn-fcgi-1.6.3-23.fc37.x86_64.rpm download location (#569)

jiridanek:
This is a manual backport for the files
missing in 2024a for the changes done in [RHOAIENG-8255](https://issues.redhat.com//browse/RHOAIENG-8255)

* https://github.com/opendatahub-io/notebooks/pull/569

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #616, opened 2024-07-12
Update trusty from 0.3 to 0.6 to fix vulnerabilities

atheo89:
Related to: https://issues.redhat.com/browse/RHOAIENG-6457
## Description
This PR updates also the 2023a image version is order to get the vulnerability of PyArrow fixed on rhoai 2.8. This should flow to downstream 2023a and then update the manifests of rhoai2.8

## Merge criteria:

### DIFF
--- a/jupyter/trustyai/ubi9-python-3.9/Pipfile
+++ b/jupyter/trustyai/ubi9-python-3.9/Pipfile
@@ -7,7 +7,7 @@ name = "pypi"
 
 [packages]
 # TrustyAI and useful extensions
-trustyai = "~=0.3.0"
+trustyai = "~=0.6.0"
 
 # Datascience and useful extensions
 elyra-pipeline-editor-extension = "~=3.15.0"
--- a/jupyter/trustyai/ubi9-python-3.9/Pipfile.lock
+++ b/jupyter/trustyai/ubi9-python-3.9/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "14c05b52434ecf9ffa6dd8a8557d925f4f69fa29fc528de14f8bed9962d27680"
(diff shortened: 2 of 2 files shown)


jiridanek:
/lgtm

would've been nice to see trivy again on this, but it's just backport, so I'm willing to believe

atheo89:
Yes, It would be nice to open a separated dummy PR and check everything. Since this is backport i will move to it

jstourac (approved):
LGTM in general.

Two points for consideration, but we can probably waive them both:
1. Shall we also update [manifests](https://github.com/opendatahub-io/notebooks/blob/2023a/manifests/base/jupyter-trustyai-notebook-imagestream.yaml#L18) for this? Truth is that these aren't used for the release in any way, so probably only advantage of doing this (apart from trying to keep consistent) is to have it as a reminder when applying this change into the actual release branch.
2. Here, since we're updating the library, we could consider to do the minimal TrustyAI package upgrade (thus to version 0.4.0 only instead of to the latest 0.6.0) as [discussed in this comment](https://github.com/red-hat-data-services/notebooks/pull/317#discussion_r1674246600). But probably not worth it since we have 0.6.0 in other branches already and we're happy with it.

atheo89:
Thanks Jan, about the first point, it doesn't really need, since from the YYYYx branches we only build the notebooks, the difference comes in place on the manifest files on the rhoai x.y branches.  About the second, yes since now we have already go to 0.6 let's keep an homogeneity and keep this consideration for future upgrades.

atheo89:
/override ci/prow/images  
These images are failing because we don't have backport this https://github.com/opendatahub-io/notebooks/pull/570

jstourac:
> /override ci/prow/images These images are failing because we don't have backport this #570

I created this to track it https://issues.redhat.com/browse/RHOAIENG-9801.

jstourac:
Since this is merged now, we need to not forget to update manifests together with the images updates in the rhoai-2.8 branch.

PR #620, opened 2024-07-15
Switch base image to UBI for AMD rocm install

harshad16:
## Description

Switch base image to UBI for AMD rocm install
Related-to: https://issues.redhat.com/browse/RHOAIENG-7501

## How Has This Been Tested?

```
1. build the base `podman build -t amd-base .`
2. check the necessary bits are available in amd-base , with `rpm -qa`
```

## Merge criteria:


jstourac:
Looks like we need to update the prow configuration then too:
* https://github.com/opendatahub-io/notebooks/actions/runs/9943091202/job/27466099229?pr=620
* https://github.com/opendatahub-io/notebooks/actions/runs/9943091202/job/27466099489?pr=620

jstourac (approved):

amd/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+USER 0
+WORKDIR /opt/app-root/bin
+
+ARG ROCM_VERSION=6.1
+ARG AMDGPU_VERSION=6.1
+
+# Enable epel-release repositories
+# TODO: Remove the CentOS Stream repos once the ROCm dependency are in UBI
+RUN yum -y install https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/centos-stream-repos-9.0-26.el9.noarch.rpm  \
```
caponetto, line 23: Is there any particular reason for not combining sequential `RUN` commands into a single one? I understand each `RUN` creates one extra layer, which extends build time and increases the container image size.

jiridanek: If your `yum install` installs new files into `/etc/yum.repos.d/`, then there need to be multiple yum invocations, one to install the new .repo files, and the other one to use the repos. Still, it could still be done in a single Dockerfile `RUN`.

If the only reason for multiple RUNs is readability, would it make sense to put all these RUNs into a single bash script and run that from the Dockerfile? Another reason may be faster and simpler local development that makes use of layer caching.

This image is IMO ok, when there are multiple `yum installs` in some other images where each downloads the yum cache, deletes it and the end, and the yum running right after downloads cache again, that's something to fix!

jiridanek:
Also Makefile should be updated (in this PR), to reference ubi9- and not c9s- dir.

amd/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+    https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-9.0-26.el9.noarch.rpm \
+    https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \  
+    yum config-manager --set-enabled crb && \
+    yum clean all && rm -rf /var/cache/yum
+
+
+# Install the ROCm rpms
+RUN echo "[ROCm]" > /etc/yum.repos.d/rocm.repo && \
```
atheo89, line 31: Just a suggestion if this makes sense to you:

Could we place the ROCm and AMDGPU RPMs in external files and then use the COPY  to include them in the Dockerfile?

amd/ubi9-python-3.9/Dockerfile
```diff
@@ -0,0 +1,49 @@
+# Install ROCm AMD from:
```
atheo89, line 1: As mentioned we need also this on the Makefile:

```
.PHONY: amd-ubi9-python-3.9
amd-ubi9-python-3.9: base-ubi9-python-3.9
	$(call image,$@,amd/ubi9-python-3.9,$<)
```

atheo89 (commented):
Great transition to UBI :slightly_smiling_face: 
I added some minor suggestions, overall looks good to me.

I was able to built this image you may find it here:
`quay.io/rh_ee_atheodor/workbench-images:amd-ubi9-python-3.9-2024a_20240717`

amd/ubi9-python-3.9/Dockerfile
harshad16: >Is there any particular reason for not combining sequential RUN commands into a single one? I understand each RUN creates one extra layer, which extends build time and increases the container image size.

Rightly pointed out, corrected it to the single RUN command

jiridanek:
We're still running out of disk space on one of the images, as @caponetto observed yesterday.

jiridanek (approved):

jiridanek:
>  ci/prow/notebook-amd-c9s-python-3-9-pr-image-mirror — Job failed.                     BaseSHA:fecd10c13ce13d66b57498abc48976f74121dd63

so openshift-ci needs to be updated

jiridanek:
@jstourac so, approving in GitHub UI gives BOTH LGTM and approved, now that we are approvers. Gotta be careful.

Makefile
```diff
@@ -267,30 +267,30 @@ rstudio-c9s-python-3.9: base-c9s-python-3.9
...
-amd-c9s-python-3.9: base-c9s-python-3.9	
-	$(call image,$@,amd/c9s-python-3.9,$<)
+####################################### Buildchain for AMD Python 3.9 using UBI9 #######################################
+.PHONY: amd-ubi9-python-3.9
+amd-ubi9-python-3.9: base-ubi9-python-3.9
+	$(call image,$@,amd/ubi9-python-3.9,$<)
 
-# We are only using c9s base image here onwards, 
+# We are only using ubi9 base image here onwards, 
 # DON'T confuse due to the ubi9 mention, it just directory name.
```
atheo89, line 276: Maybe we need to remove this comment now, to avoid confusion.

caponetto:
> We're still running out of disk space on one of the images, as @caponetto observed yesterday.

At first, I thought that was happening only during Trivy scan because it has to copy stuff around. But then I saw that CI is running out of disk space even during the build step for `amd-jupyter-pytorch-ubi9-python-3.9` (examples are this PR and some builds triggered after a merge). Considering that `amd-jupyter-pytorch-ubi9-python-3.9` is ~60 GB uncompressed, the CI is operating on its limit. If we ever need to add new things to this image, we'll probably face these storage issues more often.

atheo89:
To fix the ci please check this PR:  https://github.com/opendatahub-io/notebooks/pull/627 explain the instructions on how you can update the notebook matrix

ci/cached-builds/gen_gha_matrix_jobs.py
```diff
@@ -198,8 +198,8 @@ def test_select_changed_targets(self):
         changed_files = ["jupyter/datascience/ubi9-python-3.9/Dockerfile"]
 
         leafs = gha_pr_changed_files.filter_out_unchanged(leafs, changed_files)
-        assert set(leafs) == {'amd-jupyter-pytorch-c9s-python-3.9',
-                              'amd-jupyter-tensorflow-c9s-python-3.9',
+        assert set(leafs) == {'rocm-jupyter-pytorch-c9s-python-3.9',
+                              'rocm-jupyter-tensorflow-c9s-python-3.9',
```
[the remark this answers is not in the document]
jiridanek: @atheo89 I suggest we merge it without making the change, and I can do this later, what do you think?

This test does not run in the CI, so we won't spoil anything. And it would be good to start merging the rocm work soon, because the time is running short.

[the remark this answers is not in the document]
jiridanek: When this and the other groundwork gets merged, we can seriously try to decrease image size by removing the unneeded parts of the sdk that are not required to run pytorch/tensorflow. It would be best not to try to release with full sdk, these huge images are too hard to work with.

atheo89:
/lgtm 
/approve

/override ci/prow/images 
/override /ci/prow/notebook-rocm-jupyter-pyt-ubi9-python-3-9-pr-image-mirror 
/override ci/prow/notebook-rocm-jupyter-tf-ubi9-python-3-9-pr-image-mirror
/override ci/prow/rocm-notebooks-e2e-tests

jiridanek:
let me try

/override "build (rocm-jupyter-pytorch-ubi9-python-3.9) / build"
/override ci/prow/notebook-rocm-jupyter-tf-ubi9-python-3-9-pr-image-mirror
/override ci/prow/rocm-notebooks-e2e-tests
/override ci/prow/notebook-rocm-jupyter-pyt-ubi9-python-3-9-pr-image-mirror
/override ci/prow/images

jiridanek:
/override "build (rocm-jupyter-pytorch-ubi9-python-3.9) / build"

PR #627, opened 2024-07-19
feat(ci): add helpful error when check-generated-code gha workflow fails

jiridanek:
## Merge criteria:


.github/workflows/code-quality.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
           if [[ -z "$clean" ]]; then
               echo "Empty git status --porcelain: $clean"
           else
+              echo "::error::Please run the command from the previous step (Rerun all code generators...) and commit the result."
```
jstourac, line 26: Please, instead of telling to run command from previous step, maybe we should explicitly say what to run.

jiridanek: Ok, then, let me change this.

jiridanek: done

.github/workflows/code-quality.yaml
```diff
@@ -15,14 +15,15 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Rerun all code generators we have
-        run: python3 ci/cached-builds/gen_gha_matrix_jobs.py
+        run: bash ci/generate_code.bash
```
jstourac, line 18: I don't like the `.bash` suffix :P , I can see we have it in one file already `./ci/cached-builds/gha_lvm_overlay.bash`... still, do we want to extend this policy?

ci/generate_code.bash
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+if ! python3 ci/cached-builds/gen_gha_matrix_jobs.py; then
+  # this returns nonzero when running outside of github; detect and print explanation
+  if [[ ${GITHUB_ACTIONS:-} == true ]]; then exit 1
```
jstourac, line 6: It's a pitty that the script fails when non-run in github... I consider this solution as a workaround that will be fixed and won't be needed in the near future; we may miss other reasons why the script failed when run locally with this approach.

.github/workflows/code-quality.yaml
jiridanek: it's only tested with bash; when I get the mac, I will also test it on zsh

ci/generate_code.bash
jiridanek: Yeah, it's not a good job on my part.

.github/workflows/code-quality.yaml
jstourac: What does the file suffix have to do with the which interpreter is used?

ci/generate_code.bash
jiridanek: I was not too had to fix, did it rn

jstourac:
Thank you, I like this much more now!

/lgtm

jiridanek:
/override ci/prow/runtime-rocm-pytorch-ubi9-python-3-9-pr-image-mirror
/override ci/prow/amd-runtimes-ubi9-e2e-tests
/override ci/prow/images

one rocm run out of ephemeral storage, and tests hit the chicken-and-egg problem from https://github.com/openshift/release/pull/54567

PR #637, opened 2024-07-24
chore(ci): Run Trivy scan `fs` instead of `image` for rocm+pytorch image due to resource constraints

caponetto:
## Description

Run Trivy scan `fs` instead of `image` for rocm+pytorch image due to resource constraints. At least, the daily report and PR will pass and will show the results for the lock file while we don't reduce the image size.

## How Has This Been Tested?

Run the updated workflow for the pytorch image and another one to check both behaviors (image and fs)

## Merge criteria:


jiridanek:
I don't like how complicated is the ci script becoming. So many variables. Let's see first if the grouping of run commands can make trivy image scan run?

Maybe there's still something else to shrink the images they can be tried?

caponetto:
TBH I don't like to see a red CI every day. So I'd rather have this in while we progress with the image optimizations.

jiridanek:
/lgtm
i'm not happy, but if we want to have passing trivy, this is the way

PR #641, opened 2024-07-24
feat(Dockerfiles): switch from s2i python images to plain ubi/cs9 ones

jiridanek:
## Description

The main benefit is size and cve exposure, as the python images come with packages we don't use; python and pip is enough for us.

Additionally, using plain ubi makes things more explicit.

## How Has This Been Tested?

I'll test this when making this change is planned into a sprint.

## Merge criteria:

### DIFF
--- a/base/ubi8-python-3.8/Dockerfile
+++ b/base/ubi8-python-3.8/Dockerfile
@@ -5,14 +5,16 @@ FROM registry.access.redhat.com/ubi8/ubi:latest
 ENV APP_ROOT="/opt/app-root"
 ENV VIRTUAL_ENV="${APP_ROOT}"
 ENV PATH="${VIRTUAL_ENV}/bin:${PATH}"
-RUN mkdir --parents --mode 0771 "${APP_ROOT}/src" && chown --recursive 1001:0 ${APP_ROOT} && \
+ENV PYTHON_VERSION=3.8
+ENV PIP_NO_CACHE_DIR=off
+RUN mkdir --parents "${APP_ROOT}/src" && chmod --recursive 0771 ${APP_ROOT} && chown --recursive 1001:0 ${APP_ROOT} && \
     useradd --uid 1001 --gid 0 --no-create-home --home-dir "${APP_ROOT}/src" \
       --comment "Default Application User" --shell /bin/bash default && \
     dnf install -y python38-pip && dnf clean all && rm -rf /var/cache/yum/*
 COPY --from=quay.io/sclorg/s2i-core-c9s:latest /usr/bin/fix-permissions /usr/bin/fix-permissions
 
 USER 1001
-RUN python3.8 -m venv "${VIRTUAL_ENV}"
+RUN python${PYTHON_VERSION} -m venv "${VIRTUAL_ENV}"
 
 LABEL name="odh-notebook-base-ubi8-python-3.8" \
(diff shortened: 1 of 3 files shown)


base/ubi8-python-3.8/Dockerfile
```diff
@@ -1,4 +1,21 @@
-FROM registry.access.redhat.com/ubi8/python-38:latest
+FROM registry.access.redhat.com/ubi8/ubi:latest
```
caponetto, line 1: Can we use `ubi-minimal` instead?

jiridanek: That would either mean living with microdnf, or installing regular dnf first thing, at which points the benefits or the different ubi base are pretty much lost, me thinks.

In any case, I'd want to do such large changes in multiple steps, across multiple PRs.

So far, I don't even have the modest change in this PR groomed. It's necessary to move slowly and cautiously, and to build consensus.

The disk savings are small, 300mb across three base images.

harshad16 (changes requested):
/hold

@jiridanek can you share there is JIRA or github issue linked for this work.

This work has significant impact on the overall product, without changing the architecture , this change would disrupt some flow i believe.

the s2i image bring working dir: /opt/app-root/src
i don't think ubi:latest, has that working dir, 
this might work only because, the PVC gets mounted at the path, 
As mount would create the path. 
however we should thoroughly check these changes.

jiridanek:
> This PR is great, definitely will be good to have however

Great! In that case I'll go forward creating a Jira so this can be groomed. https://issues.redhat.com/browse/RHOAIENG-10350

jiridanek:
> the s2i image bring working dir: /opt/app-root/src

our images explicitly set this in the `WORKDIR` instruction, so we're good in that respect

> however we should thoroughly check these changes.

sure, that's why I created a jira for doing all the work in organized manner; the disk space savings are not all that significant, but I like how this removes much of the s2i magic, so that's what makes this seem worth doing to me

base/ubi8-python-3.8/Dockerfile
jiridanek: /retest

jstourac:
BTW, I was wondering what are the most important changes, so I prepared some basic script for the image comparison (I have a plan to improve it further and put it into our CI/GHA for our convenience). I tried to compare one of the base images you change here with what we have (note that your build is 2 weeks old, I took the quay image from today):

1. checkout this branch https://github.com/opendatahub-io/notebooks/compare/main...jstourac:notebooks:compareImages
2. run this:
```
./ci/compare-images.sh quay.io/opendatahub/workbench-images@sha256:e92bf20e127e545bdf56887903dc72ad227082b8bc23f45ff4f0fc67e6430318 ghcr.io/jiridanek/notebooks/workbench-images:base-ubi9-python-3.9-jd_ubi_base_adedd4a943977ecdcb67bc6eb9eda572d10c3ddc
```

atheo89:
+1 for this update, I like this. :slightly_smiling_face: As Harshad mentioned, we should thoroughly test everything to ensure there are no issues with the PVCs. Otherwise, it /lgtm

jiridanek:
> we should thoroughly test everything

yeah; was just thinking whether nginx gets properly installed after this change; I did not think of it before

jiridanek:
@coderabbitai review

PR #652, opened 2024-07-31
RHOAIENG-10783: fix(rocm): de-vendor the bundled rocm libraries from pytorch

jiridanek:
* [RHOAIENG-10783](https://issues.redhat.com/browse/RHOAIENG-10783) Optimize ROCm images to reduce the size so they can be built on OpenShift CI

Use the script from instructlab to remove the duplicate copy of rocm libs from the image. This will make the image significantly smaller.

Script lives at https://github.com/tiran/instructlab-containers/blob/main/containers/rocm/de-vendor-torch.sh

## Description

This gives me

```
rocm-jupyter-pytorch-ubi9-python-3.9-2024a_20240731               ab8eaa47fff8  34 seconds ago  35.7 GB
```

While on master we currently have

```
rocm-jupyter-pytorch-ubi9-python-3.9-main_37df13c29fdde3fa4f8e455d30b5bf39d80f6dfb      9a4a772dbee7  4 days ago   46.4 GB
```

This way the torch image size is in line with tensorflow image size

```
rocm-jupyter-tensorflow-ubi9-python-3.9-main_37df13c29fdde3fa4f8e455d30b5bf39d80f6dfb   465069df6467  4 days ago   35.5 GB
```

In conclusion, 10 GB was saved.

## How Has This Been Tested?
```
CONTAINER_BUILD_CACHE_ARGS="--platform=linux/amd64  --cache-from ghcr.io/jiridanek/notebooks/workbench-images/build-cache" make rocm-jupyter-pytorch-ubi9-python-3.9
```

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #653, opened 2024-07-31
RHOAIENG-10783: fix(rocm): remove more files that instructlab also removes

jiridanek:
* [RHOAIENG-10783](https://issues.redhat.com/browse/RHOAIENG-10783) Optimize ROCm images to reduce the size so they can be built on OpenShift CI

## Description

Turns out we can remove the llvm installation if we do it forcibly, without removing dependencies. Additionally, there are gfx files for all supported cards, so since we support less, we can remove many.

See https://github.com/tiran/instructlab-containers/blob/main/containers/rocm/Containerfile.c9s#L47

## How Has This Been Tested?

* https://github.com/opendatahub-io/notebooks/pull/652

before 
```
rocm-ubi9-python-3.9-main_37df13c29fdde3fa4f8e455d30b5bf39d80f6dfb  1f64a39a93d7  7 days ago   27.8 GB
```

after
```
rocm-ubi9-python-3.9-99b24d8574bffe38c4c480007a0cf69f9eeb48ce                     c7dfe8bee592  3 minutes ago   23 GB
```

## Merge criteria:

### DIFF
--- a/rocm/ubi9-python-3.9/Dockerfile
+++ b/rocm/ubi9-python-3.9/Dockerfile
@@ -19,6 +19,9 @@ WORKDIR /opt/app-root/bin
 ARG ROCM_VERSION=6.1
 ARG AMDGPU_VERSION=6.1
 
+# default: same targets and ROCm version as https://github.com/tiran/instructlab-containers/blob/main/containers/rocm/Containerfile.c9s#L47
+ARG AMDGPU_TARGETS=gfx900;gfx906:xnack-;gfx908:xnack-;gfx90a:xnack-;gfx90a:xnack+;gfx942;gfx1030;gfx1100
+
 # Enable epel-release repositories
 
 # Install the ROCm rpms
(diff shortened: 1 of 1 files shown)


jstourac:
Looks good, but we should really assure that this isn't missing in the final image for any use-case we should support.

rocm/ubi9-python-3.9/Dockerfile
```diff
@@ -18,6 +18,9 @@ WORKDIR /opt/app-root/bin
 ARG ROCM_VERSION=6.1
 ARG AMDGPU_VERSION=6.1
 
+# default: same targets and ROCm version as https://github.com/tiran/instructlab-containers/blob/main/containers/rocm/Containerfile.c9s#L47
+ARG AMDGPU_TARGETS=gfx900;gfx906:xnack-;gfx908:xnack-;gfx90a:xnack-;gfx90a:xnack+;gfx942;gfx1030;gfx1100
```
jstourac, line 22: Can we get some official set of AMD GPUs we should/need to support? I mean - are we sure that this default is what we want?

jiridanek: > Looks good, but we should really assure that this isn't missing in the final image for any use-case we should support.

It's much better to miss something and add it later, than trying to remove things afterwards when we actually have users, imo.

harshad16: Plan is to support: AMD MI210, MI300 and later on MI250
Based on the strat we have https://issues.redhat.com/browse/RHOAISTRAT-135.

i agree, we can add the details if the opportunity presents, and try this out for now.
however what are these gfx files, are we aware what functionality is this removing ?

jiridanek:
@harshad16 The `gfx` files are card-specific compiled code. The filenames end in `.bc`, `co` and `. hsaco`

> There is at least 3 types of device-side (aka GPU) binary files -
> * .bc for bitcode,
> * .hsaco for HSA code object and
> * .co for code object.
> ([source](https://groups.google.com/g/linux.debian.devel.mentors/c/HULn-__74kI))

> NVCC and HIP-Clang target different architectures and use different code object formats: NVCC is `cubin` or `ptx` files, while the HIP-Clang path is the `hsaco` format. ([source](https://github.com/ROCm/HIP/blob/6b5b7c943848c7f96d3fe247f4a405739c6b974a/docs/how-to/hip_porting_driver_api.md?plain=1#L61))

harshad16:
@jiridanek , with this https://github.com/opendatahub-io/notebooks/pull/673 getting merged
do we want to invest time in getting to test the additional changes made here ?

jiridanek:
not planned for now

PR #665, opened 2024-08-07
RHOAIENG-10972: Create a script to automate library upgrades

paulovmr:
## Description
Created the scripts folder with its first script, the `update_library_version.sh`. This script updates the version of a specified library in Pipfile and requirements-elyra.txt files, only if the new version is higher. It can optionally run `pipenv lock` after updating the version.

## How Has This Been Tested?
Manual executions, comparing the results with merged library upgrades PRs.

## Merge criteria:


scripts/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Parameters
+
+- `directory`: The root directory to start searching for Pipfile and requirements-elyra.txt files.
+- `library_name`: The name of the library to update.
+- `new_version`: The new version to set for the library.
+- `include_paths_with`: A pipe-separated list of substrings; only files in directories containing at least one of these substrings will be processed.
+- `exclude_paths_with`: A pipe-separated list of substrings; files in directories containing any of these substrings will be excluded.
+- `lock_files`: Whether to run `pipenv lock` after updating the library version (`true` or `false`).
+
```
jiridanek, line 21:
```suggestion

```

I'd not put it here, it's enough to have it on top of the script, IMO. Fewer things to keep in sync. The Examples is what matters.

scripts/update_library_version.sh
```diff
@@ -0,0 +1,201 @@
...
+  for ((i = 0; i < ${#ver1_parts[@]} || i < ${#ver2_parts[@]}; i++)); do
+    # Use 0 if a part is missing
+    v1=${ver1_parts[i]:-0}
+    v2=${ver2_parts[i]:-0}
+    
+    # Compare the parts
+    if ((v1 > v2)); then
+      return 1
+    elif ((v1 < v2)); then
+      return -1
```
jstourac, line 66: This will effectively be 255 since you can't return negative values in bash IMO. So maybe let's use 255 directly to avoid possible confusion?

paulovmr: @jstourac good catch, I changed the function using your suggestion.

scripts/update_library_version.sh
```diff
@@ -0,0 +1,205 @@
+#!/bin/bash
+
+# Check if the script is running on Linux
+if [[ "$OSTYPE" != "linux-gnu"* ]]; then
+  echo "Error: This script can only be run on Linux."
```
caponetto, line 5: One thing we can consider in the future is to run this script in a container to make it independent of the host environment.

PR #669, opened 2024-08-09
[WIP] this script allows us to compare two images

jiridanek:
## Merge criteria:

### DIFF
--- a/ci/compare-images.sh
+++ b/ci/compare-images.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+#
+# This script serves to compare two docker images using skopeo tool. This gives
+# a brief information regarding the following image differences:
+#   - size
+#   - architecture
+#   - operating system
+#   - config
+#     - default user
+#     - exposed ports
+#     - environment variables
+#     - entrypoint
+#     - working directory
+#     - labels
+#   - Python packages
+#   - RPM packages
+#
+# It uses the skopeo TODO downloads images locally...
+#
+# Local execution: ./ci/compare-images.sh <image-1> <image-2>
+#   Note: <image-*> is in the format <repository@sha256:SHA>
+#
+# Example usage:
+#    ./ci/compare-images.sh quay.io/opendatahub/workbench-images@sha256:e92bf20e127e545bdf56887903dc72ad227082b8bc23f45ff4f0fc67e6430318 ghcr.io/jiridanek/notebooks/workbench-images:base-ubi9-python-3.9-jd_ubi_base_adedd4a943977ecdcb67bc6eb9eda572d10c3ddc
+
+shopt -s globstar
+
+
+function gather_metadata() {
+    local image="${1}"
+    local tmp_dir="${2}"
+
+    local ret_code=0
+
+    echo "Gathering the metadata for the image: '${image}'"
+
+    local image_sha
+    image_sha=$(echo "${image}" | cut -d ':' -f2)
+    echo "Image SHA: '${image_sha}'"
+
+    # Get image size
+    skopeo inspect --raw "docker://${image}" | jq '[ .layers[].size ] | add' > "${tmp_dir}/${image_sha}-size.txt"
+
+    # Get image metadata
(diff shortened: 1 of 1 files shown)


ci/compare-images.sh
```diff
@@ -0,0 +1,109 @@
...
+#
+# It uses the skopeo TODO downloads images locally...
+#
+# Local execution: ./ci/compare-images.sh <image-1> <image-2>
+#   Note: <image-*> is in the format <repository@sha256:SHA>
+#
+# Example usage:
+#    ./ci/compare-images.sh quay.io/opendatahub/workbench-images@sha256:e92bf20e127e545bdf56887903dc72ad227082b8bc23f45ff4f0fc67e6430318 ghcr.io/jiridanek/notebooks/workbench-images:base-ubi9-python-3.9-jd_ubi_base_adedd4a943977ecdcb67bc6eb9eda572d10c3ddc
+
+shopt -s globstar
```
jiridanek, line 26: What about

```
set -Eeuxo pipefail
```

https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

to fail if something needed is missing?

I happen to not have skopeo, so

```
./ci/compare-images.sh quay.io/opendatahub/workbench-images@sha256:e92bf20e127e545bdf56887903dc72ad227082b8bc23f45ff4f0fc67e6430318 ghcr.io/jiridanek/notebooks/workbench-images:base-ubi9-python-3.9-jd_ubi_base_adedd4a943977ecdcb67bc6eb9eda572d10c3ddc
./ci/compare-images.sh: line 26: shopt: globstar: invalid shell option name
Gathering the metadata for the image: 'quay.io/opendatahub/workbench-images@sha256:e92bf20e127e545bdf56887903dc72ad227082b8bc23f45ff4f0fc67e6430318'
Image SHA: 'e92bf20e127e545bdf56887903dc72ad227082b8bc23f45ff4f0fc67e6430318'
./ci/compare-images.sh: line 42: skopeo: command not found
./ci/compare-images.sh: line 45: skopeo: command not found
```

ci/compare-images.sh
jiridanek, line 1: I like this. We should (eventually) have the results of this available everytime we update metadata in release-2.8 etc. release branch of images, also everytime we update 2024a etc., and for every PR that makes changes to images.

ci/compare-images.sh
```diff
@@ -0,0 +1,109 @@
...
+    echo "Metadata for image '${image}' gathered."
+}
+
+function compare_metadata() {
+    local tmp_dir="${1}"
+
+    echo "Let's compare the image metadata now:"
+
+    diff -y "${tmp_dir}"/*-size.txt
+    diff -y "${tmp_dir}"/*-metadata.txt
```
jiridanek, line 68: This is essentially a json, or it can be printed as json, so the diff should use that

```
diff <(jq --sort-keys . A.json) <(jq --sort-keys . B.json)
```

some people on stackoverflow suggest `jd -set`, https://stackoverflow.com/questions/31930041/using-jq-or-alternative-command-line-tools-to-compare-json-files

ci/compare-images.sh
jstourac: yep, that is exactly my point; I'll create a tracking issue for this so I can plan this for some of next sprints then

caponetto:
Have you seen [this](https://github.com/reproducible-containers/diffoci)? Maybe it could be useful.

ci/compare-images.sh
jiridanek, line 1: > Have you seen [this](https://github.com/reproducible-containers/diffoci)? Maybe it could be useful.

Not bad, but the functionality seems different. This script compares packages, whereas the linked tool compares files in the images. Guess it could be added to the script, but then it would require us to have the tool preinstalled.

jstourac: @caponetto Actually, I tried that tool before I started to implement this my own thing. I didn't like the output of that script much. It's too verbose and shows complete differences of the image on the file level instead of the package level. I'm not sure whether this is what we want. That is why I went with our own way.

caponetto: Cool! I'm glad you've already considered it.

jiridanek:
I'm not going to be working on this, so best to close my PR.

PR #671, opened 2024-08-13
RHOAIENG-11300: Improve the library upgrade script for alternative sources

paulovmr:
## Description

Modify the library upgrade script to also work for libraries that have a alternative source defined in the Pipfile.

Currently, dependencies with the most used format are supported, for example: `tensorboard = "~=2.16.2"`. The dependencies that define both version and alternative source, for example: `torch = {version = "~=2.4.0", index = "pytorch"}` should also be supported.

## How Has This Been Tested?

Manual executions, comparing the results with merged library upgrades PRs.

## Merge criteria:


jstourac:
Some examples in the comments for those regexps would make this a bit better to read, though :slightly_smiling_face: 

/lgtm

PR #659, opened 2024-08-01
RHOAIENG-4533: Add images based on Python 3.11

caponetto:
https://issues.redhat.com/browse/RHOAIENG-4533

## Description

Add images based on Python 3.11

## How Has This Been Tested?

Notebook tests plus manual sanity checks on some images. We can handle issues separately if we find them afterward.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #672, opened 2024-08-13
RHOAIENG-11216: Add script to automate the creation of a new Python-based image from an existing one

caponetto:
https://issues.redhat.com/browse/RHOAIENG-11216
## Description

Adding a Python script that generates a new folder structure for a Python-based project by copying an existing one and updating Python version references. It performs the following steps:

1. Copy Folder: Duplicates the specified folder structure.
1. Update Python Versions: Replaces occurrences of the source Python version with the target version throughout the folder.
1. Update Lock Files: Executes `pipenv lock` to regenerate lock files based on the new Python version.

If `pipenv lock` encounters errors, manual intervention is required to resolve compatibility issues between dependencies. Review the errors reported by `pipenv` and adjust the dependencies as needed.

See the `README` file for examples.

## How Has This Been Tested?

Manual executions, comparing the results with #659.

## Merge criteria:

### DIFF
--- a/scripts/new_python_based_image.py
+++ b/scripts/new_python_based_image.py
@@ -6,6 +6,7 @@ import platform
 import subprocess
 import logging
 import re
+from dataclasses import dataclass
 
 
 LOGGER = logging.getLogger(__name__)
@@ -27,24 +28,16 @@ def configure_logger(log_level: str):
     LOGGER.setLevel(log_level)
 
 
+@dataclass
 class Args:
     """
     Class to encapsulate command-line arguments.
-
-    Attributes:
-        context_dir: The directory to use as the context for searching.
-        source: The source Python version.
-        target: The target Python version.
-        match: The string to match with the paths.
-        log_level: The logging level.
     """
-
-    def __init__(self, context_dir: str, source: str, target: str, match: str, log_level: str):
-        self.context_dir = context_dir
-        self.source = source
-        self.target = target
-        self.match = match
-        self.log_level = log_level
+    context_dir: str
+    source: str
+    target: str
+    match: str
+    log_level: str
 
     def __str__(self):
         return (f"Arguments:\n"
--- a/scripts/new_python_based_image.py
+++ b/scripts/new_python_based_image.py
@@ -11,12 +11,12 @@ import re
 LOGGER = logging.getLogger(__name__)
 
 
-def configure_logger(log_level):
+def configure_logger(log_level: str):
     """
     Configures the logging settings based on the provided log level.
 
     Args:
-        log_level (str): The logging level to set (e.g., 'INFO', 'DEBUG').
+        log_level: The logging level to set (e.g., 'INFO', 'DEBUG').
     """
     level = getattr(logging, log_level.upper(), logging.INFO)
     logging.basicConfig(
@@ -32,14 +32,14 @@ class Args:
     Class to encapsulate command-line arguments.
 
     Attributes:
-        context_dir (str): The directory to use as the context for searching.
-        source (str): The source Python version.
-        target (str): The target Python version.
-        match (str): The string to match with the paths.
-        log_level (str): The logging level.
+        context_dir: The directory to use as the context for searching.
+        source: The source Python version.
+        target: The target Python version.
+        match: The string to match with the paths.
+        log_level: The logging level.
     """
 
-    def __init__(self, context_dir, source, target, match, log_level):
+    def __init__(self, context_dir: str, source: str, target: str, match: str, log_level: str):
         self.context_dir = context_dir
         self.source = source
         self.target = target
@@ -94,12 +94,12 @@ def extract_input_args():
     return Args(args.context_dir, args.source, args.target, args.match, args.log_level)
 
 
-def extract_python_version(version):
+def extract_python_version(version: str):
     """
     Extracts the major and minor version components from a Python version string.
 
     Args:
-        version (str): The Python version string (e.g., '3.9').
+        version: The Python version string (e.g., '3.9').
 
     Returns:
         list: A list containing the major and minor version components as strings.
@@ -107,12 +107,12 @@ def extract_python_version(version):
     return version.split(".")[:2]
 
 
-def check_python_version(version):
+def check_python_version(version: str):
     """
     Validates the format of a Python version string.
 
     Args:
-        version (str): The Python version string to validate.
+        version: The Python version string to validate.
 
     Exits the program with an error if the format is invalid.
     """
@@ -122,12 +122,12 @@ def check_python_version(version):
         sys.exit(1)
 
 
-def check_target_python_version_installed(version):
+def check_target_python_version_installed(version: str):
     """
     Checks if the specified Python version is installed on the system.
 
     Args:
-        version (str): The Python version to check.
+        version: The Python version to check.
 
     Exits the program with an error if the version is not installed.
     """
@@ -137,13 +137,13 @@ def check_target_python_version_installed(version):
         sys.exit(1)
 
 
-def check_input_versions_not_equal(source_version, target_version):
+def check_input_versions_not_equal(source_version: str, target_version: str):
     """
     Ensures that the source and target Python versions are different.
 
     Args:
-        source_version (str): The source Python version.
-        target_version (str): The target Python version.
+        source_version: The source Python version.
+        target_version: The target Python version.
 
     Exits the program with an error if the versions are the same.
     """
@@ -176,12 +176,12 @@ def check_pipenv_installed():
         sys.exit(1)
 
 
-def check_requirements(args):
+def check_requirements(args: Args):
     """
     Performs various checks to ensure that all requirements are met.
 
     Args:
-        args (Args): An instance of the Args class containing the command-line arguments.
+        args: An instance of the Args class containing the command-line arguments.
     """
     check_os_linux()
     check_python_version(args.source)
@@ -191,14 +191,14 @@ def check_requirements(args):
     check_pipenv_installed()
 
 
-def find_matching_paths(context_dir, source_version, match):
+def find_matching_paths(context_dir: str, source_version: str, match: str):
     """
     Finds directories in the context directory that match the specified source version and match criteria.
 
     Args:
-        context_dir (str): The directory to search in.
-        source_version (str): The Python version to match.
-        match (str): The string to match with the paths.
+        context_dir: The directory to search in.
+        source_version: The Python version to match.
+        match: The string to match with the paths.
 
     Returns:
         list: A list of directories that match the criteria and contain a Dockerfile.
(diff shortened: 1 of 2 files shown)


scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+    logging.basicConfig(
+        level=level,
+        format="[%(levelname)s] %(asctime)s: %(message)s",
+        datefmt="%H:%M:%S"
+    )
+    LOGGER.setLevel(log_level)
+
+
+class Args:
+    """
```
jiridanek, line 31: Consider making this a https://docs.python.org/3/library/dataclasses.html, it is significantly less verbose that way.

scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+    missing_args = [arg for arg, value in vars(args).items() if value is None]
+    if missing_args:
+        print(f"Missing required arguments: {', '.join(missing_args)}")
+        parser.print_help()
+        sys.exit(1)
+
+    return Args(args.context_dir, args.source, args.target, args.match, args.log_level)
+
+
+def extract_python_version(version):
```
jiridanek, line 97:
```suggestion
def extract_python_version(version: str):
```

the optional type annotations are more practical than weiting types into comments

scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+    paths = with_logged_execution(lambda: find_matching_paths(args.context_dir, args.source, args.match),
+                                  f"Finding matching paths with '{args.match}' and Python {args.source}")
+
+    paths_dict = replace_python_version_on_paths(paths,
+                                                 args.source,
+                                                 args.target)
+
+    if len(paths_dict) == 0:
+        LOGGER.info("No paths found to copy.")
+        return
```
jstourac, line 531: Shall we exit gracefully or is this rather a reason to exit with error?

scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+    LOGGER.info(
+        f"{len(success_paths)} folders have been copied successfully whereas {len(failed_paths)} failed.")
+
+    if len(success_paths) > 0:
+        with_logged_execution(lambda: process_paths(success_paths, args.source, args.target),
+                              "Processing copied folders")
+
+    if len(failed_paths) > 0:
+        LOGGER.warning(
+            f"Failed to copy the following paths:\n{list_to_str(failed_paths)}")
```
jstourac, line 548: Again, is it okay to exit successfully once there are some failed paths?

scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+
+    Args:
+        context_dir (str): The directory to search in.
+        source_version (str): The Python version to match.
+        match (str): The string to match with the paths.
+
+    Returns:
+        list: A list of directories that match the criteria and contain a Dockerfile.
+    """
+    blocklist = [os.path.join(".", path) for path in [".git", ".github", "ci"]]
```
jstourac, line 206: Maybe also:
```suggestion
    blocklist = [os.path.join(".", path) for path in [".git", ".github", "ci", "docs", "manifests", "scripts", "tests"]]
```
?

jstourac (commented):
I put some comments, but LGTM in general!

scripts/new_python_based_image.py
caponetto: I don't see it as an error if you provide inputs that lead to nothing being processed.

scripts/new_python_based_image.py
caponetto: The idea is to continue the execution with the script for the other folders if this condition is met for a particular folder. Ultimately, you can see that this path has failed and act accordingly.

scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+    try:
+        result = subprocess.run(
+            ["pipenv", "lock", "--python", target_version],
+            cwd=os.path.dirname(pipfile_path),
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            env=env
+        )
+        LOGGER.debug(result.stdout.decode())
```
jiridanek, line 488: use `subprocess.check_call` as a "shorthand" for check=True (it's not really that shorter, so just an idea)

but do set the `encoding` keyword argument (= "uft-8"); that way you don't need to call .decode() every time

scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+        "Review and make the appropriate changes in the Makefile and CI-related files.",
+        "Push the changes to a new branch on your fork to build the new images with GitHub workflows.",
+        "Test the new images."
+    ]
+
+
+def main():
+    """
+    Main function to execute the script.
+    """
```
jiridanek, line 513:
```suggestion
    """
    Main function to execute the script.
    """
```

this would be overcommenting, for sure

scripts/new_python_based_image.py
```diff
@@ -0,0 +1,555 @@
...
+        help="The Python version to base the new image from.")
+    parser.add_argument(
+        "--target",
+        help="The Python version to be used in the new image.")
+    parser.add_argument(
+        "--match",
+        help="The string to match with the paths to base the new image from.")
+    parser.add_argument(
+        "--log-level", default="INFO",
+        help="Set the logging level. Default: INFO.")
```
jiridanek, line 84:
```suggestion
        help="Set the logging level. Default: %(default)s.")
```

read all about it https://docs.python.org/3/library/argparse.html#help

caponetto:
@jiridanek Please let me know when you finish with the review.

jiridanek (commented):
Guess I'm done. My main complaint is probably that there's huge amount of comments even for things that are fairly obvious. Hopefully if you switch to using python type annotations, you'd feel less need to comment every parameter of a 3 line function, where it's equally easy to read the implementation body as to read the comment.

I would not be offended if you decide to keep the comments, but please remove the comment from main(), that's way too much ;P

scripts/new_python_based_image.py
jstourac: Hm, probably depends on point of view. In case I provide wrong input and didn't realize, the script will finish with 0. But yeah, I don't insist. Keep it that way then.

scripts/new_python_based_image.py
caponetto: Thanks for the tip with `encoding`. In this case, it's better to stick with `run`. From `check_call` docs:

```
Code needing to capture stdout or stderr should use run() instead
```

scripts/new_python_based_image.py
jstourac: Yes, I don't mean to stop the execution, but to exit with non-zero at the end... I guess that we will be running this manually, so it really doesn't matter that much, but in case of having this in some automation, then some non-zero exit code would be nice to realize that some of the paths failed to copy... otherwise one would have to grep for the warnings.

scripts/new_python_based_image.py
caponetto: Although there must be always a manual review after the script execution, I'll return non-zero code for non-happy scenarios.

jiridanek: Here's example from Python itself. Should a test run that ended up ignoring every test and not actually running any be considered a success or a failure? :P Here's it in pytest, but I also recall there was something like that in the stdlib unittest https://github.com/pytest-dev/pytest/pull/817

scripts/new_python_based_image.py
```diff
@@ -27,40 +29,32 @@ def configure_logger(log_level):
...
-        self.target = target
-        self.match = match
-        self.log_level = log_level
+    context_dir: str
+    source: str
+    target: str
+    match: str
+    log_level: str
 
     def __str__(self):
```
jiridanek, line 43: https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass generates __repr__ method for you, is __str__ still necessary?

scripts/new_python_based_image.py
```diff
@@ -81,7 +75,7 @@ def extract_input_args():
         help="The string to match with the paths to base the new image from.")
     parser.add_argument(
         "--log-level", default="INFO",
-        help="Set the logging level. Default: INFO.")
+        help="Set the logging level. Default: %(default)s).")
```
jiridanek, line 78:
```suggestion
        help="Set the logging level. Default: %(default)s.")
```

scripts/new_python_based_image.py
```diff
@@ -94,58 +88,52 @@ def extract_input_args():
     return Args(args.context_dir, args.source, args.target, args.match, args.log_level)
 
 
-def extract_python_version(version):
+def extract_python_version(version: str) -> list:
```
jiridanek, line 91:
```suggestion
def extract_python_version(version: str) -> list[str]:
```

scripts/new_python_based_image.py
```diff
@@ -191,19 +175,25 @@ def check_requirements(args):
     check_pipenv_installed()
 
 
-def find_matching_paths(context_dir, source_version, match):
+def find_matching_paths(context_dir: str, source_version: str, match: str) -> list:
```
jiridanek, line 178:
```suggestion
def find_matching_paths(context_dir: str, source_version: str, match: str) -> list[str]:
```

scripts/new_python_based_image.py
```diff
@@ -225,30 +215,30 @@ def find_matching_paths(context_dir, source_version, match):
...
 
     Returns:
-        dict: A dictionary where keys are original paths and values are modified paths with the target version.
+        A dictionary where keys are original paths and values are modified paths with the target version.
     """
     return {path: path.replace(source_version, target_version) for path in paths_list}
 
 
-def copy_paths(paths_dict):
+def copy_paths(paths_dict: dict) -> tuple:
```
jiridanek, line 233:
```suggestion
def copy_paths(paths_dict: dict) -> tuple[list[str], list[str]]:
```

scripts/new_python_based_image.py
```diff
@@ -1,22 +1,24 @@
...
+import subprocess
+import sys
+from dataclasses import dataclass
 
 
 LOGGER = logging.getLogger(__name__)
 
 
-def configure_logger(log_level):
+def configure_logger(log_level: str):
```
jiridanek, line 16:
```suggestion
def configure_logger(log_level: str) -> None:
```

scripts/new_python_based_image.py
caponetto: The output from `__repr__` is not so pretty but I can live with that. Removed.

jiridanek (commented):
Added comments about two aspects of type hints

* List, tuple, dict are parametric types, you can specify "list of what"
* When a function does not return anything, give it `-> None:` type annotation

scripts/new_python_based_image.py
caponetto: That seems redundant to me but why not?!

jiridanek: You'd get a warning when you actually attempt to return non None value, or when you try to use the return value from calling this. Guess that's pretty much it.

jstourac:
/lgtm

Thank you @caponetto, no more comments from my side :slightly_smiling_face:

jiridanek:
/lgtm

me too

caponetto:
Thanks for the review guys!
/approve

PR #682, opened 2024-08-21
[GHA] Pin kustomize version we use in our test to the particular version

jstourac:
Recently the kustomize test started to fail for us since the kustomize version 5.4 introduced some breaking changes for which we will need to update our manifests to be compatible. This work will be tracked in [1]. In the meantime, we should pin kustomize version in our GHA to the version that is used by the opendatahub-operator in runtime [2].

[1] https://issues.redhat.com/browse/RHOAIENG-11679
[2] https://github.com/red-hat-data-services/rhods-operator/blob/rhoai-2.12/go.mod#L30-L31

## Merge criteria:

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -116,16 +116,23 @@ jobs:
           echo "Hadolint done"
 
       # This simply checks that the manifests and respective kustomization.yaml finishes without an error.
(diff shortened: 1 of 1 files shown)


jiridanek:
why is ci running rocm tests? looks like the regex in openshift ci config for it is matching too much?

/lgtm

PR #709, opened 2024-09-25
chores: set the python 3.11 content for refresh-pipfile-lock cmd in Makefile

harshad16:
## Description

set the python 3.11 content for refresh-pipfile-lock cmd in Makefile

This would help with automation github action for updating the pipfile.lock.

## How Has This Been Tested?

Execute: ` make refresh-pipfilelock-files`

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -629,6 +629,22 @@ refresh-pipfilelock-files:
 	cd runtimes/tensorflow/ubi9-python-3.9 && pipenv lock
 	cd runtimes/rocm-tensorflow/ubi9-python-3.9 && pipenv lock
 	cd runtimes/rocm-pytorch/ubi9-python-3.9 && pipenv lock
+	cd base/c9s-python-3.11 && pipenv lock
+	cd base/ubi9-python-3.11 && pipenv lock
+	cd codeserver/ubi9-python-3.11 && pipenv lock
+	cd jupyter/minimal/ubi9-python-3.11 && pipenv lock
+	cd jupyter/datascience/ubi9-python-3.11 && pipenv lock
+	cd jupyter/pytorch/ubi9-python-3.11 && pipenv lock
+	cd jupyter/tensorflow/ubi9-python-3.11 && pipenv lock
+	cd jupyter/trustyai/ubi9-python-3.11 && pipenv lock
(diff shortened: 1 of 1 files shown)


atheo89:
Thanks Harshad for this fix, moreover could you inactivate/remove the scheduler from the `piplock-renewal-YYYYx.yml` [link](https://github.com/opendatahub-io/notebooks/blob/main/.github/workflows/piplock-renewal-2023a.yml) and keep only the dispatch since we are not going to that way anymore, and also we will eliminate the conflicts with the downstream syncs PRs

jstourac:
For the record - following paths are excluded from addition by design. Either they don't contain any Pipfile to updated or they are supposed to be maintained by external contributors (Intel):

```
$ find . -type d -name "*3.11"
...
./cuda/c9s-python-3.11
./cuda/ubi9-python-3.11
...
./jupyter/intel/ml/ubi9-python-3.11
./jupyter/intel/pytorch/ubi9-python-3.11
./jupyter/intel/tensorflow/ubi9-python-3.11
...
./rstudio/c9s-python-3.11
...
./intel/base/gpu/ubi9-python-3.11
./intel/runtimes/ml/ubi9-python-3.11
./intel/runtimes/pytorch/ubi9-python-3.11
./intel/runtimes/tensorflow/ubi9-python-3.11
...
./rocm/ubi9-python-3.11
```

Also, we may consider to have this sorted alphabetically but rather in some future PR to keep this simple.

atheo89:
>Also, we may consider to have this sorted alphabetically but rather in some future PR to keep this simple.

Or we can use args for the positions of ubix and pyx.y, allowing them to be utilized dynamically by the GHA according to our on-demand needs.

harshad16:
>moreover could you inactivate/remove the scheduler from the piplock-renewal-YYYYx.yml [link](https://github.com/opendatahub-io/notebooks/blob/main/.github/workflows/piplock-renewal-2023a.yml) and keep only the dispatch since we are not going to that way anymore

@atheo89 , let do that in future, once we know that rhoai-2.8 is EOS. 
till then lets keep it, we need to discuss this more in depth.

harshad16:
Thanks for all the reviews.
If anything more is needed, let do that in a follow up PR.

/approve
/override ci/prow/images

PR #716, opened 2024-09-26
[RHOAIENG-12621] Update the updater github action to the latest branch

jstourac:
This updates the digest updater GHA scripts for both regular workbench images and also runtime images.

On top of that, the runtime images GHA script was a bit enhanced and cleaned of a couple of shellcheck warnings (I can put it as a separate PR if it will be preferred).

https://issues.redhat.com/browse/RHOAIENG-12621

## How Has This Been Tested?
Started relevant GHA via https://github.com/jstourac/notebooks/actions

Standard workbench images:
* https://github.com/jstourac/notebooks/actions/runs/11057100656
* Resulting PR - TBD

Runtime images:
* https://github.com/jstourac/notebooks/actions/runs/11057072140
* Resulting PR - TBD

---

The GHAs failed as we don't have any 2024b builds in https://quay.io/repository/opendatahub/workbench-images?tab=tags&tag=latest --- will re-run once we have some.

Note: one can see results of the openshift-ci jobs here https://prow.ci.openshift.org/?repo=opendatahub-io%2Fnotebooks.

Also, at the moment there is one extra commit in this PR that replaces names images for version N from python-3.9 to python-3.11 because otherwise it would still search for the python-3.9 based images! Version N-1 should be good without this hack. As a result, I plan to incorporate the update for the images as a part of this PR. This hack won't be necessary in the followup updates until we change the python version again.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #738, opened 2024-10-05
Remove unncessary one-time workaround from the digest updater GHA

jstourac:
This change was never intended to be merged as it was marked with the DONT COMMIT message in the draft PR. Since it got merged and the GHA has been executed updating the base images for the new N (2024b) version, we can safely remove this workaround now.

This is a followup of #716.

Note: I presume that 2024b branch manifests in this opendatahub-io org will be updated by merging changes from main to 2024b, correct? :thinking: @harshad16 If not, then this should stay open until this GHA is run over the 2024b branch also.

## Merge criteria:

### DIFF
--- a/.github/workflows/notebooks-digest-updater-upstream.yaml
+++ b/.github/workflows/notebooks-digest-updater-upstream.yaml
@@ -82,8 +82,7 @@ jobs:
             skopeo_metadata=$(skopeo inspect --retry-times 3 "docker://${img}")
 
             src_tag=$(echo "${skopeo_metadata}" | jq '.Env[] | select(startswith("OPENSHIFT_BUILD_NAME=")) | split("=")[1]' | tr -d '"' | sed 's/-amd64$//')
-            src_tag2=$(echo $src_tag | sed 's/python-3.9/python-3.11/')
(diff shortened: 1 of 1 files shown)


jstourac:
> Note: I presume that 2024b branch manifests in this opendatahub-io org will be updated by merging changes from main to 2024b, correct? :thinking: @harshad16 If not, then this should stay open until this GHA is run over the 2024b branch also.

Actually, it depends what branch we will use for the manifest reference... probably `main` is used for the ODH, in which case, we don't care about the manifests content in `2024b` branch?

jstourac:
> > Note: I presume that 2024b branch manifests in this opendatahub-io org will be updated by merging changes from main to 2024b, correct? :thinking: @harshad16 If not, then this should stay open until this GHA is run over the 2024b branch also.
> 
> Actually, it depends what branch we will use for the manifest reference... probably `main` is used for the ODH, in which case, we don't care about the manifests content in `2024b` branch?

This was confirmed on Monday meeting.

jstourac:
Since this is just gha change, I'll go ahead with just one lgtm.

/approve

PR #721, opened 2024-10-01
Updated the notebook manifests with 2024b image reference

dibryant:
fixes for https://issues.redhat.com/browse/RHOAIENG-12628
## Description

Updated the notebook manifests with 2024b and 2024a image reference
## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #742, opened 2024-10-11
[Fix] some pkg versions for the Tensorflow ImageStream manifest

jstourac:
this fixes versions of the CUDA bundle and Numpy packages

This is a followup for the #721. [This comment exactly](https://github.com/opendatahub-io/notebooks/pull/721#discussion_r1796788416).

## Merge criteria:


atheo89:
Good catch! 
/lgtm 
/approve

jstourac:
/override ci/prow/images

Thanks for your reivew, guys!

harshad16:
Thanks for the changes.
@jstourac , the numpy needs to be 1.26: https://github.com/opendatahub-io/notebooks/blob/7fb4961ba567fba9aedbb6c87270e8605af9ef3e/jupyter/tensorflow/ubi9-python-3.11/Pipfile.lock#L2266

jstourac:
Hmm, did I check the wrong image content then? I'll try to check on Monday again, cause its weird then.

jstourac:
> Thanks for the changes. @jstourac , the numpy needs to be 1.26:
> 
> https://github.com/opendatahub-io/notebooks/blob/7fb4961ba567fba9aedbb6c87270e8605af9ef3e/jupyter/tensorflow/ubi9-python-3.11/Pipfile.lock#L2266
> 
> 
> please correct me if i m looking at wrong place.

@harshad16 , I checked the [tensorflow image](https://quay.io/repository/opendatahub/workbench-images/manifest/sha256:2f4539e638188bb63f65823aa03a3e4614effc071041008be9cf7caa721cb401?tab=packages) referenced in our [params.env](https://github.com/opendatahub-io/notebooks/blob/7fb4961ba567fba9aedbb6c87270e8605af9ef3e/manifests/base/params.env#L9C81-L9C145) file currently. You can see that there are two Numpy versions shown by the quay.io actually:

```
numpy	1.26.4
numpy	2.1.1
```

Each was introduced in a different layer during the image build.
* The 2.1.1 version was introduced via [cuda-jupyter-datascience-ubi9-python-3.11-amd64](https://github.com/opendatahub-io/notebooks/blob/7fb4961ba567fba9aedbb6c87270e8605af9ef3e/Makefile#L214).
* The 1.26.4 was introduced via the following layer [cuda-jupyter-tensorflow-ubi9-python-3.11-amd64](https://github.com/opendatahub-io/notebooks/blob/7fb4961ba567fba9aedbb6c87270e8605af9ef3e/Makefile#L219).

In the result image, we can truly see the following:
```
(app-root) sh-5.1# pip list | grep numpy
numpy                        1.26.4

```

Truth is that I checked this with the downstream for 2.14RC1 build only via the test. Since the other images have the numpy updated to 2.1 already and the test showed this for this TensorFlow image also, I simply wen't on and updated this. But same data that I checked with the upstream image, can be seen also for the downstream 2.14 RC1 build TensorFlow image.

So, I'll raise a new PR to revert this version and will check what the heck was with the test to understand this a bit.

Thank you for noticing!

jstourac:
So, I understand why the test failed now. I misinterpreted the test failure - it actually failed on the reason that the Numpy package version doesn't match to what we have in the other packages. So I amended test so it doesn't expect Numpy (and couple more because of the recently added check for the Trusty AI image) to have the version aligned - red-hat-data-services/ods-ci#1923.

For the manifest partial revert of this PR, I raised #743 to make this consistent again.

@harshad16 thank you, again, for noticing this! :slightly_smiling_face:

PR #747, opened 2024-10-17
RHOAIENG-14520: fix(vscode): change supervisord files location

guimou:
The current supervisord configuration creates two file in /opt/app-root/src:
- supervisord.pid
- supervisord.log

This can be confusing for users, who don't know where they come from, what they do, if they can/should delete them... Especially as even if they are deleted they will keep coming back.

This PR modifies supervisord configuration to put those files under /tmp. That way they are hidden to the user, and reinitialized at each restart of the workbench (instead of having the log file growing).

### DIFF
--- a/codeserver/ubi9-python-3.11/supervisord/supervisord.conf
+++ b/codeserver/ubi9-python-3.11/supervisord/supervisord.conf
@@ -1,5 +1,7 @@
 [supervisord]
 nodaemon=true
+logfile=/tmp/supervisord.log
+pidfile=/tmp/supervisord.pid
 
 [program:fcgiwrap]
 command=/usr/sbin/fcgiwrap -s unix:/var/run/fcgiwrap.socket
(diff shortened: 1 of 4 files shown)


jiridanek:
/retitle RHOAIENG-14520: fix(vscode): change supervisord files location

jiridanek:
/lgtm

can you eventually rebase the PR on main, to get a recent-ish fix for gha?

atheo89:
Thank you @guimou for this fix! 
Tested locally
`podman run -it -p 8787:8787 quay.io/opendatahub/workbench-images:codeserver-ubi9-python-3.11-pr-747`

Looks much better now :slightly_smiling_face: 
image

/approve

jiridanek:
We're ready to have this for 2.16. The tests will fail anyways, so there's still lots of time to object if anyone objects ;P

/unhold
/retest-required

PR #754, opened 2024-10-25
RHOAIENG-14518: feat(codeserver): disable the VSCode workspace trust mechanism and the trust prompt

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-14518

## Description

Trust the authors popup appears the first time user opens code-server (VSCode) workbench. It is super scary and confuses users. What is this folder /opt/app-root/src? Is a security risk for my company if I authorize this?

We are taking the decision to spare our users such considerations by removing the VSCode workspace trust completely. Therefore the workspace will be automatically trusted and user will not be prompted about no trust whatsoever.

Alternatively, same thing could've been implemented by running `code-server --disable-workspace-trust` instead of modifying settings.

I looked into the Workspace Trust Editor, https://github.com/microsoft/vscode/blob/56b535f40900080fac8202c77914c5ce49fa4aae/src/vs/workbench/contrib/workspace/browser/workspaceTrustEditor.ts, trying to find where it stores trusted paths, since then maybe we could preconfigure /opt/app-root/src to be already trusted, but I did not find this.

## How Has This Been Tested?

* 3.9 image: quay.io/opendatahub/workbench-images:codeserver-ubi9-python-3.9-pr-754
* 3.11 image: quay.io/opendatahub/workbench-images:codeserver-ubi9-python-3.11-pr-754

I've loaded images into rhoai 2.13, checked that I don't see scary modal anymore at startup, and checked settings to see that they look like what I expected.

Screenshot 2024-10-25 at 2 27 21 PM

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #755, opened 2024-10-27
RHOAIENG-15152: feat(codeserver/e2e): add initial playwright e2e test project skeleton

jiridanek:
Made for https://issues.redhat.com/browse/RHOAIENG-14518 but it's essentially unrelated work.

* https://github.com/opendatahub-io/notebooks/pull/754

Filled new Jira just for the test https://issues.redhat.com/browse/RHOAIENG-15152

## Description

As you are well aware, we've been struggling to pick our e2e toolset and to write some left-shifted tests. Thankfully, we have a helpful Architect in Red Hat Middleware-Application Services (hello David), who recommended Playwright in TypeScript as the only sensible e2e option for web stuff that even developers are willing to touch.

I did consider Cypress and various Selenium bindings; I would not mind switching what I have here into that if it's more liked in the team. Just I think we should polish and merge this Playwright first, even if we're going to abandon it, so that we at least have something.

The `await` porn is even more stupid than I expected that to be. Mandatory reading https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/ And there's some theory about effect systems if you want to discuss this with academically minded people!?!

I'm using the Playwright test runner. That's by no means required; it's possible to use just the automation library part of Playwright and structure the tests differently. I guess I decided that since I don't have any test runner I like in the JavaScript world, I may just as well pick the Playwright one. It has fixtures like Pytest, so there may be some familiarity there.

Besides that, we have here testcontainers, because they are also the most popular and therefore should be relatively unobjectionable choice for a library to start a container in the background.

### What does it look like

Playwright has a fancy html report. I do pretty much agree with @zhimin's prolific blogs that are saying that all that is not actually well suited for a traditional QE team (edit: that's working in true agile way), but we aren't that, and also the number of e2e tests we're going to have here to run on every PR is going to be small. We're still free to test differently in QE (or DevTestOps as we now say).

Screenshot 2024-10-28 at 12 33 16 PM

### What's it like to work with

The tests can be set to run against fresh browser as well as a browser you provide (the `--remote-debugging-port` is what does the trick). Check out `playwright.config.ts` where you can switch between the two modes.

```
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --remote-debugging-port=9222 --user-data-dir=./data23 --enable-automation --no-first-run
```

In the same vein, the tests can run against already started codeserver image, or they can start its own container, fresh one for every test. (I'm not sure if we should try to reuse codeserver instance or start fresh every time. Fresh should make writing tests easier as cleanup becomes unnecessary and the cost in increased test runtime is I think worth that. If we add more suites besides this "basic functionality" one, then maybe we can reassess that there. Good test hygiene is hard, especially keeping it up for a longer time.)

```
podman run -p 8787:8787 --rm -it quay.io/opendatahub/workbench-images:codeserver-ubi9-python-3.11-2024b_20241026
```

### Debugging tests

Running against provided browser and provided codeserver, possibly commenting out few unnecessary steps before the interesting part, it is usually possible to iterate on the thing quickly and get it working. Then check with fresh browser and fresh image.

Playwright has the `codegen` and `test --ui` tools. The codegen is ok to give you decent locators to use, and the ui is not bad either; video recording on steroids.

The `test --ui` thing is pretty much what Cypress has also; the thing we were so impressed with in Savitha's demo earlier this year.

For some reason, default Playwright settings will popup a _new browser tab_ with report every time a test run fails. I've disabled that popup thing and instead enabled the `line` reporter that gives you some stacktrace; thats way more reasonable.

I do have two fancy test helpers in some utils file. They are sufficiently complex that there may need to be special tests to test these tests (helpers) ;P

### The tests themselves

I essentially copied that from Coder Inc. I did preserve copyright header and put MIT licence, same as they have, so we should be good. The difference from them is that they control `code-server` invocation, whereas for us the parameters given to it are fixed. So my tests can be simpler.

## How Has This Been Tested?

GHA and played with it on my machine.

image

## Merge criteria:

I'll cleanup the git history later. Please have a look now nevertheless.


jiridanek:
The CI fails are missing `---` in the pnpm lock file yaml, and unreliable trivy scanner. For `---` I think we should put that file into ignore, and for trivy not sure what to do. Ganesh (former colleague of sorts) just disabled it in their

* https://github.com/skupperproject/skupper-router/pull/1643

jiridanek:
One funny realization; depending on the platform where you run codeserver (or most likely where and what browser you're opening it with), the buttons for trusting and not trusting workspace are switched. Either the ack button is on the left, or on the right. Never realized that it actually changed order for me after I switched from KDE Plasma to MacOS!

codeserver/ubi9-python-3.11/README.md
```diff
@@ -1,3 +1,4 @@
+
```
jiridanek, line 1: don't forget to undo this

codeserver/ubi9-python-3.9/README.md
```diff
@@ -1,3 +1,4 @@
+
```
jiridanek, line 1: don't forget to undo this

.github/workflows/build-notebooks-TEMPLATE.yaml
```diff
@@ -134,7 +134,7 @@ jobs:
         if: "${{ fromJson(inputs.github).event_name == 'pull_request' }}"
         env:
           IMAGE_TAG: "${{ github.sha }}"
-          IMAGE_REGISTRY: "localhost:5000/workbench-images"
+          IMAGE_REGISTRY: "ghcr.io/${{ github.repository }}/workbench-images"
```
jiridanek, line 137: We aren't pushing this, so we can use any registry we like; so let's keep it consistent.

jstourac:
I didn't actually run this on my machine. LGTM in general. Do we want to create this stub only in the codeserver indeed? If we're introducing this, don't we want to create it somehow generic, so we can reuse at least portions of code/tests between the diferent images?

Also, looks like the Notebooks 2.0 will use Cypress in the end: kubeflow/notebooks#75. Should we be concerned? :thinking:

jiridanek:
Here's nothing to run, actually. It's not a walking skeleton, just a skeleton.

I originally intended to keep this in code server for the time being, but you're right that maybe starting in tests/e2e or some directory like that makes more sense. I'll do that change tomorrow.

Regarding cypress, I am actually not concerned. Playwright is a better design (real browser automation library) and if we ever wanted to copy code from somebody else's cypress tests, we can do that with not too much effort.

jiridanek:
/override ci/prow/images
not touching no images

jstourac (commented):
/lgtm

PR #772, opened 2024-11-13
RHOAIENG-11156: Add a loading spinner to JupyterLab index page

caponetto:
https://issues.redhat.com/browse/RHOAIENG-11156

## Description

This PR includes a script to inject a simple loading spinner into JupyterLab's `index.html` page. The script runs as part of the build image process. The set of necessary files had to be duplicated to be present on all JupyterLab-based images. The duplication can be removed as part of [this task](https://issues.redhat.com/browse/RHOAIENG-11274). 

This is a workaround; ideally, it should be handled at https://github.com/jupyterlab/jupyterlab/issues/5463.

Demo (cache disabled + Fast 4G):

https://github.com/user-attachments/assets/4d2ffc46-33c8-4f64-a34c-1982f1598384

## How Has This Been Tested?

Run a JupyterLab-based image from this PR, for instance:
```bash
podman|docker run -p 8888:8888 ghcr.io/caponetto/opendatahub-io-notebooks/workbench-images:jupyter-intel-ml-ubi9-python-3.9-RHOAIENG-11156_80df8ffc8ea12f30a5727d91a6befe0fa8a3bfb3

podman|docker run -p 8888:8888 ghcr.io/caponetto/opendatahub-io-notebooks/workbench-images:jupyter-trustyai-ubi9-python-3.11-RHOAIENG-11156_80df8ffc8ea12f30a5727d91a6befe0fa8a3bfb3 
```
Wait for the image to start and open JupyterLab URL (the URL with token appears in the terminal):
```
http://127.0.0.1:8888/lab?token=XXXXXX
```
Once opened, you should see a loading spinner instead of a blank screen. If you're using Chrome, you can open DevTools and go to the Network tab to disable cache and select Fast/Slow 4G, which slows down everything.

## Merge criteria:

### DIFF
--- a/jupyter/intel/ml/ubi9-python-3.11/utils/addons/apply.sh
+++ b/jupyter/intel/ml/ubi9-python-3.11/utils/addons/apply.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+# See https://github.com/jupyterlab/jupyterlab/issues/5463
+# This is a hack to apply partial HTML code to JupyterLab's `index.html` file
+# Look for the other duplicates in case a change is needed to this file
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+pf_url="https://unpkg.com/@patternfly/patternfly@6.0.0/patternfly.min.css"
+
+static_dir="/opt/app-root/share/jupyter/lab/static"
+index_file="$static_dir/index.html"
+
+head_file="$script_dir/partial-head.html"
+body_file="$script_dir/partial-body.html"
+
+if [ ! -f "$index_file" ]; then
+  echo "File '$index_file' not found"
+  exit 1
+fi
+
+if [ ! -f "$head_file" ]; then
+  echo "File '$head_file' not found"
+  exit 1
+fi
+
+if [ ! -f "$body_file" ]; then
+  echo "File '$body_file' not found"
+  exit 1
+fi
+
+curl -o "$static_dir/pf.css" "$pf_url"
+
+head_content=$(tr -d '\n' <"$head_file" | sed 's/@/\\@/g')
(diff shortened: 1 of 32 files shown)


jiridanek:
What does this look with a dark theme?

caponetto:
> What does this look like with a dark theme?

Same

jiridanek:
I don't like how this is implemented, but I guess it just means I should get a move on the deduplication Jira.

atheo89:
Cool, I like that! 

/lgtm

Seems that the background music effects will soon be available while training a model on the JL :rofl:

jupyter/intel/ml/ubi9-python-3.11/utils/addons/apply.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+# See https://github.com/jupyterlab/jupyterlab/issues/5463
+# This is a hack to apply partial HTML code to JupyterLab's `index.html` file
+# Look for the other duplicates in case a change is needed to this file
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+pf_url="https://unpkg.com/@patternfly/patternfly@6.0.0/patternfly.min.css"
```
jiridanek, line 9: @jstourac was scared for a moment what would happen on disconnected cluster here! but the good news is that this stylesheet is fetched on users browser, and that does have general network access, or does it, actually?

it would work in the DevTestOps disconnected setup, but I imagine that companies that use disconnected setups for their clusters may also restrict the user's computers net access?

caponetto: This URL is downloaded as part of the script that runs during image build time. The CSS file becomes available locally for the page to use at runtime.

PR #775, opened 2024-11-22
 RHOAIENG-16076: tests(gha): run Makefile tests on opendatahub-io/notebooks Github Actions

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-16076

This is building up on the preparatory work merged in
* https://github.com/opendatahub-io/notebooks/pull/780
* https://github.com/opendatahub-io/notebooks/pull/777
* https://github.com/opendatahub-io/notebooks/pull/782
* https://github.com/opendatahub-io/notebooks/pull/783

## Description

Deploys cri-o backed kubernetes with kubeadm to run our Makefile tests.

I considered using KinD, or Microshift, or Minikube, or OpenShift Local, or some other Kubernetes, but decided not to, because

* Microshift still requires Red Hat pull secret, so it's hard to run this for PRs
  * There is something new there, though https://github.com/openshift/microshift/tree/main/okd/src
* OpenShift Local takes long time to start, the OpenShift version needs pull-secret, and the OKD version is not getting much love
* KinD, etc. with CRI-O requires custom node image https://github.com/cri-o/cri-o/blob/release-1.31/tutorials/crio-in-kind.md#cri-o-in-kind

The reason why I need cri-o is that I want to share the podman-built images into the cluster without having to copy them. Copying images is difficult because the images can be huge.

* seems that k3d would've been also possible, https://devopstales.github.io/kubernetes/k3s-crio/

### Known issues that I do intend to fix

```
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... failed: Name or service not known.
wget: unable to resolve host address ‘raw.githubusercontent.com’
```

* [x] This needs a retry, because it happens from time to time. Also, we should not fetch test data from the main branch of the repo, we should use what's in the current checkout.

This went away after I redid my Kubernetes network. I had this problems with flannel, and since then I switched to a plain bridge with firewall masquerade. I think that (when I noticed this problem the first time, or maybe at some later point) my pods were completely unable to connect to the outside (also unable to connect to other pods, like the coredns pod, but I was dealing with other issues so I mistakenly thought the networking issues are a flakiness.

---

```
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7f95ec277fd0>: Failed to establish a new connection: [Errno -2] Name or service not known')': /simple/papermill/
[...]

PR #782, opened 2024-11-24
RHOAIENG-16076: tests(gha): change to using rootful podman, so that we can share `containers/storage` with cri-o later

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-16076

Another chunk of the

* https://github.com/opendatahub-io/notebooks/pull/775

changes to be reviewed and merged.

## Description

I need to save the disk space, so I can't just copy the built images into KinD or some such local instance of kubernetes where I can run the Makefile tests. Therefore, I need to build the images in a rootful Podman, and then (subsequent PR) I'll be able to reuse it's containers/storage directory in cri-o backed kubernetes.

https://podman-desktop.io/blog/sharing-podman-images-with-kubernetes-cluster#introduction

## How Has This Been Tested?

GHA

* https://github.com/jiridanek/notebooks/actions/runs/12025659264

## Merge criteria:


ci/cached-builds/podman.socket
```diff
@@ -0,0 +1,12 @@
+# cat /usr/lib/systemd/system/podman.socket
+
+[Unit]
+Description=Podman API Socket
+Documentation=man:podman-system-service(1)
+
+[Socket]
+ListenStream=%t/podman/podman.sock
+SocketMode=0666
```
jiridanek, line 9: https://en.wikipedia.org/wiki/Number_of_the_beast

jstourac:
I have to admit I don't fully follow all of the configuration changes here, but LGTM in general overview and if it works and helps to reduce the disk-space utilized, I'm all for this.

Thank you, Jiri!

/lgtm

.github/workflows/build-notebooks-TEMPLATE.yaml
```diff
@@ -20,6 +20,12 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     env:
+      # Some pieces of code (image pulls for example) in podman consult TMPDIR or default to /var/tmp
+      TMPDIR: /home/runner/.local/share/containers/tmpdir
+      # Use the rootful instance of podman for sharing images with cri-o
+      # https://podman-desktop.io/blog/sharing-podman-images-with-kubernetes-cluster#introduction
+      # https://access.redhat.com/solutions/6986565
+      CONTAINER_HOST: unix:///var/run/podman/podman.sock
```
jiridanek, line 28: when podman cli is started, it consults this env variable; when it is present, then it connects (as a client) to this socket, so things work pretty much as they would in the docker world

the server can be a systemd service running as root, and if current user has permissions on the socket (it has, because of `0666`), then we run rootful podman

.github/workflows/build-notebooks-TEMPLATE.yaml
```diff
@@ -101,22 +107,45 @@ jobs:
...
+          sudo mount --bind -o rw,noexec,nosuid,nodev,bind $TMPDIR /var/tmp
+
+          # podman from brew has its own /etc (was giving me Failed to obtain podman configuration: runroot must be set)
+          # the (default) config location is also where cri-o gets its storage defaults (that can be overriden in crio.conf)
+          sudo cp ci/cached-builds/containers.conf /etc/containers.conf
+          sudo cp ci/cached-builds/containers.conf /home/linuxbrew/.linuxbrew/opt/podman/etc/containers.conf
+          sudo cp ci/cached-builds/storage.conf /etc/containers/storage.conf
+          sudo cp ci/cached-builds/storage.conf /home/linuxbrew/.linuxbrew/opt/podman/etc/containers/storage.conf
+          sudo cp ci/cached-builds/registries.conf /etc/containers/registries.conf
+          sudo cp ci/cached-builds/registries.conf /home/linuxbrew/.linuxbrew/opt/podman/etc/containers/registries.conf
```
jiridanek, line 123: these are just the default config files, but I need to have some nondefault values in them, mostly the path to containers/storage; it's pretty much the same as it was before, but i'm putting them into the systemwide location, not just into user directory, because now we run rootful podman

also, I am putting it into the systemwide locations for crio as well as podman, and because podman is installed from homebrew, it lives under /home/linuxbrew/.linuxbrew/opt/podman/etc; weird but so is life

ci/cached-builds/registries.conf
```diff
@@ -0,0 +1,6 @@
+# https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md
+
+# prevent this kyverno error
+#  Failed to pull image "bitnami/kubectl:1.26.4": reading manifest 1.26.4 in quay.io/bitnami/kubectl: unauthorized: access to the requested resource is not authorized
+unqualified-search-registries = [ "docker.io" ]
+short-name-mode = "enforcing"
```
jiridanek, line 6: i only saw this when I was doing something with kyverno (I don't intend to push that), but it looks like a good default; kyverno only has images in docker.io, but with the previous defaults somehow an attempt to pull from quay.io was made; let's stop that

jiridanek:
thanks for the review; I added some comments, and I'd be happy to discuss this more; it is possible I am doing something overly complicated here; it's always best to have simple infra as possible

/approve

PR #803, opened 2024-12-11
RHOAIENG-16517: chore(tests): add sandboxing so that Dockerfile builds can only access files we know they access

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-16517

## Description

This(will be) important for OCP-CI and for Konflux, because there it is not possible to run the changed files detection up front, but instead the rules must be written into a .tekton directory (konflux) or in openshift/release (ocp-ci).

We must keep track of what files go into each image build, otherwise we cannot determine affected subset of images when testing a PR and would have to build everything.

## How Has This Been Tested?

* trial build at https://github.com/opendatahub-io/notebooks/actions/runs/12275919093

## Merge criteria:


jiridanek:
/approve
this also increases disk space on /, so we want this sooner rather than later

jiridanek:
@jstourac I like the override idea

PR #805, opened 2024-12-11
Update Piplock Renewal GitHub Action

atheo89:
Related to: https://issues.redhat.com/browse/RHOAIENG-15393

## Description

This PR consolidates multiple changes related to Pipfile.lock files generation. The main goal is to streamline the process of updating Pipfile.lock files across 

The following updates have been made:
- Refined the `piplock-renewal.yaml` GitHub Action to facilitate automated updates and automerging of Pipfile.lock files, with `user-configurable` Python versions, branch and optional directory updates.
- Modified the refresh-pipfilelock-files makefile to support updates for optional directories.
- Included the codeserver notebook in the refresh recipe.
- Improved compatibility for the Intel flavor of the makefile as they are not official supported
- Removed `.github/workflows/piplock-renewal-YYYYx.yml` as we will not use them anymore
~- Add `.gitignore` files to block user push pipfile.lock, as from now on will be populated only by ` piplock-renewal.yaml` action.~
- Update the `Readme.md` and `CONTRIBUTING.md` files  

## How Has This Been Tested?

The functionality of these updates tested locally:
`DIRS` for testing purposes https://github.com/atheo89/notebooks/blob/main/Makefile#L643 

Having optional dirs = false
https://github.com/atheo89/notebooks/actions/runs/12274941817/job/34248935523

Having optional dirs = true
https://github.com/atheo89/notebooks/actions/runs/12275077390/job/34249373408

 `workflow_dispatch` inputs from the user, none of them is required:
 image
 
## Merge criteria:


Makefile
```diff
@@ -636,41 +636,63 @@ validate-rstudio-image: bin/kubectl
...
+        base/ubi9-python-$(PYTHON_VERSION) \
+        jupyter/minimal/ubi9-python-$(PYTHON_VERSION) \
+        jupyter/datascience/ubi9-python-$(PYTHON_VERSION) \
+        jupyter/pytorch/ubi9-python-$(PYTHON_VERSION) \
+        jupyter/tensorflow/ubi9-python-$(PYTHON_VERSION) \
+        jupyter/trustyai/ubi9-python-$(PYTHON_VERSION) \
+        jupyter/rocm/tensorflow/ubi9-python-$(PYTHON_VERSION) \
+        jupyter/rocm/pytorch/ubi9-python-$(PYTHON_VERSION) \
+		codeserver/ubi9-python-$(PYTHON_VERSION) \
+		runtimes/minimal/ubi9-python-$(PYTHON_VERSION) \
```
jstourac, line 653: indentation intentional?

Makefile
```diff
@@ -636,41 +636,63 @@ validate-rstudio-image: bin/kubectl
...
+
+# Default value is false, can be overiden
+# The below directories are not supported on tier-1
+INCLUDE_OPT_DIRS ?=false
+OPT-DIRS := jupyter/intel/ml/ubi9-python-$(PYTHON_VERSION) \
+			jupyter/intel/pytorch/ubi9-python-$(PYTHON_VERSION) \
+			jupyter/intel/tensorflow/ubi9-python-$(PYTHON_VERSION) \
+			intel/runtimes/ml/ubi9-python-$(PYTHON_VERSION) \
+			intel/runtimes/pytorch/ubi9-python-$(PYTHON_VERSION) \
+			intel/runtimes/tensorflow/ubi9-python-$(PYTHON_VERSION)
```
jstourac, line 668: indentation intentional?

.github/workflows/piplock-renewal.yaml
```diff
@@ -0,0 +1,65 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      # Checkout the specified branch from the specified organization
+      - name: Checkout code from the specified branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch }}
+          token: ${{ secrets.GH_ACCESS_TOKEN }}
```
jstourac, line 41: why we have the naming different to what we use in [kubeflow](https://github.com/opendatahub-io/kubeflow/pull/489/files#diff-9c97925bf1481f29538c73d5dd78b0639e2629a148e381aec814d25669f0f3b5R87)? Is there some reason for this? Note that I like this naming more than the one in the kubeflow TBH.

.github/workflows/piplock-renewal.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+      # Install pipenv
+      - name: Install pipenv
+        run: pip install pipenv
+
+      # Run makefile recipe to refresh Pipfile.lock and push changes back to the branch
+      - name: Run make refresh-pipfilelock-files and push the changes back to the branch
+        run: |
+          make refresh-pipfilelock-files PYTHON_VERSION=${{ github.event.inputs.python_version }} INCLUDE_OPT_DIRS=${{ github.event.inputs.update_optional_dirs }}
+          git add -f .
```
jstourac, line 63: hmm, just a guess but - will the Pipfile.lock files be actually added to the commit since they are in the .gitignore file?

base/c9s-python-3.11/.gitignore
```diff
@@ -0,0 +1,6 @@
+# pipenv
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
+#   However, in case of collaboration, if having platform-specific dependencies or dependencies
+#   having no cross-platform support, pipenv may install dependencies that don't work, or not
+#   install all needed dependencies.
+Pipfile.lock
```
jstourac, line 6: Is there a reason we don't use one global .gitignore file and having this in multiple directories?

Also, I'm not sure I truly like it - it's in the repo but we put it into ignore... maybe I would rather have something like pre-commit or GHA that checks that the Pipfile.lock isn't changed by user? :thinking: but maybe even that no... we should simply tell that the PR is bad if it is updating the Pipfile.lock instead of using this new GHA, no? In other words - why having this file in gitignore? :thinking: :shrug:

Makefile
atheo89: Oh.. I don't know why it is shown like this on the Git preview on vscode looks aligned... :thinking: 
 
 
image

.github/workflows/piplock-renewal.yaml
atheo89: This was the old original name that this GHA used. Thus, I keep the same name, but now I issued the token from the `ide-developer` account. The old one I have a feeling that issued by personal account.

base/c9s-python-3.11/.gitignore
atheo89: >Is there a reason we don't use one global .gitignore file and having this in multiple directories?

Yes, this way, if we deprecate a notebook flavor will deprecate also this restriction, the same if we add a new notebook.  On the other side, on the centralized .gitignore file we need to keep maintain the paths of the notebooks  that is applies. 

>why having this file in gitignore?
To avoid different variations of Pipfile.lock that generated the users local envs while they push to the repo and cause conflicts. 
Added a section on the readme file and on the contributing.md: 

**Pipfile.lock Generation**
To maintain **consistency across different local environments** and prevent platform-specific issues, users should not push their Pipfile.lock files after local execution. Collaborators may encounter discrepancies due to varying local setups, leading to potential conflicts during the lock file generation. Instead, users can update these files through the [piplock-renewal.yaml](https://github.com/opendatahub-io/notebooks/blob/main/.github/workflows/piplock-renewal.yaml) GitHub Action. 

This action allows users to specify a branch for updating and automerging the Pipfile.lock files, select the desired Python version, and choose whether to include optional directories in the update.
*NOTE:* To ensure the GitHub Action runs successfully, users must create a `GH_ACCESS_TOKEN` secret in their fork.

Makefile
jstourac: tab vs space probably?

Makefile
jstourac: hm, okay, I don't follow why, though :thinking: you just feel this better readable or am I missing some technical aspect also? :slightly_smiling_face:

base/c9s-python-3.11/.gitignore
jstourac: > Yes, this way, if we deprecate a notebook flavor will deprecate also this restriction, the same if we add a new notebook. 

I don't follow. What does it mean when we deprecate? When we don't want this restriction to be in place anymore?

> To avoid different variations of Pipfile.lock that generated the users local envs while they push to the repo and cause conflicts.

But people don't push to the repo usually. We use PRs for this. And we can tell user to revert such changes in their PR. Also, I don't know in which situation user has different content of the Pipfile.lock files than what is in the repo. Usually only in case of development, no? TBH, I don't feel comfortable to seeing different results when I build image locally compared to when it is done for us via openshift-ci/GHA and pushed to quay... having these files in gitignore will mask me the information that I have some local changes on the repository that will result in different image to what is on the upstream repo with the very same commit e.g.

Makefile
atheo89: :P It is because of readability as you mentioned the same as the `BASE_DIRS`
image

.github/workflows/piplock-renewal.yaml
jstourac: Okay, but maybe it would make sense to unify this among the repositories in the future to avoid confusion as I guess these are both same things with just different naming, right? :slightly_smiling_face:

Makefile
atheo89: huhhh why the VScode shows them well, but here are so bad?

jstourac: it's a tab vs spaces issue as above - at least that is what I see in my vscode when I fetch your changes locally

Makefile
atheo89: This fixed now, but the other `OPT_DIR` still looks shity

base/c9s-python-3.11/.gitignore
jiridanek: >> Is there a reason we don't use one global .gitignore file and having this in multiple directories?

>Yes, this way, if we deprecate a notebook flavor will deprecate also this restriction, the same if we add a new notebook. On the other side, on the centralized .gitignore file we need to keep maintain the paths of the notebooks that is applies.

add a top level .gitignore and put there

```
**/Pipfile.lock
```

problem solved

> Two consecutive asterisks ("**") in patterns matched against full pathname may have special meaning:
>
> A leading "**" followed by a slash means match in all directories.
>
> https://git-scm.com/docs/gitignore#_pattern_format

atheo89: When we deprecate a feature, it means we mark it as outdated or no longer recommended for use, with the intention of removing it in the future. In this specific case, if we deprecate a notebook flavor, it means we no longer actively support it. Consequently, any restrictions associated with it—such as this one—would also be deprecated. Similarly, if we add a new notebook flavor, relevant restrictions would naturally apply to that new flavor.


And about your concerns about masking differences and ensuring consistency in generated images are noted, but let me address the key issues:

We should rely on a single, consistent source for the Pipfile.lock generation to avoid discrepancies. The images generated by GHA (GitHub Actions) or OpenShift CI (ocp-ci) serve this purpose because these are the images ultimately delivered to customers. This way, we can ensure that all builds, whether local or in CI, align with the same source of truth.

For users, the process is straightforward:

- Trigger the GHA to generate the Pipfile.lock files for their branch and Python version.
- Fetch the generated files back with a simple git pull.

This eliminates any masking or differing results between local and upstream, providing complete transparency and consistency.

The main objective of this PR is to address conflicts arising between upstream and downstream branches, as detailed in the Jira issue. These conflicts are caused by OS-specific differences (e.g., markers in Pipfile.lock) that occur when Pipfile.lock is generated across various local environments. Such discrepancies lead to unnecessary diffs and merge conflicts, as observed in PRs like [#423](https://github.com/red-hat-data-services/notebooks/pull/423) and [#424](https://github.com/red-hat-data-services/notebooks/pull/424#issuecomment-2453039347). Centralizing this process resolves these issues by standardizing the generated files.

If there is a more effective way to eliminate these conflicts while maintaining consistency across environments and CI, we are open to hearing it. Until then, this approach ensures a clean, reliable, and unified process.

jiridanek: > having these files in gitignore will mask me the information that I have some local changes on the repository that will result in different image to what is on the upstream repo with the very same commit e.g.

isn't that the same problem as if you forget commit some new files and then you wonder why locally you get different results than what ci gives you?

jiridanek (commented):
The tab formatting is difficult. I remember we were struggling with it every time Makefile had to be updated. We should seriously consider

* https://github.com/opendatahub-io/notebooks/pull/806

base/c9s-python-3.11/.gitignore
jstourac: > add a top level .gitignore and put there

yeah, exactly what I meant - If we really need it then have it globally on one place and not on multiple places... then I didn't understand the note about the deprecation, so I'm missing something here.

> Consequently, any restrictions associated with it—such as this one—would also be deprecated. Similarly, if we add a new notebook flavor, relevant restrictions would naturally apply to that new flavor.

Right, but I still don't understand how is this related to the .gitignore and Pipfile.lock files.

> We should rely on a single, consistent source for the Pipfile.lock generation to avoid discrepancies. The images generated by GHA (GitHub Actions) or OpenShift CI (ocp-ci) serve this purpose because these are the images ultimately delivered to customers. This way, we can ensure that all builds, whether local or in CI, align with the same source of truth.

Definitely agree, that is what I wanted in the past already - having a GHA to update the Pipfile.lock files for us with a unified way. So this part is good from my side and I never objected it.

> These conflicts are caused by OS-specific differences (e.g., markers in Pipfile.lock) that occur when Pipfile.lock is generated across various local environments. Such discrepancies lead to unnecessary diffs and merge conflicts, as observed in PRs like https://github.com/red-hat-data-services/notebooks/pull/423 and https://github.com/red-hat-data-services/notebooks/pull/424#issuecomment-2453039347. Centralizing this process resolves these issues by standardizing the generated files.

Right, but still - do we really want to have this in .gitignore since it's part of the repository anyway? As I wrote above - people make changes to the repository via PRs. We control what goes in via PR. We can say that the change in pipfile.lock need to be removed from the PR...

> isn't that the same problem as if you forget commit some new files and then you wonder why locally you get different results than what ci gives you?

Okay, so I tried locally now. I thought that the change in the file won't be shown in `git diff` command. But it isn't true - it actually is. Also, looks like if the Pipfile.lock is already part of the repository, then the .gitignore doesn't work for it for me? I can add the file even without the `-f` option and commit it... what am I doing wrong? If I understand correctly - us adding these into .gitignore and still having them in the repo actually doesn't change anything from the user POV? :thinking:

jstourac: > Okay, so I tried locally now. I thought that the change in the file won't be shown in git diff command. But it isn't true - it actually is. Also, looks like if the Pipfile.lock is already part of the repository, then the .gitignore doesn't work for it for me? I can add the file even without the -f option and commit it... what am I doing wrong? If I understand correctly - us adding these into .gitignore and still having them in the repo actually doesn't change anything from the user POV? 🤔

Okay, so to stop tracking - I should probably do this: https://stackoverflow.com/a/18277622 (haven't tried that yet). But then, my original concern is back - I won't see my local changes of the file against to the content in the repo so I will be confused why my local image build is different to what is in quay.io....

jstourac: > If there is a more effective way to eliminate these conflicts while maintaining consistency across environments and CI, we are open to hearing it.

I think I tried to propose someting already :slightly_smiling_face: : 

> Also, I'm not sure I truly like it - it's in the repo but we put it into ignore... maybe I would rather have something like pre-commit or GHA that checks that the Pipfile.lock isn't changed by user?

atheo89:
I removed the `.gitignore` from the subfolders and i added on top level as Jiri proposed. 

>The tab formatting is difficult. I remember we were struggling with it every time Makefile had to be updated. We should seriously consider
NO-JIRA: chore(Makefile): replace ^\t (that's the default .RECIPEPREFIX) with >  #806

That's cool, I will rebase and apply the same, as soon your PR get merged

base/c9s-python-3.11/.gitignore
jiridanek: > If there is a more effective way to eliminate these conflicts while maintaining consistency across environments and CI, we are open to hearing it.

maybe just run pip from the makefile, in podman container, in the right python version, and with the various switches for platform, etc.

```
python3.11 -m pipenv lock --extra-pip-args "--only-binary=:all: --platform=manylinux_2_28_x86_64 --python-version=3.11 --implementation=cp"
``` 

stuff like that

jstourac: Yep, basically what the GHA would do, just on your local machine. Still, we wouldn't be sure that user generated these this way maybe... so some GHA gating this on PR would be needed anyway.

jiridanek: > so some GHA gating this on PR would be needed anyway

we have already a Jira for this from @paulovmr, needs to be pre-refined, refined, planed, .... let's not get ahead of ourselves https://issues.redhat.com/browse/RHOAIENG-8778

atheo89: Conclusion:
Removed the .gitignore from this implementation as we discussed on the release improvements meeting. 
We will do a better investigation in the future about a better tool to manage the python packages, to avoid several issues that the pipenv cause.

jstourac:
Thank you for amending this Adriana! Let's get these changes in.

/lgtm

atheo89:
Thank you for your review guys. 
/approve

PR #806, opened 2024-12-12
NO-JIRA: chore(Makefile): replace `^\t` (that's the default `.RECIPEPREFIX`) with `> `

jiridanek:
## Description

Turns out that progress is not as fast as it could be because developers have difficulties using the `\t` character in Makefile correctly. In order to accelerate the development, we should stop using the difficult old-school significant-whitespace syntax and instead we should switch to use the `>` character, that is fit for the sensibilities of the 21th century.

```
future: past
> ONWARDS!
```

## How Has This Been Tested?

* https://github.com/jiridanek/notebooks/actions/runs/12294951030
* after rebase when https://github.com/opendatahub-io/notebooks/pull/805 merged https://github.com/jiridanek/notebooks/actions/runs/12315590805

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,14 +11,13 @@ SHELL := bash
 MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 
-# todo: leave the default recipe prefix for now
 ifeq ($(origin .RECIPEPREFIX), undefined)
   $(error This Make does not support .RECIPEPREFIX. Please use GNU Make 4.0 or later)
 endif
-.RECIPEPREFIX =
+.RECIPEPREFIX = >
 
 IMAGE_REGISTRY   ?= quay.io/opendatahub/workbench-images
-RELEASE	 		 ?= 2024b
+RELEASE          ?= 2024b
 # additional user-specified caching parameters for $(CONTAINER_ENGINE) build
 CONTAINER_BUILD_CACHE_ARGS ?= --no-cache
 # whether to build all dependent images or just the one specified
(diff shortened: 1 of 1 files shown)


jiridanek:
what's @harshad16 gonna say when all tabs are gone after he comes from pto?

andyatmiami:
I am with @jstourac in that I feel the `>` is a bit of an odd choice.... but, while I have no personally yet experienced the pain, I can appreciate the motivate outlined in https://tech.davis-hansson.com/p/make/.

I would be just as happy to leave it with `^\t` - but also have no problem adopting `>` if that is the direction the team wants to go.

However, this does potentially lead to some pain/vigilance when we sync from upstream, right?  any changes in the `Makefile` made upstream will require then manual intervention from us to add the prefix character?

jiridanek:
> I am with @jstourac in that I feel the > is a bit of an odd choice

no disagreement there, but almost anything is better than the default significant-whitespace syntax, imo

c.f. https://en.wikipedia.org/wiki/Whitespace_(programming_language) for an extreme case of significant-whitespace in action

caponetto:
> I'm not sure I like the `>` style... I would have to get used to it. But if that means we can use just spaces and not being confused around, I think I can live with that. I'll give lgtm, but this should definitely be agreed by wider team members first.
> 
> /lgtm

+1 for Jan's comment but I'm not against it, so
/lgtm

It'd be nice if we could have a format checker on every commit/PR to guarantee the rules we agree on.

jiridanek:
Discussed in office-hours. Let's not do this, because

1. we want to remove the long shell scripts from Makefile anyways, and when the Makefile is simple this is not nearly as beneficial
2. other Makefiles in other repos are still using tabs, and we can't change those, because they come from kubeflow/kubeflow, so doing this would lead to confusion
3. intelliJ does not support changing the recipeprefix

PR #809, opened 2024-12-13
NO-JIRA: chore(.editorconfig): introduce `.editorconfig` file for the project

jiridanek:
https://editorconfig.org/

## Description

image

## How Has This Been Tested?

Works on my machine (in IntelliJ, but `.editorconfig` is supported in all major editors).

## Merge criteria:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# https://editorconfig.org/
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[Makefile]
+indent_style = tab


jstourac:
/lgtm

If I understand correctly - then e.g. for VS code, I need to install [this plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig), right?

.editorconfig
jiridanek, line 1: > If I understand correctly - then e.g. for VS code, I need to install [this plugin] (https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig), right?
>
> Update: yeah, based on this, looks like that https://editorconfig.org/.

yes, that one

jiridanek, line 1: > Update2: should we put some note about this somewhere in our readme (not only for a newcomers) to be aware of this?

Projects mostly don't mention this, for example vscode ohly has this comment on top of .editorconfig itself and that's that

```editorconfig
# EditorConfig is awesome: https://EditorConfig.org
```

https://github.com/microsoft/vscode/blob/31092ac1a0f9e68026cd0796f184fd3dfe978069/.editorconfig#L1

jiridanek: I would not mention it either, if we mentioned everything that could be mention, nobody would ever find strength of will to read our bloated readme.

jstourac: okay, I don't have strong opinion on this

jstourac:
/override ci/prow/images

These are irrelevant for this change.

.editorconfig
jiridanek: I wanted to also add the `EditorConfig is awesome` part, but since I got lgtms already, I will merge what I have. Thanks for taking a look.

PR #811, opened 2024-12-17
Make the pipfile renewal to work when cron firing

atheo89:
Related to: https://issues.redhat.com/browse/RHOAIENG-15393

## Description

Didn't specify what the gha should do in case of scheduled firing and that's why we got that failure: https://github.com/opendatahub-io/notebooks/actions/runs/12362324472/job/34501399571#step:6:65 

Also changed the cron to Wenesday to be closer to release day. 

With this update now it works as expected. 

## How Has This Been Tested?
Tested localy: https://github.com/atheo89/notebooks/actions/runs/12374398338 
image

## Merge criteria:

### DIFF
--- a/.github/workflows/piplock-renewal.yaml
+++ b/.github/workflows/piplock-renewal.yaml
@@ -3,9 +3,9 @@
 name: Pipfile.locks Renewal Action
 
 on:  # yamllint disable-line rule:truthy
-  # Triggers the workflow every Monday at 22pm UTC am 0 22 * * 1
+  # Triggers the workflow every Wednesday at 1am UTC
   schedule:
-    - cron: "0 22 * * 1"
+    - cron: "0 1 * * 3"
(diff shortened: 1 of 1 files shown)


.github/workflows/piplock-renewal.yaml
```diff
@@ -3,9 +3,9 @@
 name: Pipfile.locks Renewal Action
 
 on:  # yamllint disable-line rule:truthy
-  # Triggers the workflow every Monday at 22pm UTC am 0 22 * * 1
+  # Triggers the workflow every Wednesday at 1am UTC
```
jstourac, line 6: Maybe slightly earlier time so that at least America has night would be better, but that doesn't really matter in the end.

jstourac:
/lgtm

/override ci/prow/images

this test is irrelevant for this PR

PR #828, opened 2025-01-06
NO-JIRA: fix(build): use yq to modify kustomization.yaml during deploy targets

andyatmiami:
## Description
The current method of using `sed` to modify `kustomization.yaml` when running the `deploy9-` `Makefile` target runs into issues if the developer workstation is on MacOS.

Given the intended behavior here is to swap out a couple values in the `yaml` file - this PR adds `yq` as a self-contained dependency to the `Makefile` to handle this `yaml` modification.

The `bin/yq` target added to `Makefile` follows the existing conventions established around the `bin/kubectl` handling within the `Makefile`.

## How Has This Been Tested?
I implemented this fix as part of my efforts working on RHOAING-12822 which is documented in part here:
- https://gitlab.cee.redhat.com/astonebe/rhoia-onboarding-scripts/-/wikis/Home/Building-and-Testing-Workbench-Images

I have been able to successfully deploy multiple workbench image files while running this fix.  Running any `deploy9-` or `deploy-c9s` target will invoke the modified code.

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,8 @@ OS_ARCH=$(shell go env GOOS)/$(shell go env GOARCH)
 IMAGE_TAG		 ?= $(RELEASE)_$(DATE)
 KUBECTL_BIN      ?= bin/kubectl
 KUBECTL_VERSION  ?= v1.23.11
+YQ_BIN      ?= bin/yq
+YQ_VERSION  ?= v4.44.6
 NOTEBOOK_REPO_BRANCH_BASE ?= https://raw.githubusercontent.com/opendatahub-io/notebooks/main
 REQUIRED_RUNTIME_IMAGE_COMMANDS="curl python3"
 REQUIRED_CODE_SERVER_IMAGE_COMMANDS="curl python oc code-server"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -285,17 +287,28 @@ ifeq (,$(wildcard $(KUBECTL_BIN)))
 	@chmod +x $(KUBECTL_BIN)
 endif
 
+# Download yq binary
+.PHONY: bin/yq
```
andyatmiami, line 291: One point of interest here (as this merely follows the pattern established from `bin/kubectl`) ...

I wonder if we should actually **NOT** declare a `.PHONY` target here?  as the `bin/yq` (and likewise `bin/kubectl`) are **actual** files on the filesystem... such that we could simply allow `make` to _do what it does_ ?

pros:
- wouldn't run and download "stuff" on every invocation

cons:
- users would ned to manually intervene if we needed to upgrade the version of the binary we pull down

jiridanek:
> I wonder if we should actually NOT declare a .PHONY target here? as the bin/yq (and likewise bin/kubectl) are actual files on the filesystem... such that we could simply allow make to do what it does ?

Subsequent PR, maybe? and change in both places? IMO its phony because somebody did not want to deal with Makefile details and just used it as task runner

andyatmiami:
@daniellutz had raised a concern that introducing `yq` here could necessitate additional paperwork documenting `yq` as a "build tool" to _satisfy_ IBM/RH policies and processes.

In discussion during scrum - @harshad16 stated that while we are on `ci-operator` - this is not an issue.  While it **WILL** be more of a "front of mind" concern when we move to Konflux - for the time being we can proceed with this change.

PR #834, opened 2025-01-12
RHOAIENG-16587: fix(test): ensure papermill tests run successfully for all supported notebooks

andyatmiami:
## Description
Our testing framework had a fundamental issue in that it would, for certain images, naively run the test suite of parent images for a child image.  In the case of `tensorflow` images, this caused problems because the (child) `tensorflow` image would **purposefully** _downgraded_ a package version given compatability issues. Furthermore, when attempting to verify the changes to address this core issue, numerous other bugs were encountered that required to be fixed.

Significant changes introduced in this commit:
- logic of the `test-%` Makefile target was moved into a shell script named `test_jupyter_with_papermill.sh`
- test resources required to be copied to pod are now retrieved via the locally checked out repo
    - previously there were pulled from a remote branch via `wget` (defaulting to `main` branch)
    - this ensure our PR checks are now always leveraging any updated files
- test_notebook.ipynb files now expect an `expected_versions.json` file to exist within the same directory.
    - `expected_versions.json` file is derived from the relevant `...-notebook-imagestream.yaml` manifest and leveraged when asserting on dependency versions
	- admittedly the duplication of various helper functions across all the notebook files is annoying - but helps to keep the `.ipynb` files self-contained
- `...-notebook-imagestream.yaml` manifest had annotations updated to include any package that had a unit test in `test_notebook.ipynb` asserting versions
- CUDA tensorflow unit test that converts to ONNX was updated to be actually functional

Minor changes introduced in this commit:
- use `ifdef OS` in Makefile to avoid warnings about undefined variable
- all `test_notebook.ipynb` files:
    - have an `id` attribute defined in metadata
    - specify the `nbformat` as `4.5`
- the more compute-intensive notebooks had `readinessProbe` and `livenessProbe` settings updated to be less aggressive
    - was observing liveness checks sporadically failing while the notebook was being tested - and this update seemed to help
- `trustyai` notebook now runs the minimal and datascience papermill tests (similar to `tensorflow` and `pytorch`) instead of including the test code within its own `test_noteook.ipynb` file
- various "quality of life" improvements where introduced into `test_jupyter_with_papermill.sh`
    - properly invoke tests for any valid/supported target
        - previously certain test targets required manual intervention in order to run end to end
    - improved logging (particularly when running with the `-x` flag)
    - more modular structure to hopefully improve readability
    - script now honors proper shell return code semantics (i.e. returns `0` on success)

It should also be noted that while most `intel` notebooks are now passing the papermill tests - there are issues with the `intel` `tensorflow` unit tests still.  More detail is captured in the JIRA ticket related to this commit.  However, we are imminently removing `intel` logic from the `notebooks` repo entirely... so I didn't wanna burn any more time trying to get that last notebook to pass as it will be removed shortly!

Further details on `expected_versions.json`:
- `yq` (which is installed via the `Makefile` is used to:
    - query the relevant imagestream manifest to parse out the `["opendatahub.io/notebook-software"]` and `["opendatahub.io/notebook-python-dependencies"]` annotations from the first element of the `spec.tags` attribute
    - inject name/version elements for `nbdime` and `nbgitpuller` (which are asserted against in `minimal` notebook tests)
    - convert this `yaml` list to a JSON object of the form: `{<package>: <version>}`
- this JSON object is then copied into the running notebook workload in the same directly that `test_notebook.ipynb` resides
- each `test_notebook.ipynb` has a couple helper functions defined to then interact with this file:
    - `def load_expected_versions() -> dict:`
    - `def get_expected_version(dependency_name: str) -> str:`
        - The argument provided to the `get_expected_version` function should match the `name` attribute of the JSON structure defined in the imagestream manifest

Related-to: https://issues.redhat.com/browse/RHOAIENG-16587

## How Has This Been Tested?
Testing can be performed simply by invoking `make` with a target that matches the `test-%` pattern-matching rule.

ex: `gmake test-jupyter-pytorch-ubi9-python-3.11 -e NOTEBOOK_REPO_BRANCH_BASE=https://raw.githubusercontent.com/andyatmiami/notebooks/fix-e2e-asserts` for basic sanity checking.  However, all known/supported/expected `test-%` patterns should be verified.

I have also created a utility in https://gitlab.cee.redhat.com/astonebe/notebook-utils that can be used to more efficiently test notebooks "in bulk".  While it still runs sequentially, it handles deploying, testing, and undeploying a given list of targets.
- I used this utility script to test all notebooks that have a corresponding imagestream manifest.

## Merge criteria:


jiridanek:
## Alternative proposal

The package versions we should have in our images are the ones given in the docs, which is at https://access.redhat.com/articles/rhoai-supported-configs (section "Supported workbench images"). This table is generated (or written manually, before https://github.com/red-hat-data-services/notebooks/pull/476), based on our ImageStream manifests at https://github.com/opendatahub-io/notebooks/tree/main/manifests/base (and https://github.com/red-hat-data-services/notebooks/tree/main/manifests/base for RHOAI)

https://github.com/opendatahub-io/notebooks/blob/8a24d011bdb554cb9c423cc17acb74a0a2b5cd63/manifests/base/jupyter-rocm-pytorch-notebook-imagestream.yaml#L27-L33

We should not duplicate that information in the test ipython notebook files. The tests should check that what's in the image corresponds to what's given in the manifests.

andyatmiami:
@jiridanek .. while i do agree with you.. the documentation is wrong (at least spot-checking in the case of `pytorch` for `rocm`). 

Should be version `2.2` as evidenced by this failure...

```
- rocm-jupyter-pytorch-ubi9-python-3.9
AssertionError: '2.4' != '2.2'
- 2.4
?   ^
+ 2.2
?   ^
 : incorrect version
 ```
 
 So, given I question the source of truth (presently) and given **today** in our tests we **are** duplicating this information...
 
 do you want me to address your suggestion in **this** PR or a follow up PR?
 
Again, for emphasis - I wholly agree with your proposal - I've just scope-creeped the heck outta this present PR already 😎

**Comment by jiridanek** [comment]
> do you want me to address your suggestion in this PR or a follow up PR?

We did not do enough PRs of this kind for me to discern a pattern, but on my previous team, there was preference for moving in small steps, e.g. first move script embedded in Makefile to it's own file unchanged, then modify said script.

Guess this is largely a mater of personal taste. And also of review speed, it's hard to split things up when reviews for the parts are not forthcoming.

**Comment by jstourac** [comment]
I got to this for a brief review. I agree with Jiri to re-use what we define in our manifest files in imagestreams for the testing purposes instead of duplicating - but yeah, let's do it step by step as you wrote above. So, I am happy to see these tests being extracted from the Makefile making it a bit more readable. What I'm not a big fan here is that we're creating yet another bash script for this. I think we should utilize the existing pytest framework with the testcontainers for as much tests as possible so that we can run these in a unified environment.

Anyway, since this aims only to extract these out of Makefile and to keep this compatible with the existing approach, I'm not against this in general right now. But to the future, I think we should think to migrate this again - this time to the pytest where most of our tests will be, hopefully.

**Comment by jiridanek** [comment]
I'm with Jan, pulling the bash code out of Makefile is a good improvement in itself.

**Comment by andyatmiami** [comment]
/retest ci/prow/notebooks-ubi9-e2e-tests

(weird GCP errors on building the cluster - hopefully transient 🤞 )

**Comment by andyatmiami** [comment]
/test ci/prow/notebooks-ubi9-e2e-tests

(weird GCP errors on building the cluster - hopefully transient 🤞 )

**Comment by andyatmiami** [comment]
/test notebooks-ubi9-e2e-tests

(weird GCP errors on building the cluster - hopefully transient 🤞 )

**Comment by andyatmiami** [comment]
**PASSING TESTS!**

image

File: jupyter/intel/ml/ubi9-python-3.11/kustomize/base/statefulset.yaml [hunk 1]
```diff
@@ -36,17 +36,19 @@ spec:
...
           readinessProbe:
             httpGet:
               path: /notebook/opendatahub/jovyan/api
               port: notebook-port
               scheme: HTTP
-            initialDelaySeconds: 10
-            periodSeconds: 5
+            initialDelaySeconds: 15
+            periodSeconds: 10
+            timeoutSeconds: 5
```
jiridanek, line 51: in production, these timeouts come from dashboard; maybe we should use what dashboard sets?

https://github.com/opendatahub-io/odh-dashboard/blob/440d8fef243d9ae78ed5a3a4c51002d655521cb6/frontend/src/api/k8s/notebooks.ts#L160-L176

scripts/test_jupyter_with_papermill.sh
```diff
@@ -0,0 +1,430 @@
...
+        nbdime_version='3.2'
+    fi
+    local nbgitpuller_version='1.2'
+
+    expected_versions=$("${yqbin}" '.spec.tags[0].annotations | .["opendatahub.io/notebook-software"] + .["opendatahub.io/notebook-python-dependencies"]' "${test_version_truth_filepath}" |
+        "${yqbin}" -N -p json -o yaml | 
+        nbdime_version=${nbdime_version} nbgitpuller_version=${nbgitpuller_version} "${yqbin}" '. + [{"name": "nbdime", "version": strenv(nbdime_version)},{"name": "nbgitpuller", "version": strenv(nbgitpuller_version)}]' |
+        "${yqbin}" -N -o json '[ .[] | (.name | key) = "key" | (.version | key) = "value" ] | from_entries')
+
+    # Following disabled schellcheck intentional as the intended behavior is for those ${1}, ${2} variables to only be expanded when running within kubernetes
```
jiridanek, line 263:
```suggestion
    # Following disabled shellcheck intentional as the intended behavior is for those ${1}, ${2} variables to only be expanded when running within kubernetes
```

jupyter/intel/ml/ubi9-python-3.11/kustomize/base/statefulset.yaml
andyatmiami: Dashboard also takes a user-supplied configured value for `resources`:
- https://github.com/opendatahub-io/odh-dashboard/blob/440d8fef243d9ae78ed5a3a4c51002d655521cb6/frontend/src/api/k8s/notebooks.ts#L150

And at some level.. the amount of compute resources available to the workload are going to influence the liveness/readiness probe performance..

I don't personally believe its a good apples-to-apples comparison..

That being said - I didn't spend time determining which "lever" of the problems was the problem - I suspect its actually the `timeoutSeconds` - as compute heavy workloads like TF, PYT, and Intel-ML (whatever that is 😎 ) do a lot of initial _churning_ when firing up...

`timeoutSeconds: 1` is the default behavior.. so I fear following your recommendation would simply "revert" my change.. and I have plenty of evidence from running these tests to know without any changes.. these workloads **WILL DIE** leading to flaky tests..

However, if the desire is to keep these properties as close to Dashboard as possible - I can spend a little more time profiling to perhaps land on something like `timeoutSeconds: 3 | 4` (with all other values unchanged) will keep things consistently passing... I do think some deviation from the Dashboard probe settings is unavoidable though (just to set expectations)

jiridanek: I find this number-fiddllng to be deeply suspicious, that's my whole and entire basis for questioning it.

In Dashboard, users are given just a few resources presets by default. Admins can add more presets, not sure if they do. There is no ui to edit presets, has to be done in yamls. Resources don't configure the probe timeouts, timeouts is a separate (hardcoded) configuration.

> the more compute-intensive notebooks had readinessProbe and livenessProbe settings updated to be less aggressive
was observing liveness checks sporadically failing while the notebook was being tested - and this update seemed to help

Should we adjust the production deployment probes too, then? In Dashboard codebase? Imagine customer does something compute-intensive in their workbench, and it gets killed by probe under their hands.

.gitignore
```diff
@@ -128,6 +128,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+.DS_store
```
jiridanek, line 131: 🍎

jiridanek:
/lgtm

this is a huge improvement over what we had before, so yeah, I'm happy to have it merged

atheo89 (commented):
Great improvement Andy! I just added few comments

scripts/test_jupyter_with_papermill.sh
```diff
@@ -0,0 +1,430 @@
...
+##      test_jupyter_with_papermill.sh <makefile test target>
+##          - Intended to be invoked from the test-% target of the Makefile
+##          - Arguments
+##              - <makefile test target>
+##                  - the resolved wildcard value from the Makefile test-% pattern-matching rule
+##  
+##
+
+
+set -uox pipefail
```
jiridanek, line 41:
```suggestion
-Eeuxo pipefail
```

https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

jstourac (commented):
/lgtm

manifests/base/jupyter-pytorch-notebook-imagestream.yaml
```diff
@@ -27,6 +27,7 @@ spec:
         # language=json
         opendatahub.io/notebook-python-dependencies: |
           [
+            {"name": "JupyterLab","version": "4.2"},
```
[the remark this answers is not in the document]
andyatmiami: Found by the tests once I changed them to rely on the `-notebook-imagestream.yaml` as the source of truth for driving version assertions...

scripts/test_jupyter_with_papermill.sh
andyatmiami: I had deliberately **not** included `-e` initially...  I have "mixed feelings" on `-e` 's usefulness... as there are caveats/edge cases on when it doesn't actually do anything I still don't have a great handle on..

I am happy to add this in if this is a strong preference - but will then need to do a couple more passes to retest script to make sure do regression in behavior is introduced...

jupyter/intel/ml/ubi9-python-3.11/kustomize/base/statefulset.yaml
jiridanek: I agree with the no-wrong-answer comment. Do what seems appropriate. Since these are changes in tests, it does not change the product itself. "pretend-nothing-happened-and-move-on" is a valid option, since customers won't be affected

jiridanek: Creating Jira sounds right too, assuming we actually want to be looking into this soonish. Otherwise I'd prefer to keep jira clean of nebulous ideas and track _real_ _issues_.

scripts/test_jupyter_with_papermill.sh
jiridanek: We used to have -e in the Makefile. I'm sad to see it go, it feels like an achievement lost (because I added it there ;p) but I don't do much of Bash so I don't have hard Bash preferences.

andyatmiami: Based on discussions in daily standup - for **NOW** we will proceed without seeing any `-e` / `-E` flag.

This is based on the fact I have not seen the need for it based on my repeated runs of the test script - which seems to have a reasonable outcome in both success and failure cases I have encountered.

Happy to revisit this decision in the future as a specific need/shortcoming arises.

jupyter/intel/ml/ubi9-python-3.11/kustomize/base/statefulset.yaml
andyatmiami: For now, with emphasis on tests consistently passing.. I am going to merge this PR as-is (to then also unblock the removal of intel image builds).

However, I will spend some time today capturing data on the tests that sporadically fail with the original probe settings and spin that out into a JIRA issue we can review with Dashboard team to get alignment and/or explicit understanding here on...

Or rather I just write up a document to first review with Dashboard team - and based on their input/preference - we can decide on if a JIRA is warranted...  i.e. I won't make this an issue until we get confirmation its a "real issue" (as I like that suggestion!)

andyatmiami:
/override intel-notebooks-e2e-tests  notebooks-ubi9-e2e-tests rocm-notebooks-e2e-tests

andyatmiami:
/override ci/prow/intel-notebooks-e2e-tests ci/prow/notebooks-ubi9-e2e-tests ci/prow/rocm-notebooks-e2e-tests

andyatmiami:

(same old build timeout BS.. i'm working on getting to the bottom of that!)

andyatmiami:
/override ci/prow/notebook-jupyter-intel-tf-ubi9-python-3-11-pr-image-mirror 

(same old quay 502 issues and retries not handling it)
```
error: unable to copy layer sha256:45d40005ddd1bccd29a7b1abc2a88c08732ab10f9904c997cc0c322f77319bce to quay.io/opendatahub/workbench-images: received unexpected HTTP status: 502 Bad Gateway
info: Mirroring completed in 33.41s (4.343MB/s)
error: one or more errors occurred while uploading images
2025-01-20T17:08:23+00:00 ERROR Unable to mirror image
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2025-01-20T17:08:23Z"}
error: failed to execute wrapped command: exit status 1
```

andyatmiami:
/override ci/prow/notebook-jupyter-pytorch-ubi9-python-3-11-pr-image-mirror 

(same old build timeout BS.. i'm working on getting to the bottom of that!)

andyatmiami:
/override ci/prow/notebook-rocm-jupyter-pyt-ubi9-python-3-11-pr-image-mirror 

(same old build timeout BS.. i'm working on getting to the bottom of that!)

andyatmiami:
/override ci/prow/notebook-rocm-jupyter-tf-ubi9-python-3-11-pr-image-mirror 

(same old build timeout BS.. i'm working on getting to the bottom of that!)

andyatmiami:
/override ci/prow/images

(same old build timeout BS.. i'm working on getting to the bottom of that!)

PR #868, opened 2025-01-27
RHOAIENG-18459: chore(tests/containers/workbenches): make the ipv6 listening test work on macOS

jiridanek:
* https://issues.redhat.com/browse/RHOAIENG-9707
* https://issues.redhat.com/browse/RHOAIENG-18459

follow-up for
* https://github.com/opendatahub-io/notebooks/pull/866

## Description

Uses the pydantic library to load output from `podman machine info` and sets-up a ssh tunnel from the podman machine vm to the machine running the test.

This is a workaround for macOS/rootful issues in Podman
* https://github.com/containers/podman/issues/14491
* https://github.com/containers/podman/issues/15140

## How Has This Been Tested?

* works on my macOS
* https://github.com/jiridanek/notebooks/actions/runs/12986604939

```
poetry run pytest tests/containers \
    -k 'TestWorkbenchImage and test_ipv6_only' \
    --image=quay.io/opendatahub/workbench-images:codeserver-ubi9-python-3.11-20250124-6a93e34
```

## Merge criteria:

### DIFF
--- a/tests/containers/podman_machine_utils.py
+++ b/tests/containers/podman_machine_utils.py
@@ -2,19 +2,22 @@ import json
 import logging
 import socket
 import subprocess
-from tests.containers.schemas import PodmanMachineInspect
+from typing import Callable
+
+import tests.containers.schemas
 
 logging.basicConfig(level=logging.DEBUG)
 
-def open_ssh_tunnel(machine_name: str, local_port: int, remote_port: int, remote_interface: str = "localhost") -> subprocess.Popen:
+def open_ssh_tunnel(machine_predicate: Callable[[tests.containers.schemas.PodmanMachine], bool],
+                    local_port: int, remote_port: int, remote_interface: str = "localhost") -> subprocess.Popen:
     # Load and parse the Podman machine data
-    json_data = subprocess.check_output(["podman", "machine", "inspect", machine_name], text=True)
-    inspect = PodmanMachineInspect(machines=json.loads(json_data))
+    json_data = subprocess.check_output(["podman", "machine", "inspect"], text=True)
+    inspect = tests.containers.schemas.PodmanMachineInspect(machines=json.loads(json_data))
     machines = inspect.machines
 
-    machine = next((m for m in machines if m.Name == machine_name), None)
+    machine = next((m for m in machines if machine_predicate(m)), None)
     if not machine:
-        raise ValueError(f"Machine '{machine_name}' not found")
+        raise ValueError(f"Machine matching given predicate not found")
 
     ssh_command = [
         "ssh",
@@ -30,7 +33,7 @@ def open_ssh_tunnel(machine_name: str, local_port: int, remote_port: int, remote
     # Open the SSH tunnel
     process = subprocess.Popen(ssh_command)
 
-    logging.info(f"SSH tunnel opened for {machine_name}: {remote_interface}:{local_port} -> localhost:{remote_port}")
+    logging.info(f"SSH tunnel opened for {machine.Name}: {remote_interface}:{local_port} -> localhost:{remote_port}")
     return process
 
 def find_free_port() -> int:
@@ -46,8 +49,8 @@ def find_free_port() -> int:
 # Usage example
 if __name__ == "__main__":
     tunnel_process = open_ssh_tunnel(
-        "podman-machine-default",
-        8080, 8080,
+        machine_predicate=lambda m: m.Name == "podman-machine-default",
+        local_port=8080, remote_port=8080,
         remote_interface="[fc00::2]"
     )
 
--- a/tests/containers/podman_machine_utils.py
+++ b/tests/containers/podman_machine_utils.py
@@ -1,0 +1,59 @@
+import json
+import logging
+import socket
+import subprocess
+from tests.containers.schemas import PodmanMachineInspect
+
+logging.basicConfig(level=logging.DEBUG)
+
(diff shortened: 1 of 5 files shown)


tests/containers/podman_machine_utils.py
```diff
@@ -0,0 +1,59 @@
...
+    with socket.socket(socket.AF_INET6, socket.SOCK_STREAM) as s:
+        s.bind(('', 0))  # Bind to a free port provided by the system
+        s.listen(1)
+        port = s.getsockname()[1]
+    return port
+
+# Usage example
+if __name__ == "__main__":
+    tunnel_process = open_ssh_tunnel(
+        "podman-machine-default",
```
andyatmiami, line 49: not sure of a good/right place to document this.. but we should ideally call out (for purposes of running locally) - we **require** a machine named `podman-machine-default` to exist.  I realize it errors _after the fact_ appropriately in this script... but this is codifying a requirement worth calling out in a readme or something maybe?

(as a case in point - I don't (as of now - but will soon!) have a `podman-machine-default` machine - as its too underpowered to run many notebooks)

tests/containers/schemas.py
```diff
@@ -0,0 +1,41 @@
+from pydantic import BaseModel
```
andyatmiami, line 1: Did you look into https://podman-py.readthedocs.io/en/stable/index.html ?  Can we get some of this boilerplate "for free" if we use the Podman Python SDK? (or, conversely, is the Podman Python SDK "more hassle than its worth" ?)

jiridanek: there's quite a lot of various docker/podman libraries like this

since I went with testcontainers (which uses docker-py) originally, I intended to keep to docker-compatible apis; but, turns out, more is oftentimes needed

besides testcontainers, there's all that Adriel mentioned at one of the qe meetings
* https://terratest.gruntwork.io/examples/
* https://github.com/pytest-dev/pytest-testinfra
* https://dagger.io/

there are other stuff, for example something from red hat people that runs `podman` commands as a subprocess; so whatever you script in python, you have an obvious way to reproduce on commandline

jiridanek: I don't think that podman machine api is exposed there... but I'd have to check more thoroughly to find out.

I spent few minutes and no joy so far. I really think podman-machine not part of the libpod api.

tests/containers/podman_machine_utils.py
jiridanek: so, you have a differently named podman machine? in that case this should probably detect whatever machine you're using and use that, wdyt?

andyatmiami: yes, perhaps if we just had the system choose the currently active connection - that makes it "one less stumbling block"

```
➜ ~/ $ podman system connection ls
Name                URI                                                         Identity                                                        Default     ReadWrite
rhoai-default       ssh://core@127.0.0.1:50943/run/user/501/podman/podman.sock  /Users/astonebe/.local/share/containers/podman/machine/machine  false       true
rhoai-default-root  ssh://root@127.0.0.1:50943/run/podman/podman.sock           /Users/astonebe/.local/share/containers/podman/machine/machine  true        true
```

seems `podman system connection ls` can give you a data structure that lists which machine is currently `default`

tests/containers/schemas.py
andyatmiami: totally fine - not worth spending much time on it - appreciate you sharing thoughts/motivation behind the question... you can consider this conversation resolved.

jiridanek: it's a good thought, cc @jstourac maybe we'd just stop pretending we use docker and will fully embrace podman

there's also https://issues.redhat.com/browse/RHOAIENG-18060 that asks to write these test so that they can be run on docker/podman and also openshift; that's gonna be another challenge, but maybe a lesser one than all this macOS stuff

tests/containers/podman_machine_utils.py
jiridanek: very good point, I added a commit here that addresses that; I look what the `realpath docker_socket` is, and compare that with the sockets given in `podman machine info`. I did not need podman system connections ls after all

tests/containers/workbenches/workbench_image_test.py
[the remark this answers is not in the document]
jiridanek: /override ci/prow/images
this is my override thread

daniellutz:
I tested here and looks good!

/lgtm

andyatmiami:
/lgtm

I have confirmed if I have a machine named `podman machine default` - the given test command succeeds:

```
poetry run pytest tests/containers \
    -k 'TestWorkbenchImage and test_ipv6_only' \
    --image=quay.io/opendatahub/workbench-images:codeserver-ubi9-python-3.11-20250124-6a93e34
```

```
...
PASSEDDEBUG:urllib3.connectionpool:http://localhost:None "DELETE /v1.41/networks/e252ab284eaa64991cb4b40e68e99cd337f9723544d308caadd7de65b0a6c215 HTTP/1.1" 204 0


=================================================================================================== 1 passed, 4 deselected in 199.02s (0:03:19) ===================================================================================================
```

ℹ️ I am on MacOS

tests/containers/podman_machine_utils.py
jiridanek: And, turned out this does not really fix anything with 1+ podman machines present. That's because `podman machine inspect` without machine name inspects the default machine only. Followup pr needed.

PR #870, opened 2025-01-28
RHOAIENG-17251: add a test for the proxy env configuration in RStudio + some other minor changes

jstourac:
## Description

https://issues.redhat.com/browse/RHOAIENG-17251

## How Has This Been Tested?

```
DOCKER_HOST=unix:///run/user/$UID/podman/podman.sock poetry run pytest tests/containers --image quay.io/opendatahub/workbench-images@sha256:e98d19df346e7abb1fa3053f6d41f0d1fa9bab39e49b4cb90b510ca33452c2e4
```

## Merge criteria:


jiridanek:
lgtm, just

here's what I'd do in addition

i'm a big fan of table driven tests (https://dave.cheney.net/2019/05/07/prefer-table-driven-tests#Introducing-table-driven-tests), so

```python
    @allure.issue("RHOAIENG-16604")
    def test_http_proxy_env_propagates(self, subtests, image: str) -> None:
        """
        This checks that the lowercased proxy configuration is propagated into the RStudio
        environment so that the appropriate values are then accepted and followed.
        """
        skip_if_not_rstudio_image(image)

        class TestCase(NamedTuple):
            name: str
            name_lc: str
            value: str

        test_cases: list[TestCase] = [
            TestCase("HTTP_PROXY", "http_proxy", "http://localhost:8080"),
            TestCase("HTTPS_PROXY", "https_proxy", "https://localhost:8443"),
            TestCase("NO_PROXY", "no_proxy", "google.com"),
        ]

        container = WorkbenchContainer(image=image, user=1000, group_add=[0])
        for tc in test_cases:
            container.with_env(tc.name, tc.value)

        try:
            # We need to wait for the IDE to be completely loaded so that the envs are processed properly.
            container.start(wait_for_readiness=True)

            # Once the RStudio IDE is fully up and running, the processed envs should include also lowercased proxy configs.
            for tc in test_cases:
                with subtests.test(tc.name):
                    output = check_output(
                        container,
                        f"/usr/bin/R --quiet --no-echo -e 'Sys.getenv(\"{tc.name_lc}\")'")
                    assert '"' + tc.value + '"' in output
        finally:
            docker_utils.NotebookContainer(container).stop(timeout=0)
```

one advantage of subtest is that when variable is missing, it will also check all the other variables before failing the test

it also makes things simpler if there's additional variable to be checked, just add it to the test_cases table

jiridanek:
/lgtm
in case you discover you aren't fan of table driven tests ;p

jstourac:
@jiridanek thanks for your review - I updated based on your comment. Please, check again.

tests/containers/workbenches/rstudio/rstudio_test.py
```diff
@@ -153,3 +186,10 @@ def default(self, o):
 
 
 _ = StructuredMessage  # optional shortcut, to improve readability
+
+
+class TestCase():
+    def __init__(self, name, name_lc, value):
+        self.name = name
+        self.name_lc = name_lc
+        self.value = value
```
jiridanek, line 195: I'd nest this under `def test_`, because this only applies for that one test and no other

and using `NamedTuple`, or `@dataclass` is better here, imo; avoids having to repeat the field names twice

* https://docs.python.org/3.12/tutorial/classes.html#odds-and-ends
* https://docs.python.org/3/library/typing.html#typing.NamedTuple

compared do my version, you dropped the `:str` type annotations ;p

jstourac: good point, updated

jstourac:
Thank you for your review and tips, @jiridanek !

/approve

PR #871, opened 2025-01-29
RHOAIENG-9707: chore(tests/containers): check shared objects with ldd

jiridanek:
* https://issues.redhat.com/browse/RHOAIENG-9707
* https://issues.redhat.com/browse/RHOAIENG-18904

This sanity check will report any ELF files with dynamic linking that have unsatisfied dependencies.

* apply the `"LD_LIBRARY_PATH": os.path.dirname(dlib)` fix/workaround

cuda image already has some LD_LIBRARY_PATH, this is to be preserved

```
(app-root) sh-5.1$ echo $LD_LIBRARY_PATH
/usr/local/nvidia/lib:/usr/local/nvidia/lib64
```

* use subtests to handle multiple libs not found

* skip known issue and false positives

## Description

Running `ldd` one by one on every file is quite lengthy on macOS, due to cross-architecture emulation

```
(app-root) bash-5.1$ time bash -c 'i=0; while [[ $i -lt 100 ]]; do ldd /bin/bash >/dev/null; i=$((i+1)); done'

real    0m5.684s
user    0m4.527s
sys     0m0.795s
```

compare that with native linux vm, without this cross-arch translation

```
[jdanek@lima-default workbenches]$ time bash -c 'i=0; while [[ $i -lt 100 ]]; do ldd /bin/bash >/dev/null; i=$((i+1)); done'

real    0m0.220s
user    0m0.130s
sys     0m0.061s
```

Since there's 1000 to 2000 dynamic libs to check in each of our images, the check on a macOS when using ldd can take 70 or more seconds.

When mentioned on a daily meeting, ppl weren't very concerned, so I'm not going to optimize this now.

For the future, already tried out one approach, using the debug/elf package in Go's standard library to simply read the DT_NEEDED field out of the libraries and check that. It's quite a lot of code, but it seems to work, and work fast.  
[check_elf_file_2.go.txt](https://github.com/user-attachments/files/18584127/check_elf_file_2.go.txt)

Alternatively, it should be possible to just go around loading the dlls from the python script directly, to see if it loads or not

```python
            from ctypes import CDLL
            try:
                CDLL(lib_path)
                print(f"Successfully loaded: {lib}")
            except Exception as e:
                print(f"Error loading {lib}: {e}")
```

Question is whether binaries can be casually dlopened like this. I did this in skupper-router (dlopening the current executable ;) so it is possible sometimes, just don't know if always; worry that binary has to be compiled with some special options.

## How Has This Been Tested?

* https://github.com/jiridanek/notebooks/actions/runs/13026586540

Checked all what it found by hand and populated an allowlist to suppress false positives.

Over all, the usefulness of this check will be best seen when we do version upgrades or some significant version upgrades for 2025.a images. Is this going to be full of false positives for trivialities, or is it going to save us from dll hell? So far it found a true positive https://issues.redhat.com/browse/RHOAIENG-18904, but I do worry that the maintenance of the allowlist is not going to be worth it.

Possibly running `import xyz` for every python package we have is more useful. Adding to that small helloworld script to check functionality would be even more powerful.

## Merge criteria:

### DIFF
--- a/tests/containers/base_image_test.py
+++ b/tests/containers/base_image_test.py
@@ -1,10 +1,16 @@
 from __future__ import annotations
 
+import binascii
+import inspect
+import json
 import logging
 import pathlib
+import re
 import tempfile
-from typing import TYPE_CHECKING
+import textwrap
+from typing import TYPE_CHECKING, Any, Callable
 
+import pytest
 import testcontainers.core.container
 import testcontainers.core.waiting_utils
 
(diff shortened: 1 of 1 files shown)


tests/containers/base_image_test.py
```diff
@@ -20,6 +26,84 @@
...
+            for path in dirs:
+                count_scanned = 0
+                unsatisfied_deps: list[tuple[str, str]] = []
+                for dlib in glob.glob(os.path.join(path, "**"), recursive=True):
+                    # we will visit all files eventually, no need to bother with symlinks
+                    s = os.stat(dlib, follow_symlinks=False)
+                    isdirectory = stat.S_ISDIR(s.st_mode)
+                    isfile = stat.S_ISREG(s.st_mode)
+                    executable = bool(s.st_mode & (stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH))
+                    if isdirectory or not executable or not isfile:
```
jstourac, line 57: I wonder whether we should log/print the skipped/ignored files here with the explanation?

tests/containers/base_image_test.py
```diff
@@ -20,6 +26,84 @@
...
+                        continue
+                    with open(dlib, mode='rb') as fp:
+                        magic = fp.read(4)
+                    if magic != b'\x7fELF':
```
jstourac, line 61: Similar as above - do we want to log/print files like these for the convenience?

tests/containers/base_image_test.py
jiridanek: maybe, but for /opt/app-root this will produce a lot of output that nobody in their right mind will have the energy to read through

cc @opendatahub-io/notebook-devs wdyt? do you like your pytest tests to be vvvverbose or not?

jiridanek: logged ticket for improving output

* https://github.com/opendatahub-io/notebooks/issues/876

jiridanek:
/override ci/prow/images
not a relevant check here

PR #873, opened 2025-01-29
RHOAIENG-9707: chore(tests/containers): try to install the cowsay package to check the python venv is writable

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-9707

## Description

```
  _________________________________________________
 /                                                 \
| Checks that the Python virtualenv in the image is |
|  writable.                                        |
 \                                                 /
  =================================================
                                                 \
                                                  \
                                                    ^__^
                                                    (oo)\_______
                                                    (__)\       )\/\
                                                        ||----w |
                                                        ||     ||
```

## How Has This Been Tested?

* https://github.com/jiridanek/notebooks/actions/runs/13070449624

## Merge criteria:

### DIFF
--- a/tests/containers/base_image_test.py
+++ b/tests/containers/base_image_test.py
@@ -163,6 +163,23 @@ class TestBaseImage:
             finally:
                 docker_utils.NotebookContainer(container).stop(timeout=0)
 
+    def test_pip_install_cowsay_runs(self, image: str):
+        """Checks that the Python virtualenv in the image is writable."""
+        container = testcontainers.core.container.DockerContainer(image=image, user=23456, group_add=[0])
+        container.with_command("/bin/sh -c 'sleep infinity'")
+        try:
+            container.start()
+
+            ecode, output = container.exec(["python3", "-m", "pip", "install", "cowsay"])
+            logging.debug(output.decode())
+            assert ecode == 0
+
+            ecode, output = container.exec(["python3", "-m", "cowsay", "--text", "Hello world"])
+            logging.debug(output.decode())
+            assert ecode == 0
+        finally:
+            docker_utils.NotebookContainer(container).stop(timeout=0)
+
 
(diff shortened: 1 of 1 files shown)


jstourac:
LGTM, just wondering whether we should rather/also check the variant where `python3` isn't used in the command. I mean, when user will simply execute `pip install cowsay` and then `cowsay -t Hello`. Technically these results should be identical as the very same python env should be used in both cases. :thinking:

daniellutz:
from my understanding, the `python3 -m pip install` command would be more reliable, because you are already specifying the desired python version on it, and to me, technically, `python3 -m pip install` and `pip install` would be the same, so no issues regarding that, from my understanding.

this one looks good to me!

/lgtm

jiridanek:
> because you are already specifying the desired python version on it

we're already in a python virtual environment when the image starts, and that sets up PATH to /opt/app-root/bin, where `pip` is present; so although in general case `python3 -m pip` is preferable, and `/opt/app-root/bin/python3 -m pip` is even better, in this case should be all the same.

the setting of python virtual env for auto activation is done through some magic around `BASH_ENV` variable.

```
    # The following echo adds the unset command for the variables set below to the \
    # venv activation script. This is inspired from scl_enable script and prevents \
    # the virtual environment to be activated multiple times and also every time \
    # the prompt is rendered. \
    echo "unset BASH_ENV PROMPT_COMMAND ENV" >> ${APP_ROOT}/bin/activate

# For RHEL/Centos 8+ scl_enable isn't sourced automatically in s2i-core
# so virtualenv needs to be activated this way
ENV BASH_ENV="${APP_ROOT}/bin/activate" \
    ENV="${APP_ROOT}/bin/activate" \
    PROMPT_COMMAND=". ${APP_ROOT}/bin/activate"
```

https://github.com/sclorg/s2i-python-container/blob/master/3.11/Dockerfile.rhel9#L84-L94

I was looking into it for
* https://github.com/opendatahub-io/notebooks/pull/641

jiridanek:
this is only a rebase, reapplying the lgtm myself

PR #884, opened 2025-02-04
NO-JIRA: chore(tests/containers): detect socket for ryuk's use with rootless podman on macOS

jiridanek:
## Description

We should support rootless Podman on macOS.

The disadvantage is that it becomes necessary to stop pretending we work with Docker. While it is possible to send raw api query to `libpod/info`, it's best to just import the podman library and use the `.info()` method.

Either case, the result is untyped dict, so pydantic to the rescue.

* [x] consider adding more logging for the macOS case

## How Has This Been Tested?

Followed steps in README.md to run some tests.

* ~https://github.com/jiridanek/notebooks/actions/runs/13139973981~
* ~https://github.com/jiridanek/notebooks/actions/runs/13152311827~
* https://github.com/jiridanek/notebooks/actions/runs/13152432440

## Merge criteria:


tests/containers/pydantic_schemas/podman_info.py
```diff
@@ -0,0 +1,466 @@
+from __future__ import annotations
```
jstourac, line 1: This comes from somewhere or it's written by you, Jiri? If the source is somewhere else, maybe we should link the source for eventual reference? If not, then I'm just fine.

jstourac (commented):
I put one comment, otherwise LGTM. I can't test it myself, but I trust you it works as expected.

/lgtm

tests/containers/pydantic_schemas/podman_info.py
jiridanek: This is created by smart tooling that was given examples of the json data and it then generates pydantic schema for it. So it's all my work.

jiridanek:
Thanks for the checks. GHA tries linux with rootful podman, and I now also tried lima to check linux rootless podman. Should be good to to.
/approve

PR #894, opened 2025-02-10
NO-JIRA: docs(tests/): add interactive PyTest tutorial

jiridanek:
## Merge criteria:

### DIFF
--- a/tests/pytest_tutorial/test_01_intro.py
+++ b/tests/pytest_tutorial/test_01_intro.py
@@ -1,0 +1,122 @@
+"""Welcome!
+Welcome to our pytest-based testsuite.
+
+This series of tests is designed to help you understand how pytest works,
+and how we use it to test our workbench images."""
+
+
+def test_hello_world():
+    """There are three ways to specify a test in PyTest.
+    This one is the most straightforward.
+
+    In a file that in named test_*.py or *_test.py,
+    define a function named test_*, and pytest will discover it and run it as a test."""
+
+    # Assertions in pytest are usually done with the `assert` keyword.
+    # Pytest does magicks to produce an "expected x but got y"-style message from this.
+    # In normal Python code without pytest, we'd only get a generic AssertionError exception without these details.
+    assert 2 + 2 == 4
+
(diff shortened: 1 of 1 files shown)


tests/pytest_tutorial/test_01_intro.py
```diff
@@ -0,0 +1,122 @@
...
+    def tearDown(self):
+        """This method will be run after each test method"""
+
+    @classmethod
+    def tearDownClass(cls):
+        """This method will be run after all test methods were run."""
+
+
+"""Pytest runs in two phases: test collection and then test execution.
+In the collection phase, all files matching the patter (test_*.py or *_test.py) are imported and scanned
```
jstourac, line 85:
```suggestion
In the collection phase, all files matching the pattern (test_*.py or *_test.py) are imported and scanned
```

jstourac:
Thank you for this, Jiri. But I'm wondering - do we really want this as part of this repository? :thinking: Wouldn't some links to an existing tutorial be enough? :thinking: From the current content, it looks like you plan to extend this. How big do you want this to be?

jiridanek:
> Thank you for this, Jiri. But I'm wondering - do we really want this as part of this repository? 🤔

I originally wanted to do this already in https://github.com/skodjob/odh-e2e. I was learning how to do things with the fabric8 client and this is imo a good way to keep notes,

> Wouldn't some links to an existing tutorial be enough? 🤔

The advantage of doing it this way is that whoever reads it can immediately run the sample tests and modify the code. It can be useful when writing a new test and wanting to use some sample code to start from.

> From the current content, it looks like you plan to extend this. How big do you want this to be?

Pytest basics (essentially what's there already), and kubernetes-py basics, and ocp_resources basics.

jiridanek:
Here's what I intend as continuation of this

[test_02_shiftleft.txt](https://github.com/user-attachments/files/19643249/test_02_shiftleft.txt)
[test_xx_openshift.txt](https://github.com/user-attachments/files/19643250/test_xx_openshift.txt)

PR #901, opened 2025-02-18
RHOAIENG-20088: chore(tests/containers) + GHA CI update: check images size change

jstourac:
## Description
Enhances our existing ci scripts for the params.env and existing
runtimes definitions with the actual image size check - we have preset
expected size for each of the image and we check that this size isn't
changed too much.

With this implementation we are checking the compressed image size to
avoid the image download.

---

Adds a new test to our pytest set to check the given image size compared
to the expected value. In this test we are checking the uncompressed
image size by summing up all the layeres of the image. This image is
usually downloaded on the machine where this is being run already.

---

https://issues.redhat.com/browse/RHOAIENG-20088

## How Has This Been Tested?

Local execution of the:
```
./ci/check-params-env.sh
./ci/check-runtime-images.sh
```
and for the Pytest:
```
for image_ref in $(cat manifests/base/params.env | grep "\-n=" | cut -d "=" -f 2); do DOCKER_HOST=unix:///run/user/$UID/podman/podman.sock poetry run pytest tests/containers --image "${image_ref}" -k test_image_size_change; done
```
for each image separately (e.g. runtimes):
```
DOCKER_HOST=unix:///run/user/$UID/podman/podman.sock poetry run pytest tests/containers --image quay.io/opendatahub/workbench-images@sha256:6a2da12a8bdc9cfcda27f4189827be5fbde4b4b4c4f7d92ac694d9360e3562dc -k test_image_size_change
```

We shall check all the existing images before merging this, though.

## Merge criteria:


tests/containers/base_image_test.py
```diff
@@ -189,6 +189,67 @@ def test_pip_install_cowsay_runs(self, image: str):
...
+    # 2. A change is done to the params.env file or runtimes images definitions, where we update manifest references to a new image.
+    #    Check for this scenario is being done in 'ci/[check-params-env.sh|check-runtime-images.sh]'.
+    size_treshold: int = 100 # in MBs
+    percent_treshold: int = 10
+    def test_image_size_change(self, image: str):
+        """Checks the image size didn't change extensively - treshold is 10% or XXX MB."""
+
+        # Map of image label names with expected size in MBs.
+        expected_image_name_size_map = {
+            "odh-notebook-base-centos-stream9-python-3.11": 1,
```
jstourac, line 204: yeah, just realized I need to finalize these too...

jstourac:
Okay, after some more checks also on the testing PR on my own fork #jstourac/notebooks/pull/7, it looks like the cache we use for our builds may vary between different repository forks. This may affect the resulting size of the image for different users.

After consultation with Jiri, we think that the solution might be either in disabling the cache (well, we don't like it because the builds may take quite a long time to be done - but we haven't try). Another option might be to switch from tag to sha for our base images. This is something that will be done as part of the migration to konflux.

Well, for now - so that we can move on a little bit at least, I'll split this PR into two separate ones - one (this) will be kept only for the pytest based check. The other will be for the bash based checks for GHA CI checking when images are updated in params.env file.

jstourac:
https://github.com/jstourac/notebooks/actions/runs/13545278679

PR #903, opened 2025-02-18
RHOAIENG-18400: chore(konflux): create helper scripts to generate konflux configs

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-18400

## Description

I took the automatic onboarding PR from Konflux as a basis, you can see it here

* https://github.com/opendatahub-io/notebooks/pull/931

Created a Python script that can generate such Tekton pipeline for each of our images. For now I am only committing one such pipeline, for jupyter datascience image.

## How Has This Been Tested?

* https://quay.io/repository/redhat-user-workloads/rhoai-ide-konflux-tenant/jupyter-datascience-ubi9-python-3-11?tab=tags&tag=latest

It can be run like this. I am not running this automatically because for now only one image is onboarded to konflux, and this script creates pipelines for all images we have

```
poetry run ci/cached-builds/konflux_generate_component_build_pipelines.py
```

## Note

* https://github.com/opendatahub-io/notebooks/pull/904

## Merge criteria:


jiridanek:
/override ci/prow/images
n/a for this pr

ci/cached-builds/konflux_generate_component_build_pipelines.py
```diff
@@ -0,0 +1,744 @@
...
+            continue
+
+        deps: list[str] = []
+        frontier = [x for x in direct_deps]
+        while frontier:
+            dep = frontier.pop(0)
+            deps.append(dep)
+            frontier.extend(tree[dep])
+
+        build_container_tasks = []
```
jiridanek, line 710: > While this is absolutely wonderful work to see working - I'm not sure of the benefit of merging this now.... when we have changes that remove chained builds looming...

@andyatmiami Well, that just means we would not need an array here, because there will be only a single build_container task.

I'd be quite happy getting all this working for the refactor/ branch too.

jiridanek: We still need this generating script, because we want to have essentially the same component-pipeline yamls for every image we have.

So it's just tweaks going forward, the overall approach still works imo.

jiridanek: > Was moreso just ( lightly ) probing whether reviewing this now is the best use of time... as having this working (for instance) tomorrow with chained builds doesn't buy us a whole lot... and we have weeks before we will actually be onboarded into Konflux.

ODH-io has been onboarded today, https://redhat-internal.slack.com/archives/C05NXTEHLGY/p1739803343884249

> Personally, I'd love to do a proper review here - but i'm underwater on the Konflux refactor and need to get that wrapped up before EOD Thursday...

I don't think you can get it reviewed and merged by EOD Thursday. We'll have at least one more sprint to play with all this, both your refactor, and my konflux stuff...

> Yep, totally agree here w.r.t the overall approach being a good thing ...

That's all I hoped for, at this stage. Thanks for taking a (brief) look!

jstourac:
I have totally zero knowledge here. LGTM from a brief look and should be similar to what Jiri introduced in opendatahub-io/kubeflow#514.

We should wait merging this after #924 is merged, I guess?

atheo89:
/lgtm

However, I'm curious about where the SHA tag is generated from. 🤔

jiridanek:
> I'm curious about where the SHA tag is generated from.

It's taken from what Konflux generated initially and what it keeps updated using the automated PRs in https://github.com/opendatahub-io/notebooks/tree/main/.tekton

ci/cached-builds/gha_pr_changed_files.py
```diff
@@ -106,5 +114,9 @@ def test_get_build_directory(self):
         directory = get_build_directory("rocm-jupyter-pytorch-ubi9-python-3.11")
         assert directory == "jupyter/rocm/pytorch/ubi9-python-3.11"
 
+    def test_get_build_dockerfile(self):
+        directory = get_build_dockerfile("rocm-jupyter-pytorch-ubi9-python-3.11")
+        assert directory == "jupyter/rocm/pytorch/ubi9-python-3.11/Dockerfile.rocm"
```
jstourac, line 119: Should probably rather be:
```suggestion
        dockerfile = get_build_dockerfile("rocm-jupyter-pytorch-ubi9-python-3.11")
        assert dockerfile == "jupyter/rocm/pytorch/ubi9-python-3.11/Dockerfile.rocm"
```
but it's not important, just for readability a bit.

jstourac:
I put one nitpick comment, but lgtm in general.

/lgtm

PR #909, opened 2025-02-20
Include script on workbench to copy runtime-image json to relevant path

atheo89:
Related to: https://issues.redhat.com/browse/RHOAIENG-20241

## Description

This PR include the following changes:
- Modifies the `elyra.sh` script to copy the content from mount path `/opt/app-root/pipeline-runtimes/` to elyra path `/opt/app-root/share/jupyter/metadata/runtime-images/` this take place when the DS notebook starts up. The reason is to avoid override the runtime images that the user may have import by himself. 

(`quay.io/rh_ee_atheodor/workbench-images@sha256:f2bc105afab97575007a5743e39cc4c045fa276ba8612b6a8241758d48711913`) `custom_v3`

## How Has This Been Tested?

This changes have been tested along with notebook controller (PR soon)  by importing the new data science image generated by this PR 

These are the logs when the datachience notebook when it get spin up on cluster.
The list of .json comes from the mount assigned on the notebook from the notebook-controller 
image

## Merge criteria:

### DIFF
--- a/jupyter/datascience/ubi9-python-3.11/setup-elyra.sh
+++ b/jupyter/datascience/ubi9-python-3.11/setup-elyra.sh
@@ -14,6 +14,11 @@ if [ "$(ls -A /opt/app-root/runtimes/)" ]; then
   cp -r /opt/app-root/runtimes/..data/*.json $(jupyter --data-dir)/metadata/runtimes/
 fi
 
+# Set elyra runtime images json from volume mount
+if [ "$(ls -A /opt/app-root/pipeline-runtimes/)" ]; then
+  cp -r /opt/app-root/pipeline-runtimes/..data/*.json /opt/app-root/share/jupyter/metadata/runtime-images/
+fi
+
 # Environment vars set for accessing ssl_sa_certs and sa_token
 # export PIPELINES_SSL_SA_CERTS="/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 export KF_PIPELINES_SA_TOKEN_ENV="/var/run/secrets/kubernetes.io/serviceaccount/token"
--- a/jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
+++ b/jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
@@ -91,6 +91,7 @@ RUN echo "Installing softwares and packages" && \
     rm -f ./Pipfile.lock && \
     # setup path for runtime configuration
     mkdir /opt/app-root/runtimes && \
+    mkdir /opt/app-root/pipeline-runtimes && \
     # Remove default Elyra runtime-images \
     rm /opt/app-root/share/jupyter/metadata/runtime-images/*.json && \
     # Replace Notebook's launcher, "(ipykernel)" with Python's version 3.x.y \
@@ -102,8 +103,5 @@ RUN echo "Installing softwares and packages" && \
     # Fix permissions to support pip in Openshift environments \
     chmod -R g+w /opt/app-root/lib/python3.11/site-packages && \
     fix-permissions /opt/app-root -P
-
-# Copy Elyra runtime-images definitions and set the version
-COPY ${DATASCIENCE_SOURCE_CODE}/runtime-images/ /opt/app-root/share/jupyter/metadata/runtime-images/
 
 WORKDIR /opt/app-root/src


harshad16 (commented):
great work, 

Shall we put this work , under dont-merge
till this PR is looked into:https://github.com/opendatahub-io/kubeflow/pull/513 ?

if not, lets add back the stuff that is removed in this PR, and we can merge this and later remove the code, when PR 513 is merged.

jupyter/datascience/ubi9-python-3.11/setup-elyra.sh
```diff
@@ -14,6 +14,11 @@ if [ "$(ls -A /opt/app-root/runtimes/)" ]; then
   cp -r /opt/app-root/runtimes/..data/*.json $(jupyter --data-dir)/metadata/runtimes/
 fi
 
+# Set elyra runtime images json from volume mount
+if [ "$(ls -A /opt/app-root/pipeline-runtimes/)" ]; then
+  cp -r /opt/app-root/pipeline-runtimes/..data/*.json /opt/app-root/share/jupyter/metadata/runtime-images/
```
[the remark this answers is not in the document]
harshad16: Ah i see, 
seems like the `..data` do exists
```
(app-root) sh-5.1$ ls /opt/app-root/runtimes/
..2025_02_26_04_36_44.2552146125/ datascience-ubi9-py311.json       rocm-pytorch-ubi9-py311.json      tensorflow-ubi9-py311.json        
..data/                           pytorch-ubi9-py311.json           rocm-tensorflow-ubi9-py311.json   ubi9-py311.json 
```
ok, thanks for pointing.

atheo89:
This PR has been rebased and it is ready for a final review. 

>Shall we put this work , under dont-merge
till this PR is looked into:https://github.com/opendatahub-io/kubeflow/pull/513 ?

@harshad16 it is save to be merged, as i don't remove the: `jupyter/datascience/ubi9-python-3.11/runtime-images/*.json`
Will follow up a new PR to unwire completely the `.jsons` once this is complete  https://github.com/opendatahub-io/kubeflow/pull/513

jupyter/datascience/ubi9-python-3.11/setup-elyra.sh
```diff
@@ -14,6 +14,11 @@ if [ "$(ls -A /opt/app-root/runtimes/)" ]; then
   cp -r /opt/app-root/runtimes/..data/*.json $(jupyter --data-dir)/metadata/runtimes/
 fi
 
+# Set elyra runtime images json from volume mount
+if [ "$(ls -A /opt/app-root/pipeline-runtimes/)" ]; then
```
jstourac, line 18: Probably a nitpick, but wouldn't be a bit safer to check the full directory path from where we are copying the json files?
```
```suggestion
if [ "$(ls -A /opt/app-root/pipeline-runtimes/..data/)" ]; then
```
```

**Comment by jstourac** [comment]
one comment, otherwise LGTM

we may want to create a simply pytest test for this ensuring that the set of expected json files for the runtime images are placed on the expected path - can be done as a separate task I suppose

**Comment by jiridanek** [comment]
not sure if pytest test would be particularly helpful for this, maybe

/lgtm

File: jupyter/datascience/ubi9-python-3.11/setup-elyra.sh (hunk 2 above)
Comment (atheo89) [reply jupyter/datascience/ubi9-python-3.11/setup-elyra.sh:18]: Thanks for the suggestion! I see the reasoning behind checking ..data/ directly, but in the case of a Kubernetes-mounted ConfigMap or Secret, /opt/app-root/pipeline-runtimes/ is already the correct path to check.

Since Kubernetes manages this directory and automatically updates symlinks to point to the latest version in ..data/, checking ..data/ explicitly is not necessary and could introduce unnecessary dependencies on Kubernetes' internal structure.

Comment (jstourac) [reply jupyter/datascience/ubi9-python-3.11/setup-elyra.sh:18]: Well, we're depending on that internal structure already on the very next line in this code anyway :slightly_smiling_face: 

Also, I now realized that this check will always be true won't it be? Since we're creating this directory during the [image build here](https://github.com/opendatahub-io/notebooks/pull/909/files#diff-bf1b39ab6e296ed2806e2d5b1c26436ded10fb59f70b1e94334942808596193dR94). So if I'm not missing anything, the check in its current state doesn't make much sense?

Comment (jiridanek) [reply jupyter/datascience/ubi9-python-3.11/setup-elyra.sh:18]: hmm

```
❯ mkdir empty
❯ [ $(ls -A empty) ] || echo false
false


❯ touch empty/file
❯ [ $(ls -A empty) ] || echo false
```

nothing printed in second case

jstourac: Okay, I see, I didn't read the condition properly. So the condition works as expected this way. Sorry about that.

jstourac:
> not sure if pytest test would be particularly helpful for this, maybe

@jiridanek what do you suggest instead?

jiridanek:
> what do you suggest instead

merge and pray

no, seriously, I think nothing lesser than full e2e with odh-notebook-controller is helpful here, plus maybe playwright to open up the runtimes in elyra afterwards

anything less does not really check anything useful, seems to me; but if i was shown pytest test, maybe i'd be proven wrong

jstourac:
> no, seriously, I think nothing lesser than full e2e with odh-notebook-controller is helpful here, plus maybe playwright to open up the runtimes in elyra afterwards

> anything less does not really check anything useful, seems to me; but if i was shown pytest test, maybe i'd be proven wrong

okay, no objections - my point was to have at least something. Having this in e2e is even better. Thus with our current approach that would be `opendatahub/tests` repo.

atheo89:
>we may want to create a simply pytest test for this ensuring that the set of expected json files for the runtime images are placed on the expected path - can be done as a separate task I suppose 

The cp/mounting process occurs when the notebook starts up within the RHOAI environment, as the notebook-controller incorporates the config map. So, as Jiri mentioned, we may need to test this behavior there.

atheo89:
>There is very simple check that creates kubernetes instance in GHA already, so we would need to extend this configuration properly.

Looks fair approach to me.

atheo89:
Thanks for you review guys i will move this in. 

/approve

jiridanek:
/override ci/prow/notebooks-ubi9-e2e-tests

cluster provisioning seems broken, https://redhat-internal.slack.com/archives/C060A5FJEAD/p1741153298133919

harshad16:
> This PR has been rebased and it is ready for a final review.
> 
> > Shall we put this work , under dont-merge
> > till this PR is looked into:[opendatahub-io/kubeflow#513](https://github.com/opendatahub-io/kubeflow/pull/513) ?
> 
> @harshad16 it is save to be merged, as i don't remove the: `jupyter/datascience/ubi9-python-3.11/runtime-images/*.json` Will follow up a new PR to unwire completely the `.jsons` once this is complete [opendatahub-io/kubeflow#513](https://github.com/opendatahub-io/kubeflow/pull/513)

@atheo89 , removal of runtime has been done at line 106: https://github.com/opendatahub-io/notebooks/commit/8745417c3005d15f0f504cfc4b51d427140d38aa#diff-bf1b39ab6e296ed2806e2d5b1c26436ded10fb59f70b1e94334942808596193dL106
so the workbench image won't have runtime image anymore.
Screenshot from 2025-03-05 16-23-23

atheo89:
Thanks @harshad16 as agreed, I opened this PR to fix this: https://github.com/opendatahub-io/notebooks/pull/944

PR #912, opened 2025-02-20
RHOAIENG-20088: chore(GHA CI): check images size change

jstourac:
Enhances our existing ci scripts for the params.env and existing runtimes definitions with the actual image size check - we have preset expected size for each of the image and we check that this size isn't changed too much.

With this implementation we are checking the compressed image size to avoid the image download.

https://issues.redhat.com/browse/RHOAIENG-20088

## How Has This Been Tested?
Local execution of the:

```
./ci/check-params-env.sh
./ci/check-runtime-images.sh
```

## Merge criteria:

### DIFF
--- a/ci/check-params-env.sh
+++ b/ci/check-params-env.sh
@@ -27,6 +27,12 @@ PARAMS_ENV_PATH="manifests/base/params.env"
 # images we want to have in the `params.env` file.
 EXPECTED_NUM_RECORDS=21
 
+# Size change tresholds:
+# Max percentual change
+SIZE_PERCENTUAL_TRESHOLD=10
+# Max absolute change in MB
+SIZE_ABSOLUTE_TRESHOLD=100
+
 # ---------------------------- DEFINED FUNCTIONS ----------------------------- #
 
 function check_variables_uniq() {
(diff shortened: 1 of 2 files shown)


ci/check-params-env.sh
```diff
@@ -217,6 +247,22 @@ function check_image_variable_matches_name_and_commitref() {
         echo "Image URL points to an incorrect image: expected OPENSHIFT_BUILD_NAME '${expected_build_name}'; actual '${openshift_build_name}'"
         return 1
     }
+
+    # Check the size change constraints now
+    # 1. Percentual size change
+    percent_change=$((100 * actual_img_size / expected_img_size - 100))
```
harshad16, line 253:
```suggestion
    percent_change=$((100 * actual_img_size / (expected_img_size - 100)))
```
is the 100 that is being substracted, is it SIZE_ABSOLUTE_TRESHOLD?

harshad16 (commented):
great work, just one question

ci/check-params-env.sh
jstourac: No no; it's just a substraction so that the number is actually always between -100 and 100 interval. Let's count these examples:
1. actual image size is 150 ; expected image size is 100, then:
```
100 * 150 / 100 = 150% -> we need to substract 100 to have 50% change
```
2. actua limage size is 60 ; expected image size is 100, then:
```
100 * 60 / 100 = 60% -> we need to substract 100 to have -40% change (will be changed to absolute value 40% on the next line)
```

Maybe I can change the expression to this:
```
percent_change=$((100 * (actual_img_size / expected_img_size - 1) ))
```

The effect is still same. Let me know what you prefer and whether this all make sense to you :slightly_smiling_face:

harshad16: thank you for the explanation, it makes total sense now.
no change needed.

harshad16 (commented):
/lgtm

:+1:

PR #878, opened 2025-01-31
Reapply "RHOAIENG-17306, RHOAIENG-17307, RHOAIENG-17308: feat(workbenches): tolerate IPv6 environments in codeserver, jupyterlab and rstudio"

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-17305

Reverts opendatahub-io/notebooks#877
* https://github.com/opendatahub-io/notebooks/pull/877

Which means it's to reapply opendatahub-io/notebooks#827
* https://github.com/opendatahub-io/notebooks/pull/827

PR #913, opened 2025-02-22
[main] Reapply "RHOAIENG-17306, RHOAIENG-17307, RHOAIENG-17308: feat(workbenches): tolerate IPv6 environments in codeserver, jupyterlab and rstudio"

jiridanek:
Getting this PR to be in main

* https://github.com/opendatahub-io/notebooks/pull/878

## Merge criteria:

### DIFF
--- a/rstudio/c9s-python-3.11/setup_rstudio.py
+++ b/rstudio/c9s-python-3.11/setup_rstudio.py
@@ -46,6 +46,30 @@ def _support_arg(arg):
     ret = subprocess.check_output([get_rstudio_executable('rserver'), '--help'])
     return ret.decode().find(arg) != -1
 
+def detect_env():
+    import socket
+    supports_ipv4 = supports_ipv6 = False
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.bind(('127.0.0.1', 0))
+            supports_ipv4 = True
+    except OSError:
+        pass
+    try:
+        with socket.socket(socket.AF_INET6, socket.SOCK_STREAM) as s:
+            s.bind(('::1', 0))
+            supports_ipv6 = True
+    except OSError:
+        pass
+    if supports_ipv4 and supports_ipv6:
+        return '::'  # Dual-stack
+    elif supports_ipv6:
+        return '::'
+    elif supports_ipv4:
+        return '0.0.0.0'
+    else:
+        raise EnvironmentError('No IPv4 or IPv6 support detected.')
+
 def _get_cmd(port):
     ntf = tempfile.NamedTemporaryFile()
 
@@ -60,7 +84,7 @@ def _get_cmd(port):
         '--server-working-dir=' + os.getenv('HOME'),
         '--auth-none=1',
         '--www-frame-origin=same',
-        #'--www-address=0.0.0.0',
+        '--www-address='+ detect_env(),
         '--www-port=' + str(port),
         '--www-verify-user-agent=0',
         '--rsession-which-r=' + get_rstudio_executable('R'),
--- a/tests/containers/workbenches/workbench_image_test.py
+++ b/tests/containers/workbenches/workbench_image_test.py
@@ -49,7 +49,6 @@ class TestWorkbenchImage:
         finally:
             docker_utils.NotebookContainer(container).stop(timeout=0)
 
-    @pytest.mark.skip(reason="RHOAIENG-17305: currently our Workbench images don't tolerate IPv6")
     def test_ipv6_only(self, subtests: pytest_subtests.SubTests, workbench_image: str, test_frame):
         """Test that workbench image is accessible via IPv6.
         Workarounds for macOS will be needed, so that's why it's a separate test."""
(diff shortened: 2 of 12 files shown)


tests/containers/workbenches/workbench_image_test.py
```diff
@@ -203,6 +202,8 @@ def grab_and_check_logs(subtests: pytest_subtests.SubTests, container: Workbench
         "JupyterEventsVersionWarning: The `version` property of an event schema must be a string. It has been type coerced, but in a future version of this library, it will fail to validate.",
         # This is a message from our reverse proxy nginx caused by the fact that we attempt to connect to the code-server before it's actually running (container.start(wait_for_readiness=True)).
         "connect() failed (111: Connection refused) while connecting to upstream, client",
+        # We use oauth-proxy to give us authentication, and we use OpenShift route to get HTTPS
+        "WARNING: The Jupyter server is listening on all IP addresses and not using encryption. This is not recommended.",
```
jstourac, line 206: This warning wasn't there before this change? :thinking: What particular change caused this to pop up now? I can see that we set 0.0.0.0 before, which should be quite same as what we do now :thinking:  or am I wrong?

rstudio/c9s-python-3.11/setup_rstudio.py
```diff
@@ -60,7 +84,7 @@ def _get_cmd(port):
         '--server-working-dir=' + os.getenv('HOME'),
         '--auth-none=1',
         '--www-frame-origin=same',
-        #'--www-address=0.0.0.0',
+        '--www-address='+ detect_env(),
```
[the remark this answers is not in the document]
jstourac: Okay, probably this change caused this to pop up?
```
    - ServerApp.ip=0.0.0.0 \
    + ServerApp.ip="" \
```

tests/containers/workbenches/workbench_image_test.py
jstourac: Okay, probably this change caused this to pop up?
```
    - ServerApp.ip=0.0.0.0 \
    + ServerApp.ip="" \
```

jiridanek: yes, that one. 0.0.0.0 means all ipv4 interfaces, "" means all interfaces (ipv4 and 6) and causes the warning to pop up

rstudio/c9s-python-3.11/setup_rstudio.py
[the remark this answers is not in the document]
jiridanek: we should adopt a pre-commit and a python formatter; chasing formatting issues by hand is not very productive

I usually run the intelliJ reformat when I think of it, so the format is kept more or less in decent shape

tests/containers/workbenches/workbench_image_test.py
jstourac: It's quite strange though... I see no difference WRT security between the "0.0.0.0" and ""...

rstudio/c9s-python-3.11/setup_rstudio.py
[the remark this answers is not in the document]
jstourac: good first issue for an intern maybe? :slightly_smiling_face:

tests/containers/workbenches/workbench_image_test.py
jiridanek: i do see logic of the message. Listening on 0.0.0.0 is not listening on "all IP addresses", only on ipv4 addresses, so the message does not appear then. Did not investigate further, and i think it does not deserve more investigation.

rstudio/c9s-python-3.11/setup_rstudio.py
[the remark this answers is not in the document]
jiridanek: maybe, but not @RomanFilip, he needs something in Go, and I already have the thing in mind

tests/containers/workbenches/workbench_image_test.py
jstourac: it can be "all IP addresses" on IPv4 only stack; anyway, I believe this messages should be a security warning thus my point above; but yeah, no need to investigate further, that wasn't my intention.

jiridanek: maybe @jesuino might want to take a look, if the conditions when the message is printed may need a bit of tuning...

jiridanek: here's the code that produces the warning, I had to struggle a bit to find it

https://github.com/search?q=repo%3Ajupyter-server%2Fjupyter_server%20%22not%20using%20encryption%22&type=code

jiridanek: Looking at
```
if not app.ip:
```

if we provided :: as the IP, the warning may go away...

jstourac: yeah, I was wondering why we don't have "::" there as it is used for the other image

anyway, I don't mind that much, just wanted to clarify this... we don't need to bother with this anyhow more, I suppose

jiridanek:
/approve
/override ci/prow/codeserver-notebook-e2e-tests ci/prow/notebooks-ubi9-e2e-tests ci/prow/rstudio-notebook-e2e-tests

jiridanek:
> error: build error: Failed to push image: trying to reuse blob sha256:facf1e7dd3e0c59d3a9c051e50a2263491fd2cbfb31a8c9f6b188bf4af6d85cf at destination: unable to retrieve auth token: invalid username/password: authentication required


oh yeah!

jiridanek:

PR #920, opened 2025-02-24
[Kubeflow-Training Action] Update notebook's pipfile to sync with Kubeflow-Training SDK release 1.9.0

abhijeet-dhumal:
https://issues.redhat.com/browse/RHOAIENG-19845

:rocket: This is an automated Pull Request generated by [odh-kfto-sdk-notebooks-sync.yml](https://github.com/opendatahub-io/training-operator/tree/dev/.github/workflows/odh-kfto-sdk-notebooks-sync.yml) workflow.

This PR updates the `Pipfile` to sync with latest Kubeflow-Training SDK release.


andyatmiami:
Splatting some output here from running some of my own local verification scripts:

`codeserver`
```
Baseline image size: 2183542739

Modified image size: 2183542737
```

`datascience`

```
Baseline image size: 3228213715

Modified image size: 3228220369
```

`pytorch (cuda)`

```
Baseline image size: 11329757652

Modified image size: 11329750996
```

`pytorch (rocm)`

```
Baseline image size: 22809433042

Modified image size: 22809439701
```

`trustyai`

```
Baseline image size: 11325221842

Modified image size: 11325229012
```

✅ There are no differences in image size after applying KFTO 1.9.0

andyatmiami:
As a general comment / for future discussion... (see: [slack thread](https://redhat-internal.slack.com/archives/C060A5FJEAD/p1740427613706729?thread_ts=1740423713.378569&cid=C060A5FJEAD))

This block from pipelines team "makes me think":
- https://github.com/opendatahub-io/training-operator/blob/add-kfto-sdk-notebooks-sync/.github/workflows/odh-kfto-sdk-notebooks-sync.yaml#L81-L84

This is an adaptation of a similar line from codeflare-sdk team:
- https://github.com/project-codeflare/codeflare-sdk/blob/main/.github/workflows/odh-notebooks-sync.yml#L74-L78

In both of these logics... we are passing `--pre` flag to the `pipenv lock` command.. which is **NOT** a flag the IDE team passes when regenerating the lock file:
- https://github.com/opendatahub-io/notebooks/blob/main/.github/workflows/piplock-renewal.yaml#L69
- https://github.com/opendatahub-io/notebooks/blob/main/Makefile#L457

This leads to differences getting introduced.. only to then be "un-introduced" when our schedule piplock renewal action fires in the `notebooks/` repo... this "whiplash" in the commit history seems unnecessary and we should probably align all this at some point in the future to avoid commits like this getting "undone":
image

as a general rule of thumb (imho) - it would seem for sake of stability/etc - we should err on the side of NOT pulling in `--pre` dependencies (unless there is a specific/documented reason)

harshad16 (commented):
@abhijeet-dhumal , is the update of requirements.txt being made from https://github.com/opendatahub-io/training-operator/pull/31/files ?
maybe i m missing the line where this changes is being made.

also i m not sure, we want outside workflow to update the requirements.txt
maybe it would be best for IDE team to update that in-house based on changes in Pipfile. 
WDYT @jiridanek ? this way, we would be able to keep it stable.

abhijeet-dhumal:
> @abhijeet-dhumal , is the update of requirements.txt being made from https://github.com/opendatahub-io/training-operator/pull/31/files ? maybe i m missing the line where this changes is being made.
> 
> also i m not sure, we want outside workflow to update the requirements.txt maybe it would be best for IDE team to update that in-house based on changes in Pipfile. WDYT @jiridanek ? this way, we would be able to keep it stable.

@harshad16   If I understand correctly, The shell script generate_code.sh run is responsible for updating requirements.txt file 👀 
https://github.com/opendatahub-io/training-operator/pull/31/files#diff-43b9eca4300953f4eef18a388f4aaace196f07559af6f0d11ead836f6f81b074R143
cc: @jiridanek

harshad16:
> If I understand correctly, The shell script generate_code.sh run is responsible for updating requirements.txt file 👀 https://github.com/opendatahub-io/training-operator/pull/31/files#diff-43b9eca4300953f4eef18a388f4aaace196f07559af6f0d11ead836f6f81b074R143 

ah , right, totally missed it.
It makes sense now on why it was added, thanks for pointing me to this.

can you take a look at the `--pre` flag ? if it not intended, can you remove and recompile this file ?

abhijeet-dhumal:
> > If I understand correctly, The shell script generate_code.sh run is responsible for updating requirements.txt file 👀 https://github.com/opendatahub-io/training-operator/pull/31/files#diff-43b9eca4300953f4eef18a388f4aaace196f07559af6f0d11ead836f6f81b074R143
> 
> ah , right, totally missed it. It makes sense now on why it was added, thanks for pointing me to this.
> 
> can you take a look at the `--pre` flag ? if it not intended, can you remove and recompile this file ?

@harshad16 hey, If I try running workflow without --pre flag, it raises below issue, have you seen it before? : 
```

pipenv.exceptions.ResolutionFailure: ERROR: No matching distribution found for 
torch~=2.4.0
✘ Locking Failed!
Your dependencies could not be resolved. You likely have a mismatch in your 
sub-dependencies.
You can use $ pipenv run pip install <requirement_name> to bypass this mechanism, then 
run $ pipenv graph to inspect the versions actually installed in the virtualenv.
Hint: try $ pipenv lock --pre if it is a pre-release dependency.
ERROR: Failed to lock Pipfile.lock!
Failed to lock dependencies
```

jiridanek:
Run pipenv with --verbose. Change it in the workflow also. You want --verbose output.

jiridanek:
Here's how the locking goes if I do

`gh pr checkout 920`, push only the Pipfile changes without locking, and let our gha do locking with `make refresh-pipfilelock-files`

everything seems to lock fine

https://github.com/opendatahub-io/notebooks/pull/921

abhijeet-dhumal:
Using clened commit history, I updated version in only pipfiles
and ran make command as suggested here : https://github.com/opendatahub-io/notebooks/pull/920#issuecomment-2681736286
but... : https://privatebin.corp.redhat.com/?c8a16431184d1426#CQTxkApiWRNBVL2qYpCu9uH3mz5jXEMQCmC3LH54kvtb 
@jiridanek pointed out that the issue is with macos which I'm using, I will test it again tomorrow 👍 
In the meantime can someone please raise PR to update kfto-sdk version manually ?

jiridanek:
> In the meantime can someone please raise PR to update kfto-sdk version manually ?
https://github.com/opendatahub-io/notebooks/pull/921 should be just that

jstourac:
FTR https://issues.redhat.com/browse/RHOAIENG-19845

Also, since the #921 has been merged, should this be closed as obsolete now? @abhijeet-dhumal

jiridanek:
This sure looks obsolete to me.

We need to figure out the piplocking action for our fellow contributors. Looks like when they write it themselves, they either do it wrong, or they do it so that it produces different results than what ours does (which is also wrong, for some measure of wrongness).

PR #924, opened 2025-02-27
feat(containers): Remove reliance on "chained builds" architecture for image builds

andyatmiami:
Related openshift/release PR
* https://github.com/openshift/release/pull/62160

## Description

Commit contains following changes:

- Refactors all `Dockerfile`s to to be "self-contained" - not rely on any "build chains" for internal images
    - please note in the final form of this PR - many intermediate stages will be combined... this current form makes it easier to understand how the `Dockerfile`s present in the "build chain" are being combined - long term - we'd expect the following stages: `base`, `base-<accelerator>` (if applicable), `<final>`
- `Dockerfile` file now has a file extension/suffix that indicates CPU or accelerator
- `LABEL` directives now should be consistent/accurate
    - this probably needs a little more attention
- `base/` directory removed as its no longer relevant/required
- `rocm/` directory removed as its no longer relevant/required
- `wheel` + `setuptools` explicitly added to `runtime-` images `Pipfile`
- Makefile updated to remove pre-reqs on container image targets
- Makefile updated to handle support better across releases
    - _read: MOWR variables_
- `pytorch` `Makefile` targets now contain `cuda-` prefix
- `rocm` + `cuda` `datascience` `Makefile` targets removed as we don't ship those variants
    - with no chained builds - all datascience-dependent images contain their own required `Dockerfile` logic!
- `ENV` directive in `Dockerfile` now properly uses `=` (vs. whitespace)
- change in `buildinputs` to pull file paths from the terminal layer
    - this change particularly needs refined - but its "crude yet effective" in its current form
- scripts related to automation updates to account for these new changes
    - `gen_gha_matrix_jobs.py` now uses an `all-images` `Makefile` target to identify image-related targets (while also supported the `Make` variables that appear in our `target` names now)
    - `gha_pr_changed_files.py` adapted to: 
        - account for the fact we only ever discover a single "build directory" based on the `#\(MACHINE-PARSED LINE\)#\` in `make` output
        - support multiple `Dockerfile` (with differing extensions) existing in a single directory
            - ℹ️ Code is already/pre-emptively ignoring presence of `Dockerfile.konflux`
            - ⚠️ Inefficiency introduced in `filter_out_unchanged` that I would prefer to treat in a subsequent PR
    - `test_make_building_only_specified_images` function removed from `make_test.py` as its irrelevant now when not using chained builds
    - `has_tests.py` updated to:
        - account for the fact a given Makefile target will no longer ever include mutiple `#\(MACHINE-PARSED LINE\)#\` in its output
            - i.e. there is only a single directory involved for a given `Makefile` `target` when dealing with `gha_pr_changed_files`   
         - due to how the `Makefile` is parsed now - we need to make sure we are passing a valid `Makefile` `target` into the test case 

Related-to: https://issues.redhat.com/browse/RHOAIENG-19048

## How Has This Been Tested?

- ✅ I have written [a utility I have written to ensure the image being built is comparable to the existing image we have](https://gitlab.cee.redhat.com/astonebe/notebook-utils/-/tree/main/compare-images?ref_type=heads)
    - With the provided "ignores" configuration - I have gotten vetted/successful builds on all supported images from `main` branch
    - ⚠️ To ease some difficulties in comparing across branches - I did make a few (imho) insignificant changes to the `main` branch as seen here:
        - https://github.com/opendatahub-io/notebooks/compare/main...andyatmiami:notebooks:dockerfile-refactor-baseline
- ✅ As these changes are a pure refactor / not intending to introduce any actual functional differences - the tests ran as part of GHA and Prow checks provide decent validation of the changes

## Merge criteria:


codeserver/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -1,7 +1,35 @@
-ARG BASE_IMAGE
-FROM ${BASE_IMAGE}
+####################
+# base             #
+####################
+FROM registry.access.redhat.com/ubi9/python-311:latest AS base
```
jstourac, line 4: Since we're moving to konflux soon anyway, what do you think about moving from `latest` to an actual sha/tag for the base images? This may help also #901 maybe?

@jiridanek WDYT?

codeserver/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -105,19 +118,30 @@ RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf.sed ${NGINX_CONF_PATH} && \
...
+RUN echo "Installing softwares and packages" && \
+    micropipenv install && \
+    rm -f ./Pipfile.lock && \
+    # Fix permissions to support pip in Openshift environments \
+    chmod -R g+w /opt/app-root/lib/python3.11/site-packages && \
+    fix-permissions /opt/app-root -P
+
+WORKDIR /opt/app-root/src
+
 CMD ["/opt/app-root/bin/run-code-server.sh"]
```
jstourac, line 147: Eventually I think we should unify to use ENTRYPOINT for all our images in the future. Or do we have some reason to use ENTRYPOINT somewhere and CMD command elsewhere that I'm missing? :thinking: 

https://www.docker.com/blog/docker-best-practices-choosing-between-run-cmd-and-entrypoint/

Note: again, not something to deal with in this PR.

atheo89:
/testwith opendatahub-io/notebooks/main/notebook-jupyter-ubi9-python-3-11-pr-image-mirror openshift/release#62160

atheo89:
/testwith opendatahub-io/notebooks/main/jupyter-ds-ubi9-python-3-11-pr-image-mirror openshift/release#62160

jstourac:
@andyatmiami thank you for this. I scanned through these changes; LGTM in general. Didn't do the extended thorough check though. Just to be sure - you checked the image differences by your script for all the images we are touching here?

Regarding the testing - apart from manual effort, I think that the first step should be to make the pytest tests pass here in this PR for these changes so we utilize at least that minimal automated checks in this repository. I can take a look on some of the failing CI checks, just feel free to let me know so we aren't duplicating our work.

I put some comments, but these are more just an ideas for the future or discussion and shouldn't be addressed in this PR anyway.

andyatmiami:
@jstourac - appreciate the review...

w.r.t the testing I did do.. it was using the (admittedly initially diificult to work with) `diffoci` tool...

so all these images are of "comparable" (in most cases **identical**) in image size...

The few places the image size is different - its due to minor corrections made in the PR (example: explicitly adding `setuptools` and `wheel` dependencies in `runtimes/` images - which causes **very minor** changes in image size)

codeserver/ubi9-python-3.11/Dockerfile.cpu
jiridanek: absolutely not in this PR

I'm guessing if we let renovate to run, it will rewrite the tags first thing ;p so maybe no manual effort needed

good thought, tho

codeserver/ubi9-python-3.11/Dockerfile.cpu
jiridanek: It's that somewhere we're using the s2i entrypoint and somewhere we decided not to... We should stop using s2i also ;P

andyatmiami:
/testwith opendatahub-io/notebooks/main/notebook-jupyter-ubi9-python-3-11-pr-image-mirror `openshift/release#62160`

jiridanek:
@andyatmiami /testwith does not work, if you read the thread that andriana posted, that's what we concluded

jiridanek:
we have to merge andrianas pr first, and then it's possible to rerun the ocp-ci checks and they should start running

codeserver/ubi9-python-3.11/Dockerfile.cpu
jiridanek: * https://github.com/opendatahub-io/notebooks/issues/928

jiridanek:
Great job on the last minute fixes!

The openshift-ci -e2e- tests look like they are failing on some configuration error. This would be best fixed in subsequent pr to openshift/release. If the pr here is merged, then the openshift/release pr can be rehearsed.

I'd merge this as is.

/lgtm

atheo89:
Thank you Andy for this work! Really good upgrade on the notebooks! I would like to merge this as soon as possible as the full work is done, this way we can resolve potential bugs/findings next week while the package upgrades. 

/lgtm
/approve

jiridanek:
/override ci/prow/
rocm-runtimes-ubi9-e2e-tests

jiridanek:
/override ci/prow/
codeserver-notebook-e2e-tests

PR #923, opened 2025-02-27
RHOAIENG-20088: GHA CI enhancement: check software manifests against images

jstourac:
This adds a new check for the consistency of the information in the
ImageStream manifest files. We now check that the software versions info
in these files truly matches what is installed in the referenced images.

More information is tracked in:

* https://issues.redhat.com/browse/RHOAIENG-20569

---

The first commit in this PR is irrelevant to the software version check and is fixing the rules when the params-env check is being triggered by extending it by the change in `ci/check-runtime-images.sh` file.

## How Has This Been Tested?
```
./ci/check-software-versions.py
```

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #925, opened 2025-02-27
RHOAIENG-20569: Fix the ImageStream manifests with the correct software versions

jstourac:
This was caught by the demo PR implementation check in #923 for the 
https://issues.redhat.com/browse/RHOAIENG-20569.

Please check thoroughly - maybe I misunderstood what we want to show e.g. for the `rstudio-server`. Also, what is suspicious is the R studio version of the N-1 images and Cuda version in R studio version N.

## Merge criteria:

### DIFF
--- a/manifests/base/rstudio-gpu-notebook-imagestream.yaml
+++ b/manifests/base/rstudio-gpu-notebook-imagestream.yaml
@@ -20,14 +20,14 @@ spec:
         # language=json
         opendatahub.io/notebook-software: |
           [
-            {"name": "CUDA", "version": "12.4"},
+            {"name": "CUDA", "version": "12.1"},
             {"name": "R", "version": "v4.4"},
             {"name": "Python", "version": "v3.11"}
           ]
         # language=json
         opendatahub.io/notebook-python-dependencies: |
           [
-            {"name": "rstudio-server", "version": "4.4"}
+            {"name": "rstudio-server", "version": "2024.04"}
           ]
         openshift.io/imported-from: quay.io/opendatahub/workbench-images
         opendatahub.io/workbench-image-recommended: 'true'
@@ -38,19 +38,19 @@ spec:
       name: "2024.2"
       referencePolicy:
         type: Source
-    # N Version of the image
+    # N-1 Version of the image
     - annotations:
         # language=json
         opendatahub.io/notebook-software: |
           [
             {"name": "CUDA", "version": "12.1"},
-            {"name": "R", "version": "v4.3"},
+            {"name": "R", "version": "v4.4"},
             {"name": "Python", "version": "v3.9"}
           ]
         # language=json
         opendatahub.io/notebook-python-dependencies: |
           [
-            {"name": "rstudio-server", "version": "4.3"}
+            {"name": "rstudio-server", "version": "2023.12"}
           ]
         openshift.io/imported-from: quay.io/opendatahub/workbench-images
         opendatahub.io/workbench-image-recommended: 'false'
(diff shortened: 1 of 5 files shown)


manifests/base/rstudio-gpu-notebook-imagestream.yaml
```diff
@@ -38,19 +38,19 @@ spec:
...
         type: Source
-    # N Version of the image
+    # N-1 Version of the image
     - annotations:
         # language=json
         opendatahub.io/notebook-software: |
           [
             {"name": "CUDA", "version": "12.1"},
-            {"name": "R", "version": "v4.3"},
+            {"name": "R", "version": "v4.4"},
```
jstourac, line 47: Am I doing the R version check wrong somehow? :thinking: 

https://github.com/opendatahub-io/notebooks/blob/2024a/rstudio/c9s-python-3.9/Dockerfile#L34-L35

manifests/base/rstudio-gpu-notebook-imagestream.yaml
```diff
@@ -20,14 +20,14 @@ spec:
         # language=json
         opendatahub.io/notebook-software: |
           [
-            {"name": "CUDA", "version": "12.4"},
+            {"name": "CUDA", "version": "12.1"},
```
jstourac, line 23: this is also weird

jiridanek: yeah, I'm sure @andyatmiami would have something to say about cuda on rstudio images

manifests/base/rstudio-gpu-notebook-imagestream.yaml
jiridanek: I think not, it really is 4.4, and we just did not update the paths correctly in Dockerfiel

manifests/base/rstudio-gpu-notebook-imagestream.yaml
andyatmiami: For reasons that predate me - CUDA is very much on 12.1 for RStudio (which is inconsistent as other images do in fact use CUDA 12.4

hopefully we can resolve this discrepancy in 2025a.. but the change as proposed does accurately then reflect reality.

manifests/base/rstudio-gpu-notebook-imagestream.yaml
jstourac: Okay, I guess I know the reason now... despite we try to configure [R_VERSION](https://github.com/opendatahub-io/notebooks/blob/73f1c9172ad42db89452cad2f1717e48931e9012/rstudio/c9s-python-3.11/Dockerfile.cpu#L49) variable with a specific version of R language, we simply install the R from the [EPEL](https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/Packages/r/) repository as an rpm ([see here](https://github.com/opendatahub-io/notebooks/blob/73f1c9172ad42db89452cad2f1717e48931e9012/rstudio/c9s-python-3.11/Dockerfile.cpu#L55-L57)). At the time I'm writing this, the R version in that repository is 4.4.2.

This is the reason why the version of the rstudio-server matches our expectation but the version of the R language pack below it doesn't match. So both our images we provide (N and N-1) actually point to the latest R version in that EPEL repository.

We should count with it once we'll move R Studio in RHOAI from TP or we should fix it.

jstourac:
Thank you all for your reviews. Looks like all has been explained and we should keep an eye on [R version](https://github.com/opendatahub-io/notebooks/pull/925/files#r1975161279) in our R Studio images.

/override ci/prow/images
/approve

PR #930, opened 2025-02-28
Include runtimes ImageStreams on manifests + gha automation

atheo89:
Related to: https://issues.redhat.com/browse/RHOAIENG-19050

## Description

This PR includes the following changes:

 - Introduces six newly created ImageStreams to utilize runtime images via manifests.
The .json with its metadata incorporated under the imagestream using the `opendatahub.io/runtime-image-metadata` annotation.

 - And a new GHA that updates these files upon user's request. 

**NOTES**:
1. The gha named (temporary) as runtimes-digest-updater-upstream-v2.yaml.  This (-v2) is there in order don't break the functionality of the current runtimes-digest-updater-upstream.yaml until fully switch to these un-wiring strategy. A follow up PR will be open to update this.
 2. I didn't use .env and its correspondence vars on the imagestreams because it does not function as expected with the multi-line json. So the parameter inside the json does not updated updated properly this way. IIf we want it this way it increase a lot the complexity on the manifests files as it needs to add these jsons as configmaps an patch them. Personally, I wouldn't prefer that.

## How Has This Been Tested?

Use as `DevFlag`  on RHOAI DSC the following in order to inspect the ImageStreams.

```
     workbenches:
       devFlags:
         manifests:
           - contextDir: manifests
             sourcePath: ''
             uri: 'https://github.com/atheo89/notebooks/archive/runtimes-imagestreams.tar.gz'
       managementState: Managed
```

2.  GHA updater run: https://github.com/atheo89/notebooks/actions/runs/13571562492, and the generated PR: https://github.com/atheo89/notebooks/pull/23

## Merge criteria:

### DIFF
--- a/.github/workflows/runtimes-digest-updater-upstream-v2.yaml
+++ b/.github/workflows/runtimes-digest-updater-upstream-v2.yaml
@@ -1,0 +1,102 @@
+---
+# The aim of this GitHub workflow is to update the runtimes ImageStreams
+name: Update runtime ImageStreams SHA digests
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
+    inputs:
+      branch:
+        required: true
+        description: "Which branch do you want to update?"
+      tag_version:
+        required: true
+        description: "Provide tag version: main or YYYYx"
+      user-hash:
+        required: false
+        description: "Optional: Specify a Git hash (it should exist in the provided branch's history)"
+
+env:
+  TMP_BRANCH: tmp-branch-${{ github.run_id }}
+  BRANCH_NAME: ${{ github.event.inputs.branch || 'main' }}
+  TAG_VERSION: ${{ github.event.inputs.tag_version || 'main' }}
+  USER_HASH: ${{ github.event.inputs.user-hash }}
+
+jobs:
+  initialize:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Install Skopeo CLI
+        shell: bash
+        run: |
+          sudo apt-get -y update
+          sudo apt-get -y install skopeo yq jq
+
+      - name: Checkout branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.BRANCH_NAME }}
+
+      - name: Create a new branch
+        run: |
+         echo ${{ env.TMP_BRANCH }}
+         git checkout -b ${{ env.TMP_BRANCH }}
+         git push --set-upstream origin ${{ env.TMP_BRANCH }}
+
+  update-runtimes:
+    needs: [initialize]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Configure Git
+        run: |
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "GitHub Actions"
+
+      - name: Checkout release branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.TMP_BRANCH }}
+          fetch-depth: 0
+
+      - name: Invoke script to handle the update
+        shell: bash
+        run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/runtimes-digest-updater-upstream-v2.yaml
```diff
@@ -0,0 +1,102 @@
+---
+# The aim of this GitHub workflow is to update the runtimes ImageStreams
```
jiridanek, line 2: why is separate gha for updating runtimes still needed? Can't we now update all imagestreams at once, at the end of build?

.github/workflows/runtimes-digest-updater-upstream-v2.yaml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Checkout release branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.TMP_BRANCH }}
+          fetch-depth: 0
+
+      - name: Invoke script to handle the update
+        shell: bash
+        run: |
+          chmod +x "${GITHUB_WORKSPACE}/ci/runtimes-digest-updater.sh"
```
jstourac, line 66: Do we need this step since this script is invoked via `bash ...` on the next line? And if we do, shouldn't we simply add these permissions to this script directly in our GH repo so we don't need to do it here everytime?

manifests/base/runtime-rocm-pytorch-imagestream.yaml
```diff
@@ -0,0 +1,39 @@
+---
+apiVersion: image.openshift.io/v1
+kind: ImageStream
+metadata:
+  labels:
+    opendatahub.io/runtime-image: "true"
+  annotations:
+    opendatahub.io/runtime-image-url: "https://github.com//opendatahub-io/notebooks/tree/main/runtimes"
+    opendatahub.io/runtime-image-name: "Pytorch with ROCm and Python 3.11 (UBI9)"
```
jstourac, line 9: Maybe?
```suggestion
    opendatahub.io/runtime-image-name: "Py?orch with ROCm and Python 3.11 (UBI9)"
```

jstourac:
Do we have a plan what to do with https://github.com/opendatahub-io/notebooks/blob/main/ci/check-runtime-images.sh and relevant GHA? Do we want to update to conform these changes? Personally I would keep it around in case of some manual changes to these references... we can talk about that of course. We can amend, we can improve, we can discuss :grin:

.github/workflows/runtimes-digest-updater-upstream-v2.yaml
atheo89: Good point—I thought the same. However, until we complete this work: [RHOAIENG-19049](https://issues.redhat.com/browse/RHOAIENG-19049), we should keep it separate because we'll need to use the old procedure. Once that's done, we can easily update both notebooks and runtimes together.

manifests/base/runtime-rocm-pytorch-imagestream.yaml
atheo89: I replaced `Pytorch` with `PyTorch`; I don’t think that’s `?` what you meant :P

jiridanek:
/lgtm
too

atheo89:
Thank you for your review folks, I will move this in.
/approve

atheo89:
/override ci/prow/images
As this PR affect any image

jstourac:
btw no comment on [this one](https://github.com/opendatahub-io/notebooks/pull/930#issuecomment-2694370165)?

atheo89:
>btw no comment on https://github.com/opendatahub-io/notebooks/pull/930#issuecomment-2694370165?

 Sorry @jstourac i miss that, I would say too, to keep the `check-runtime-images.sh` as it works with the old embedded .json files structure, and write a new bash script v2 that will work with the new changes.

PR #936, opened 2025-03-04
RHOAIENG-16384: Fix build error of Rcpp for R 4.4.2

jstourac:
More information in [1].

* [1] https://issues.redhat.com/browse/RHOAIENG-16384

## Merge criteria:

### DIFF
--- a/rstudio/c9s-python-3.11/Dockerfile.cpu
+++ b/rstudio/c9s-python-3.11/Dockerfile.cpu
@@ -21,7 +21,7 @@ USER 1001
 RUN curl -L https://mirror.openshift.com/pub/openshift-v4/$(uname -m)/clients/ocp/stable/openshift-client-linux.tar.gz \
         -o /tmp/openshift-client-linux.tar.gz && \
     tar -xzvf /tmp/openshift-client-linux.tar.gz oc && \
-    rm -f /tmp/openshift-client-linux.tar.gz 
+    rm -f /tmp/openshift-client-linux.tar.gz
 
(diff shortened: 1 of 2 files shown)


jiridanek:
/lgtm

I watched Jan do the work, so I feel good about the choices made every step of the way.

jiridanek:
Is a follow-up pr against rhds/notebooks rhel-based eStudio images required? I do think it is.

jstourac:
That is a good point, Jiri. I think that this one file will be updated by changes from upstream:
* https://github.com/red-hat-data-services/notebooks/blob/main/rstudio/c9s-python-3.11/Dockerfile
but we should check.

The other two files in the downstream seem they don't install the R packages as we do in ODH for some reason (no idea):
* https://github.com/red-hat-data-services/notebooks/blob/b23e7ed3466f54a1b045f9b34f924b7a725a062d/rstudio/rhel9-python-3.11/Dockerfile#L76
* https://github.com/red-hat-data-services/notebooks/blob/b23e7ed3466f54a1b045f9b34f924b7a725a062d/rstudio/rhel9-python-3.9/Dockerfile#L74 (but it's Python 3.9 anyway so we don't care about this one).

jstourac:
Thank you for your reviews, guys!

/approve

jiridanek:
Here's a PR that also got stuck on konflux, and ppl just merged it in github with button, looks like ;(

* https://github.com/openshift/hypershift/pull/4284#issuecomment-2339782755

PR #937, opened 2025-03-05
RHOAIENG-19908: Install tf2onnx from dev-packages for overriding protobuf

harshad16:
Install tf2onnx from dev-packages to avoid conflict with other packages (kfp, codeflare-sdk) due to the protobuf pinning in tf2onnx.

## Description

Pipenv doesn't have the option to override transitive dependency. 
Use the dev-packages, we can override the protobuf pinning for tf2onnx
https://github.com/pypa/pipenv/issues/1921#issuecomment-379131833

Related-to: https://issues.redhat.com/browse/RHOAIENG-19908

With tf2onnx conflict moved to dev-packages; 
kfp, codeflare-sdk and kubeflow-training can be upgraded.

- Removed the patch explicitly applied for elyra.
Related-to: https://issues.redhat.com/browse/RHOAIENG-6486

## How Has This Been Tested?

Tested manually:
1. TF2onnx:
Executed the sample example from here:
https://onnxruntime.ai/docs/tutorials/tf-get-started.html

2. In elyra, the add_annotation, add_label and add_toleration were removed:
with: 
- Setup jupyternotebook with tf image of PR.
- Setup elyra example : used iris example: https://github.com/harshad16/data-science-pipeline-example
- set some annotations and label.
- Execute or try export function to test.

## Merge criteria:


jupyter/tensorflow/ubi9-python-3.11/Pipfile
```diff
@@ -23,7 +23,8 @@ scikit-learn = "~=1.5.1"
 scipy = "~=1.14.1"
 skl2onnx = "~=1.17.0"
 onnxconverter-common = "~=1.13.0" # Required for skl2onnx, as upgraded version is not compatible with protobuf
```
jiridanek, line 25: shouldn't this be removed/upgraded?

jupyter/rocm/tensorflow/ubi9-python-3.11/Pipfile
```diff
@@ -4,17 +4,17 @@ verify_ssl = true
 name = "pypi"
 
 [dev-packages]
+tf2onnx = "~= 1.16.1"
```
jiridanek, line 7: I'd like to see a comment here, why tf2onnx is in dev-packages

jupyter/tensorflow/ubi9-python-3.11/Pipfile
harshad16: The onnxconverter-common can't be upgraded as it has also pinned protobuf in its latest version 
https://github.com/microsoft/onnxconverter-common/blob/8d2f85f0a7039a0b0defd6c7509cb905914a4d2f/pyproject.toml#L27
if we do want to fix this, it needs to be fixed in all images
so do think it fits to this PR.

jupyter/rocm/tensorflow/ubi9-python-3.11/Pipfile
harshad16: Added details now.
let me know if that fits.

jiridanek: thanks, looks informative enough to me!

jupyter/tensorflow/ubi9-python-3.11/Pipfile
jiridanek: if it's still relevant, then I have no objections

jstourac:
LGTM in general. Regarding the failed CI jobs - looks like there is some issue for the utils directory. I didn't check the sandbox.py script fully, so I'm not sure whether this is the rootcause, but since you are removing the `utils` directories in this PR for the two dockerfiles, shouldn't we also remove it's reference for COPY commands here:
* https://github.com/harshad16/odh-notebooks/blob/fix-tf2onnx/jupyter/tensorflow/ubi9-python-3.11/Dockerfile.cuda#L218
* https://github.com/harshad16/odh-notebooks/blob/fix-tf2onnx/jupyter/tensorflow/ubi9-python-3.11/Dockerfile.cuda#L155

jiridanek:
@jstourac good observation

```
FileNotFoundError: [Errno 2] No such file or directory: 'jupyter/tensorflow/ubi9-python-3.11/utils'
```

looks like something broken, but actually, it means that Dockerfile mentions something that's not in the filesystem

which is actually neat that it can validate COPY commands up front, before the build gets to them

jiridanek:
ocp-ci fails like this, but it takes 10+ mins to get to the Dockerfile line and fail there

```
error: build error: building at STEP "COPY ${TENSORFLOW_SOURCE_CODE}/utils ./utils/": checking on sources under "/tmp/build/inputs": copier: stat: "/jupyter/tensorflow/ubi9-python-3.11/utils": no such file or directory
```

andyatmiami (commented):
Its great that we are able to remove the `utils.python_dsl_template.patch` file and its references 💯 

....but ( 😇 ) just to close the loop - could your PR comments highlight how/when/in what PR this was addressed:
> issue is fixed with kfp-kubernetes upgrade

This is from the inline comments on the Dockerfile:

```
    # Temporary:Workaround for fixing the kfp-kubernetes 1.0.0 for elyra pipeline execution
    # TODO: Remove this patch once the issue is fixed with kfp-kubernetes upgrade. 
```

andyatmiami:
/lgtm

thanks for sorting out/unblocking our dependency-conflict hell here 😇

jiridanek:
/override ci/prow/notebooks-ubi9-e2e-tests ci/prow/rocm-notebooks-e2e-tests
cluster provisioning infra issues

harshad16:
> Its great that we are able to remove the `utils.python_dsl_template.patch` file and its references 💯
> 
> ....but ( 😇 ) just to close the loop - could your PR comments highlight how/when/in what PR this was addressed:
> 
> > issue is fixed with kfp-kubernetes upgrade
> 
> This is from the inline comments on the Dockerfile:
> 
> ```
>     # Temporary:Workaround for fixing the kfp-kubernetes 1.0.0 for elyra pipeline execution
>     # TODO: Remove this patch once the issue is fixed with kfp-kubernetes upgrade. 
> ```
> 
> **UPDATE: SORRY I CAN'T READ - THIS INFO ALREADY INCLUDED**

@andyatmiami good question, perhaps i can explain a little

As kfp version was set to 2.5 earlier, the function attributes like add_annotation, add_label and add_toleration 
didn't function. So we have explicitly removed that features from causing the problem, using patch
and so that rest of feature of elyra can be used.

With these changes applied, those function attributes can be used.
To test these, following action can be used:
1. Set the tf image based jupyter notebook
2. Set up a pipeline with these attributes included
3. Execute or export the result pipeline to see the inclusion.

Screenshot from 2025-03-07 15-11-32
Attaching screen shoots of before and after
Without fix: 
Screenshot from 2025-03-07 15-17-20
With new fix:
Screenshot from 2025-03-07 15-12-15

jupyter/tensorflow/ubi9-python-3.11/Pipfile
harshad16: yes, still relevant

harshad16:
Thanks for the review all 
/approve

Issue #951, opened 2025-03-15
Need a way to trigger GHA ci for PRs that don't change the code, but change the infra itself


jiridanek:
Ideas

* when changing infra code (Makefile, YAML gha workflows, helper scripts, ...) always rebuild everything in PR
* same when changing tests that are stored in tests/ directory but they may test multiple or all images
* leave it as it is, now all images are rebuilt on push to a PR branch in PR submitter's fork, if they enabled gha there
* allow triggering jobs on demand with a PR comment, something like is possible with openshift-ci
   * https://github.com/opendatahub-io/notebooks/pull/948#issuecomment-2726423948
* trigger some image builds always on every PR, either random, or pick the "interesting" ones that give broad coverage of our CI capabilities, e.g. codeserver is interesting one

jiridanek:
@coderabbitai analyze, got some new .github/workflows

PR #962, opened 2025-03-18
RHOAIENG-18400: chore(.tekton/): implement computation of `"pipelinesascode.tekton.dev/on-cel-expression"` values

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-18400

## Description

This is necessary so that our `-pull-request.yaml` pipelines are run every time build inputs are changed, but not when build inputs have not changed.

The next step is to let the script generate pipelines for all our images, and I don't want to do that carelessly, otherwise we'd be rebuilding every image on Konflux with every PR update.

We should also consider running this when the tests change, similarly as we don't yet do (but want to) with Github Actions

* https://github.com/opendatahub-io/notebooks/issues/951

## How Has This Been Tested?

* https://github.com/jiridanek/notebooks/actions/runs/13928049662

```
PYTHONPATH=. python3 ci/cached-builds/konflux_generate_component_build_pipelines.py
```

## Merge criteria:

### DIFF
--- a/.tekton/jupyter-minimal-ubi9-python-3-11-pull-request.yaml
+++ b/.tekton/jupyter-minimal-ubi9-python-3-11-pull-request.yaml
@@ -10,7 +10,11 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/cancel-in-progress: "true"
     pipelinesascode.tekton.dev/max-keep-runs: '3'
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main" && has(body.repository) && body.repository.full_name == "opendatahub-io/notebooks"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main" && ( "jupyter/minimal/ubi9-python-3.11/Pipfile.lock".pathChanged()
+      || "jupyter/minimal/ubi9-python-3.11/start-notebook.sh".pathChanged() || "jupyter/utils/***".pathChanged()
+      || ".tekton/jupyter-minimal-ubi9-python-3-11-pull-request.yaml".pathChanged()
+      || "jupyter/minimal/ubi9-python-3.11/Dockerfile.cpu".pathChanged() ) && has(body.repository)
+      && body.repository.full_name == "opendatahub-io/notebooks"
   creationTimestamp:
   labels:
     appstudio.openshift.io/application: notebooks
--- a/ci/cached-builds/konflux_generate_component_build_pipelines.py
+++ b/ci/cached-builds/konflux_generate_component_build_pipelines.py
@@ -2,10 +2,15 @@
 
 import re
 import pathlib
+
 import yaml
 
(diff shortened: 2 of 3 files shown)


.tekton/jupyter-minimal-ubi9-python-3-11-pull-request.yaml
```diff
@@ -10,7 +10,11 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/cancel-in-progress: "true"
     pipelinesascode.tekton.dev/max-keep-runs: '3'
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main" && has(body.repository) && body.repository.full_name == "opendatahub-io/notebooks"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main" && ( "jupyter/minimal/ubi9-python-3.11/Pipfile.lock".pathChanged()
+      || "jupyter/minimal/ubi9-python-3.11/start-notebook.sh".pathChanged() || "jupyter/utils/***".pathChanged()
+      || ".tekton/jupyter-minimal-ubi9-python-3-11-pull-request.yaml".pathChanged()
+      || "jupyter/minimal/ubi9-python-3.11/Dockerfile.cpu".pathChanged() ) && has(body.repository)
```
jstourac, line 16: Just wondering - why not simply use `jupyter/minimal/ubi9-python-3.11/***`? :thinking:

jstourac:
Once we generate the tekton files for the konflux pipeline and then sometime in the future, we'll add some file in the image directory definition - will we be notified we need to re-generate the tekton files so that this newly added dependency in the build tree is taken into account for the build triggering?

.tekton/jupyter-minimal-ubi9-python-3-11-pull-request.yaml
jiridanek: This way, it's possible to make change in .cpu Dockerfile without triggering a .cuda Dockerfile rebuild

You're right that quite possibly it's not worth it, given how long these yaml conditions become.

I'll change to your suggestion.

jiridanek:
> will we be notified we need to re-generate

The generating script has to be part of scripts/generate_code.sh, it how it's called

I intended to do this way.

Just now I'm thinking im making that step running longer now, and it will need to compile the bin/buildinputs now, which needs go

.tekton/jupyter-minimal-ubi9-python-3-11-pull-request.yaml
jstourac: Ah, you're right. I missed the cpu vs cuda thing. So, if we are sure we'll have these expressions always up-to-date (generate_code.sh as you mention below), then I'm fine with this.

Both way (the current and the simplistic ***) are okay for me right now when condition above is fullfiled.

jstourac:
Thank you for this. Giving LGTM and I'll leave decision about the expressions discussed above on you or on somebody else to say.

/lgtm

PR #981, opened 2025-03-21
NO-ISSUE: chore(ci): fix expected image names in params.env check

jstourac:
With recent 2025a image version introduced, we fixed some image build names annotations and these need to be reflected in this check now.

For older images, the fixes weren't applied as they weren't rebuild, so no fix is done for these N-1 image version checks.

## How Has This Been Tested?
```
./ci/check-params-env.sh
```

Note: there are still failures in this script due to the image size changes. These will be investigated first and then fixed with a followup PR eventually.

## Merge criteria:

### DIFF
--- a/ci/check-params-env.sh
+++ b/ci/check-params-env.sh
@@ -112,7 +112,7 @@ function check_image_variable_matches_name_and_commitref_and_size() {
             expected_img_size=503
             ;;
         odh-minimal-gpu-notebook-image-n)
-            expected_name="odh-notebook-jupyter-minimal-ubi9-python-3.11"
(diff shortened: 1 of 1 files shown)


jstourac:
Thank you for your reviews guys!

/approve

andyatmiami:
Thank you for this fix @jstourac !  I had fixed the annotations are part of my Dockerfile refactor - but just totally overlooked the fact they were being tested against now... apologies for my oversight!

andyatmiami:
/override "validation-of-params-env "

this PR only fixes the image build name annotations - and we can see from the failed job that nothing is failing the name check now.

other fixes will be investigated and dealt with in separate PR!

PR #992, opened 2025-03-27
RHOAIENG-21740: Perform filesystem permission check on images

jstourac:
This adds a simple test to scan our final images for the crucial directories for their ownership and permissions configuration.

## Description
https://issues.redhat.com/browse/RHOAIENG-21740

## How Has This Been Tested?
```
for image_ref in $(cat manifests/base/params.env | grep "\-n=" | cut -d "=" -f 2); do DOCKER_HOST=unix:///run/user/$UID/podman/podman.sock poetry run pytest tests/containers --image "${image_ref}" -k test_file_permissions; done
```

## Merge criteria:

### DIFF
--- a/tests/containers/base_image_test.py
+++ b/tests/containers/base_image_test.py
@@ -15,6 +15,7 @@ import testcontainers.core.container
 import testcontainers.core.waiting_utils
 
 from tests.containers import docker_utils
+from tests.containers import utils
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/containers/base_image_test.py
```diff
@@ -9,7 +9,7 @@
 import re
 import tempfile
 import textwrap
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any, Callable, List
```
jiridanek, line 12:
```suggestion
from typing import TYPE_CHECKING, Any, Callable
```

you can use `list[str]` as the type, no need to import the capital L list

this is a new feature beginning with some version of python, but we do have it available here

tests/containers/base_image_test.py
```diff
@@ -233,3 +257,18 @@ def encode_python_function_execution_command_interpreter(python: str, function:
         print({name}({parameters}));""")
     int_cmd = [python, "-c", program]
     return int_cmd
+
+import docker.errors
+import docker.models.images
+
+def is_rstudio_image(my_image: str) -> bool:
```
jiridanek, line 264: would be nice to avoid duplicating

https://github.com/opendatahub-io/notebooks/blob/02e6ef3aa8a862dd7206d59a7250f21c83af2578/tests/containers/conftest.py#L92-L99

but I don't insist on it here

jiridanek:
```
for image_ref in $(cat manifests/base/params.env | grep "\-n=" | cut -d "=" -f 2); do DOCKER_HOST=unix:///run/user/$UID/podman/podman.sock poetry run pytest tests/containers --image "${image_ref}" -k test_file_permissions; done
```

this seems useful, also for checking releases, maybe there should be a script file with the command

tests/containers/base_image_test.py
jstourac: you're right, I'll update

tests/containers/base_image_test.py
jstourac: adjusted

jstourac:
@jiridanek thank you for your review, I just updated based on your comments!

PR #996, opened 2025-03-28
Kustomize2warnings

jstourac:
Get rid of the warnings we get during our kustomize build so we are more ready for an eventual new kustomize version bump.

## How Has This Been Tested?
First I run on main:
```
./ci/kustomize.sh
```
save the stdout file as a reference
Then I checkout this branch, run the kustomize.sh again and compare the result stdout file again - these must be identical.

## Merge criteria:

### DIFF
--- a/manifests/base/kustomization.yaml
+++ b/manifests/base/kustomization.yaml
@@ -1,660 +1,661 @@
+---
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- jupyter-minimal-notebook-imagestream.yaml
-- jupyter-datascience-notebook-imagestream.yaml
-- jupyter-minimal-gpu-notebook-imagestream.yaml
-- jupyter-pytorch-notebook-imagestream.yaml
-- jupyter-tensorflow-notebook-imagestream.yaml
-- jupyter-trustyai-notebook-imagestream.yaml
-- code-server-notebook-imagestream.yaml
-- rstudio-notebook-imagestream.yaml
-- rstudio-gpu-notebook-imagestream.yaml
-- jupyter-rocm-minimal-notebook-imagestream.yaml
-- jupyter-rocm-pytorch-notebook-imagestream.yaml
-- jupyter-rocm-tensorflow-notebook-imagestream.yaml
+  - jupyter-minimal-notebook-imagestream.yaml
+  - jupyter-datascience-notebook-imagestream.yaml
+  - jupyter-minimal-gpu-notebook-imagestream.yaml
+  - jupyter-pytorch-notebook-imagestream.yaml
+  - jupyter-tensorflow-notebook-imagestream.yaml
+  - jupyter-trustyai-notebook-imagestream.yaml
+  - code-server-notebook-imagestream.yaml
+  - rstudio-notebook-imagestream.yaml
+  - rstudio-gpu-notebook-imagestream.yaml
+  - jupyter-rocm-minimal-notebook-imagestream.yaml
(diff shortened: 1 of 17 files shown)


manifests/base/kustomization.yaml
```diff
@@ -15,355 +15,647 @@ resources:
...
       name: notebooks-parameters
-      apiVersion: v1
-    fieldref:
-      fieldpath: data.odh-minimal-notebook-image-n-1
-  - name: odh-generic-data-science-notebook-image-n
-    objref:
+      version: v1
+    targets:
+      - fieldPaths:
+          - spec.tags.1.from.name
```
jiridanek, line 58: Chatbot suggests this can be written

```
        fieldPaths:
          - spec.template.spec.containers.[name=myapp].image
```

So instead of the weird `.1.`, there may be other possibilities. Imo not very useful here, but may be useful for notebook-controller, maybe.

jstourac: Hm, so you mean we would go with something like:
```
      - fieldPaths:
          - spec.tags.[name=2025.1].from.name
```

This could be a bit more clearer when reading and would be more resistant to errors in case somebody will be changing the order of the tags in the ImageStreams, true. I shall consider your proposal.

I'm just wondering whether we won't introduce a bit more necessary changes once we will be introducing a new image versions (e.g. 2025.2), which will lead to update also this part :thinking:

jstourac: I updated this PR with the changes that allows this to be build by the latest kustomize version out there as of today (5.6.0). And just realized about this your proposal. I'm not against it, but let's check the others opinion. We can also do it as a separate change if needed. WDYT?

jiridanek: > I'm not against it, but let's check the others opinion. We can also do it as a separate change if needed. WDYT?

certainly a separate change!

manifests/base/kustomization.yaml
```diff
@@ -21,355 +21,644 @@ resources:
...
+      - commit.env
+    name: notebook
 generatorOptions:
   disableNameSuffixHash: true
-vars:
-  - name: odh-minimal-notebook-image-n
-    objref:
+
+labels:
+  - includeSelectors: true
```
jiridanek, line 36: got confused by docs whether we do or do not want these includeXYZ settings; will look again later

jiridanek:
This should be easy to check. Render before, render after, compare. Then do a rhds branch, render before, merge this in, render after, compare. But it's too late, so will do that Monday.

jstourac:
Yep, feel free to inspire by the pr description  😉

jstourac:
/override ci/prow/notebook-rocm-jupyter-tf-ubi9-python-3-11-pr-image-mirror

jstourac:
/override ci/prow/images
/override ci/prow/notebook-cuda-jupyter-pt-ubi9-python-3-11-pr-image-mirror
/override ci/prow/notebook-cuda-jupyter-ubi9-python-3-11-pr-image-mirror
/override ci/prow/notebook-jupyter-ds-ubi9-python-3-11-pr-image-mirror
/override ci/prow/notebook-jupyter-tai-ubi9-python-3-11-pr-image-mirror
/override ci/prow/notebook-jupyter-ubi9-python-3-11-pr-image-mirror
/override ci/prow/notebook-rocm-jupyter-pt-ubi9-python-3-11-pr-image-mirror
/override ci/prow/notebook-rocm-jupyter-ubi9-python-3-11-pr-image-mirror
/override ci/prow/notebooks-ubi9-e2e-tests
/override ci/prow/rocm-notebooks-e2e-tests

PR #1011, opened 2025-04-09
Updates

jstourac:
## Merge criteria:


Makefile
```diff
@@ -19,7 +19,7 @@ endif
 .RECIPEPREFIX =
 
 IMAGE_REGISTRY   ?= quay.io/opendatahub/workbench-images
-RELEASE	 		 ?= 2024b
+RELEASE	 		 ?= 2025a
```
jstourac, line 22: Shall I review also our ocp-ci jobs? Do we e.g. care about this: https://github.com/openshift/release/blob/master/ci-operator/config/opendatahub-io/notebooks/.config.prowgen#L8 ?

Makefile
```diff
@@ -36,7 +36,7 @@ endif
 DATE 		?= $(shell date +'%Y%m%d')
 WHERE_WHICH ?= which
 
-ifeq ($(RELEASE), 2025a)
+ifeq ($(RELEASE), 2025b)
 	RELEASE_PYTHON_VERSION = 3.12
```
jstourac, line 40: I'm truly not sure where did this come from, @andyatmiami ?

jiridanek:
/lgtm

and @atheo89 also seems agreeable in https://redhat-internal.slack.com/archives/C060A5FJEAD/p1744195078043939

we can discuss adding label as separate discussion

office hours are next monday, that feels too long to wait for resolution of this

Makefile
jstourac: proper fix would mean removing this case at all as we don't know for sure which image version will introduce the python 3.12 yet

jiridanek: and we could remove the 2024 case because we don't care that there we had python 3.9

atheo89: I agree, probably it would be better to remove this whole block

andyatmiami: as the (unfortunately misguided) person that added this code block - yes please remove.

it was an ill-fated attempt to "future-proof" our code when we first started on 2025a... that should have been removed before I submitted my PR (it maybe had legs when we were going to have 2024a and 2024b sub-directories - but now adds no value and worse, contributes to confusion/mental overhead!)

PR #1012, opened 2025-04-09
NO-JIRA: add a 'release' label to our images

jstourac:
Let's have a 'release' label in our images we build so that we can
easily filter them out. This change includes this update for the:
* local build via Makefile
* konflux build via .tekton pipeline definitions

There is a third place we should update - the openshift-ci job
definitions.

## Description
Note that in the end I used key `release` instead of the originally on the call announced `version`. Reason is to become compliant to the name of the variable we have there already.

## Merge criteria:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ define build_image
 	$(info # Building $(IMAGE_NAME) image...)
 
 	$(ROOT_DIR)/scripts/sandbox.py --dockerfile '$(2)' -- \
-		$(CONTAINER_ENGINE) build $(CONTAINER_BUILD_CACHE_ARGS) --tag $(IMAGE_NAME) --file '$(2)' $(BUILD_ARGS) {}\;
+		$(CONTAINER_ENGINE) build $(CONTAINER_BUILD_CACHE_ARGS) --label release=${RELEASE} --tag $(IMAGE_NAME) --file '$(2)' $(BUILD_ARGS) {}\;
 endef
 
 # Push function for the notebook image:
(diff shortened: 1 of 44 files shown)


ci/cached-builds/makefile_helper.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+
+import pathlib
+import platform
+import subprocess
+import sys
+
+def exec_makefile(target: str, makefile_dir: pathlib.Path | str, options: list[str] = []) -> str:
```
jiridanek, line 8: I don't like it. The value of the original function was in the `-nps` args, that made it do something that people who did not read the make docs may not know is possible. This turns the thing into a rather uninteresting wrapper over `subprocess.run`.

Makefile
```diff
@@ -471,3 +471,8 @@ all-images: jupyter-minimal-ubi9-python-$(RELEASE_PYTHON_VERSION) \
 	rocm-jupyter-pytorch-ubi9-python-$(RELEASE_PYTHON_VERSION) \
 	rocm-runtime-pytorch-ubi9-python-$(RELEASE_PYTHON_VERSION) \
 	rocm-runtime-tensorflow-ubi9-python-$(RELEASE_PYTHON_VERSION)
+
+# This is used primarly for konflux_generate_component_build_pipelines.py to we know the build release version
+.PHONY: print-release
+print-release:
+	@echo "$(RELEASE)"
```
jiridanek, line 478:
```
❯ gmake -nps | grep '^RELEASE '
+ which podman
+ pwd
RELEASE = 2024b
```

already sort of does what is needed, and we were pulling out stuff from Makefile with `-nps` before. If you really want custom target, add it to the list on top as well

https://github.com/opendatahub-io/notebooks/blob/4c99b519571fa9b3099b56818006c48404fcf84c/Makefile#L105

ci/cached-builds/gen_gha_matrix_jobs.py
```diff
@@ -23,41 +22,13 @@
 project_dir = pathlib.Path(__file__).parent.parent.parent.absolute()
 
 
-def parse_makefile(target: str, makefile_dir: pathlib.Path | str) -> str:
-    # Check if the operating system is macOS
```
jiridanek, line 27: This would deserve a doc comment about beauty of -nps

jiridanek (dismissed):

Makefile
jstourac: I was also considering that approach too... but having a separate target seemed a bit more clean solution to me. That's why I went this way.

jstourac:
thank you for your review

/approve

PR #1018, opened 2025-04-15
RHOAIENG-20553 - CSS is broken when loading the TensorBoard extension

daniellutz:
This will fix [RHOAIENG-20553 - CSS is broken when loading the TensorBoard extension](https://issues.redhat.com/browse/RHOAIENG-20553)

Tensorboard provides an IFrame element inside Jupyter Notebooks so it can display all the visualization toolkit properly. Patternfly did implement some changes that affected the IFrame from Tensorboard, making it smaller. This fix aims to provide a clean solution without affecting any other components and styling.

## Description
Tensorboard provides a visualization toolkit for TensorFlow and it provides its functionality [through the use of an IFrame](https://github.com/tensorflow/tensorboard/blob/aebdc82a3e4ec282a21dc6bfc7806e8ea2c42246/tensorboard/notebook.py#L365-L368) inside Jupyter Notebook, as follows:

```python
    frame_id = "tensorboard-frame-{:08x}".format(random.getrandbits(64))
    shell = """
      <iframe id="%HTML_ID%" width="100%" height="%HEIGHT%" frameborder="0">
      </iframe>
```
https://github.com/tensorflow/tensorboard/blob/aebdc82a3e4ec282a21dc6bfc7806e8ea2c42246/tensorboard/notebook.py#L365-L368

The main issue here lies on the use of reset CSS for global elements, where elements like iframe will receive normal state, in our case it will receive a `height: auto` attribute, which will break the default `height: 800px` of the IFrame element.

Since the PatternFly in our case is only being used to load the spinner at the start of the Jupyter Notebooks, we will simply replace the CSS used, going from the `minimal` to the `no-reset` option.

## References:
- [patternfly/patternfly: fix(global): remove woff, use where with globals, cleanup #5435](https://github.com/patternfly/patternfly/pull/5435)
- [patternfly/src/patternfly/base/normalize.scss](https://github.com/patternfly/patternfly/blob/e31a56f8c8994136026dab6fc74f72be5d00e7d3/src/patternfly/base/normalize.scss#L72-L82)
- [tensorboard/notebook.py](https://github.com/tensorflow/tensorboard/blob/aebdc82a3e4ec282a21dc6bfc7806e8ea2c42246/tensorboard/notebook.py#L365-L368)
- [notebooks/jupyter/utils/addons/apply.sh](https://github.com/opendatahub-io/notebooks/blob/main/jupyter/utils/addons/apply.sh#L32)

## Building
To build the images properly with the changes, the following command has been executed:

```bash
make cuda-jupyter-tensorflow-ubi9-python-3.11 \
    -e IMAGE_REGISTRY="quay.io/dlutz/workbench-images" \
    -e RELEASE="2025a" \
    -e CONTAINER_BUILD_CACHE_ARGS="" \
    -e PUSH_IMAGES="yes"
```

The following image is available for testing on Quay.io:
[quay.io/dlutz/workbench-images:cuda-jupyter-tensorflow-ubi9-python-3.11-2025a_20250415](https://quay.io/repository/dlutz/workbench-images?tab=tags&tag=cuda-jupyter-tensorflow-ubi9-python-3.11-2025a_20250415)

## How Has This Been Tested?
This has been tested in an OpenShift Local cluster.

### Testing on Cluster
To test in your cluster you should:

- the build process did take place and completed;
- the image on Quay.io is available;
- you are connected to your OpenShift cluster through the `oc login` command;

1. Now you can run the `make deploy9-` command to deploy the TensorFlow image:

```bash
$ make deploy9-cuda-jupyter-tensorflow-ubi9-python-3.11 \
     -e IMAGE_REGISTRY="quay.io/dlutz/workbench-images" \
     -e RELEASE="2025a" \
     -e CONTAINER_BUILD_CACHE_ARGS="" \
     -e PUSH_IMAGES="yes"
```

2. Create a port-forward mapping in your system:

```bash
$ kubectl port-forward jupyter-tensorflow-ubi9-python-3-11-notebook-0 8888:8888
```

3. Open your browser, pointing to `localhost`, on port `8888` and with the `jovyan` controller:

- [http://localhost:8888/notebook/opendatahub/jovyan/](http://localhost:8888/notebook/opendatahub/jovyan/)

4. Now run each of the following commands in a Jupyter Notebook cell (if you are running on a different scenario, please, adjust the `TENSORBOARD_PROXY_URL` accordingly)

```python
import os
```
```python
os.environ["TENSORBOARD_PROXY_URL"] = "/notebook/opendatahub/jovyan/proxy/6006/"
```
```python
%load_ext tensorboard
```
```python
%tensorboard --logdir /opt/app-root/src/shared
```

Note: originally, on the Jira ticket, the description uses the NB_PREFIX variable, [that is available through the Dashboard deployment](https://github.com/search?q=repo%3Aopendatahub-io%2Fodh-dashboard%20NB_PREFIX&type=code), but not available on the make deploy commands, this is why we have used the direct reference to the controller here.

## Merge criteria:

### DIFF
--- a/jupyter/utils/addons/apply.sh
+++ b/jupyter/utils/addons/apply.sh
@@ -6,7 +6,7 @@
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
-pf_url="https://unpkg.com/@patternfly/patternfly@6.0.0/patternfly.min.css"
+pf_url="https://unpkg.com/@patternfly/patternfly@6.0.0/patternfly-no-globals.css"
 
 static_dir="/opt/app-root/share/jupyter/lab/static"
 index_file="$static_dir/index.html"
@@ -30,12 +30,6 @@ if [ ! -f "$body_file" ]; then
 fi
 
 curl -o "$static_dir/pf.css" "$pf_url"
-
-# Patternfly implemented some changes, setting the default height on iframe objects to "auto", which
-# excluded the default 800px height of Tensorboard's IFrame object. This stylesheet will simply search
-# for all IFrames that contain `tensorboard` on its URL and add a default height of 800px to it, and only
-# on it.
-echo "iframe[id*=\"tensorboard\"] { height: 800px !important }" >> "$static_dir/pf.css"
 
 head_content=$(tr -d '\n' <"$head_file" | sed 's/@/\\@/g')
 body_content=$(tr -d '\n' <"$body_file" | sed 's/@/\\@/g')
--- a/jupyter/utils/addons/apply.sh
+++ b/jupyter/utils/addons/apply.sh
@@ -31,6 +31,12 @@ fi
 
 curl -o "$static_dir/pf.css" "$pf_url"
 
+# Patternfly implemented some changes, setting the default height on iframe objects to "auto", which
+# excluded the default 800px height of Tensorboard's IFrame object. This stylesheet will simply search
+# for all IFrames that contain `tensorboard` on its URL and add a default height of 800px to it, and only
+# on it.
+echo "iframe[id*=\"tensorboard\"] { height: 800px !important }" >> "$static_dir/pf.css"
+
 head_content=$(tr -d '\n' <"$head_file" | sed 's/@/\\@/g')
 body_content=$(tr -d '\n' <"$body_file" | sed 's/@/\\@/g')
 


jupyter/utils/addons/apply.sh
```diff
@@ -31,6 +31,12 @@ fi
 
 curl -o "$static_dir/pf.css" "$pf_url"
 
+# Patternfly implemented some changes, setting the default height on iframe objects to "auto", which
```
jiridanek, line 34: Isn't there still some `no-reset` version of patternfly that does not do css reset?

I found `/@patternfly/react-core/dist/styles/base-no-reset.css` in https://www.npmjs.com/package/@patternfly/react-core?activeTab=code. Looking for a better place where to get it.

jupyter/utils/addons/apply.sh
```diff
@@ -31,6 +31,12 @@ fi
...
+# excluded the default 800px height of Tensorboard's IFrame object. This stylesheet will simply search
+# for all IFrames that contain `tensorboard` on its URL and add a default height of 800px to it, and only
+# on it.
+echo "iframe[id*=\"tensorboard\"] { height: 800px !important }" >> "$static_dir/pf.css"
```
jiridanek, line 38: Nice hack, but being able to have patternfly without css reset would be much better. Who knows what else is patternfly breaking and we just don't know about it. Customers may have their own extensions like tensorboard that are broken by different patternfly css rules.

daniellutz: My main concern, at least right now, is if taking out the CSS reset out would break other layout or other implementations. I would like to have this fixed right now WITH the reset and, if possible, we could try later implementing the no-reset version and checking if everything still works as expected with proper time for testing.

jiridanek:
It's a workable solution, if it fixes current problem, I have no problem
/lgtm

jupyter/utils/addons/apply.sh
daniellutz: My main concern now is, if some of our layout, grid, etc, rely on the default reset values for our current implementation, then simply removing it could break a lot of our layout and styling. I would prefer to add a task to the next sprints so we can remove and properly check and test if it's ok.

daniellutz:
I would like to create a Jira for next sprints, if possible, to try the patternfly with no-reset option, see if our layout and components are fine, then we can replace this and the current pf.css file.

I just wanted to understand and see how it goes with no reset and give some proper time for testing.

daniellutz:
I am giving a try to the no-reset option right now, if it goes as expected, we might change this solution to simply replacing the pf.css file. Please, hold a little bit the PR.

jupyter/utils/addons/apply.sh
jiridanek: @caponetto ^^^ you were right saying on slack that adding the whole patternfly on top of jupyterlab was not the most fortunate move, it makes things harder down the line

jstourac: > My main concern, at least right now, is if taking out the CSS reset out would break other layout or other implementations.

@daniellutz AFAIK, the only reason we added this patternfly thing was for the initial loading spinner at the beginning of the workbench IDE load. So unless this thing/feature breaks, we don't need to care about anything else.

jstourac:
As Jiri wrote already - this is a hack to workaround an issue we introduced by our "hack". I would like to see the different solution for us.

1. what you propose with the no-reset may be a way
2. but what about simply extracting only the required minimum from the css file creating our own with just a couple of things that we can restrict to the particular spinner html element only without any fear breaking anything else???

---

If we really rush on this, then:
/lgtm

but I am not in favor of having another hack in our hack :stuck_out_tongue:

jupyter/utils/addons/apply.sh
jiridanek: exactly as @jstourac says, that's why I'm calling out caponetto's prescient words

jiridanek:
> but what about simply extracting only the required minimum from the css file creating our own with just a couple of things that we can restrict to the particular spinner html element only without any fear breaking anything else???

this is what @ederign wants us to do; there is a plugin for webpack that will do the extraction for you automatically

it's called tree-shaking in the web dev world

jupyter/utils/addons/apply.sh
jiridanek: Asked in internal Slack at https://redhat-internal.slack.com/archives/C04JMHKSD9C/p1744730368268949, @mcoker informed us that

> it's called "no-globals" now:  https://unpkg.com/@patternfly/patternfly@6.0.0/patternfly-no-globals.css

atheo89:
Hey all, I took a look as well, it /lgtm to resolve that bug. I am not familiar with patternfly, so I leave that to the experts of the field! 
I run this tutorial from TensorBoard official site: https://www.tensorflow.org/tensorboard/tensorboard_in_notebooks and i got the below result. 

However, without this `os.environ["TENSORBOARD_PROXY_URL"]= os.environ["NB_PREFIX"]+"/proxy/6006/"` it has difficulties to be displayed. 

Screenshot 2025-04-16 at 12 01 12

jiridanek:
@atheo89 known issue https://issues.redhat.com/browse/RHODS-4799

andyatmiami:
/lgtm

running through verification steps provided by @daniellutz - I see the properly sized Tensorboard iframe now rendering...

image

PR #1023, opened 2025-04-18
Fix runtime images

daniellutz:
For the release process, two things are missing:

1. some ImageStream objects have yet the `quay.io/opendatahub-io` URLs instead of the `quay.io/modh`.
2. in the merge commit from upstream to downstream, JSON files that should have been removed are still present.

This PR aims to fix these two issues.

## Merge criteria:


daniellutz:
Should target downstream, not upstream. Will create another PR.

Issue #631, opened 2024-07-23
When transforming branch name into image label while running CI checks, take care not to create invalid label


jiridanek:
> Error: tag ghcr.io/red-hat-data-services/notebooks/workbench-images:base-ubi8-python-3.8-renovate/registry.access.redhat.com-ubi9-ubi-minimal-9.x_fcb4541c34b36a9781c20d9164d58d9e6a57a097: invalid reference format

https://github.com/red-hat-data-services/notebooks/actions/runs/10055394214/job/27791966809?pr=287#step:9:15

> Error: unable to parse value provided `[ghcr.io/RomanFilip/notebooks-fork/workbench-images/build-cache]` to --cache-to: invalid repo "ghcr.io/RomanFilip/notebooks-fork/workbench-images/build-cache": must contain registry and repository: invalid reference format: repository name must be lowercase

https://github.com/RomanFilip/notebooks-fork/actions/runs/13499383840/job/37713825404#step:17:95

jiridanek:
```
      - name: downcase REPO
        run: |
          echo "REPO=${GITHUB_REPOSITORY,,}" >>${GITHUB_ENV}
```

https://github.com/orgs/community/discussions/25768

jiridanek:
```
      - name: downcase IMAGE_REGISTRY and CACHE
        run: |
          echo "IMAGE_REGISTRY=${IMAGE_REGISTRY,,}" >>${GITHUB_ENV}
          echo "CACHE=${CACHE,,}" >>${GITHUB_ENV}
```

jiridanek:
not completely fixed, we did not address the "branch name too long" source of problems

PR #1027, opened 2025-04-24
Issue opendatahub-io/notebooks#631: cut the ref name to avoid going over 128 character limit for Docker image tag

jiridanek:
## Description

* https://github.com/opendatahub-io/notebooks/issues/631

## How Has This Been Tested?

* https://github.com/jiridanek/notebooks/actions/runs/14642835456
* https://github.com/jiridanek/notebooks/actions/runs/14643403904

## Merge criteria:

### DIFF
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -209,7 +209,10 @@ jobs:
       - name: Calculate image name and tag
         id: calculated_vars
         run: |
-          SANITIZED_REF_NAME=$(echo "${{ github.ref_name }}" | sed 's/[^a-zA-Z0-9._-]/_/g')
+          # Need for sanitization explained in https://github.com/opendatahub-io/notebooks/issues/631
+          # For length, Docker image tags have 128-character limit, and we form them as <inputs.target>-<ref_name>_<sha>
+          # therefore since sha is 40 characters, and our target names are <40 chars, we should cut ref_name at 40
+          SANITIZED_REF_NAME=$(echo "${{ github.ref_name }}" | sed 's/[^a-zA-Z0-9._-]/_/g') | cut -c 1-40
           IMAGE_TAG="${SANITIZED_REF_NAME}_${{ github.sha }}"
 
           echo "IMAGE_TAG=${IMAGE_TAG}" >> "$GITHUB_OUTPUT"


jiridanek:
/override ci/prow/images

not applicable

.github/workflows/build-notebooks-TEMPLATE.yaml
```diff
@@ -209,7 +209,8 @@ jobs:
       - name: Calculate image name and tag
         id: calculated_vars
         run: |
-          SANITIZED_REF_NAME=$(echo "${{ github.ref_name }}" | sed 's/[^a-zA-Z0-9._-]/_/g')
+          # need for sanitization explained in https://github.com/opendatahub-io/notebooks/issues/631
+          SANITIZED_REF_NAME=$(echo "${{ github.ref_name }}" | sed 's/[^a-zA-Z0-9._-]/_/g') | cut -c 1-50
```
andyatmiami, line 213: apologies on lacking proper vision here - but i would appreciate an example in the comments and/or explanation on how/why "50" is the magic number here in terms of chopping the ref name...

We have 78 characters "fixed" here in this situation leaving only 50 remaining to play with?  what are those 78 chars reserved for?

Note i have no problem with the implementation - just a question on the context in which this operates.

andyatmiami: and then... from the original "root cause" that spawned this issue - we see that the branch name _in offense_ was:

`renovate/registry.access.redhat.com-ubi9-ubi-minimal-9.x` (which is 56 characters)

andyatmiami: I'm not sure the "take first 50 chars from `github.ref_name`" actually solves the root issue though?

If we look at the "fixed 78 characeters" I alluded to earlier - to answer my own question - that comes from:

`ghcr.io/red-hat-data-services/notebooks/workbench-images:base-ubi8-python-3.8-`

However, while _some of this_ is certainly "fixed" - this portion is more variadic in length:
- `base-ubi8-python-3.8-`

and presumably something like an "accelerated" `tensorflow` image would have a "base" length of > 78 characters.. meaning "chopping the first 50" heuristic would **still** lead to failure, yes?

jiridanek: If you want to fix it some smarter way, pls open a new PR and ask Diamond to lgtm it for you, so we can get the Dependabot thing in ;P

andyatmiami: I'm not sure the presence of those runs passing actually verifies the concern I am raising...

Looking at the output from one of those runs - I see this:

`echo "OUTPUT_IMAGE=ghcr.io/jiridanek/notebooks/workbench-images:rocm-runtime-tensorflow-ubi9-python-3.11-${IMAGE_TAG}" >> "$GITHUB_OUTPUT"`

Our "fixed" portion here is 86 characters...

I would imagine if you had named your GH branch to something that was 51 characters long ( to force the `cut` to "do something" ) - we would/should/could expect to see the same failure on the tag name being > 128 characters..

Also interestingly... it looks like something is breaking... if I look at the `push` output on the final value... I see this:

```
# Building ghcr.io/jiridanek/notebooks/workbench-images:rocm-runtime-tensorflow-ubi9-python-3.11-_8e9f8391aa6b5bb3484b0b46ee1819ee15c524f0 image...
# Pushing ghcr.io/jiridanek/notebooks/workbench-images:rocm-runtime-tensorflow-ubi9-python-3.11-_8e9f8391aa6b5bb3484b0b46ee1819ee15c524f0 image...
```

so this seems to be not-quite-working?

```
IMAGE_TAG="${SANITIZED_REF_NAME}_8e9f8391aa6b5bb3484b0b46ee1819ee15c524f0"
```

As `SANITIZED_REF_NAME` is an empty string 🤔

andyatmiami: Also the inclusion of that hash value further eats into our 50 chars "buffer" we are playing with (i think...)

andyatmiami: sorry, should have included this earlier - been looking at output here:
- https://github.com/jiridanek/notebooks/actions/runs/14642835456/job/41089248058

andyatmiami: ~maybe just relying on the hash and not doing **ANYTHING** with `SANITIZED_REF_NAME` "solves" our problem?~

that is a 40 characters...

which then doesn't fully help... given the following is 138 characters:
- `ghcr.io/red-hat-data-services/notebooks/workbench-images:rocm-runtime-tensorflow-ubi9-python-3.11-8e9f8391aa6b5bb3484b0b46ee1819ee15c524f0`

jiridanek: hmm, yeah, right, still gotta cut more agressively

jiridanek: Try harder!

PR #1024, opened 2025-04-20
RHOAIENG-24181: chore(jupyter/utils/addons): tree-shake the PatternFly CSS used to add spinner to JupyterLab-based workbenches

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-24181

Follows up on

* https://github.com/opendatahub-io/notebooks/pull/1029

## Description

This project provides add-ons for JupyterLab workbenches, primarily addressing issues related to initial load times and styling. It uses PurgeCSS to optimize PatternFly CSS and webpack to bundle JavaScript files. The add-ons improve the user experience by providing better feedback during JupyterLab's loading process and fixing CSS issues encountered with extensions like TensorBoard. The core functionality involves injecting HTML snippets and optimized CSS into the JupyterLab environment.

### Key Components
apply.sh: This shell script is the primary entry point for integrating the add-ons into a JupyterLab environment during the Docker build process. It injects the HTML snippets (partial-head.html, partial-body.html) and the optimized CSS (dist/pf.css) into the JupyterLab instance.

partial-head.html, partial-body.html: These HTML files contain snippets that are injected into the <head> and <body> sections of the JupyterLab HTML, respectively. They are responsible for adding the loading spinner and other UI elements.

dist/pf.css: This file contains the tree-shaken and optimized PatternFly CSS, minimizing the amount of CSS loaded by JupyterLab.

webpack.config.ts: This file configures the webpack build process, including the tree-shaking of CSS and bundling of JavaScript.

cleanup-webpack-plugin.mts: This custom webpack plugin removes unnecessary files generated during the build process, keeping the output clean.

test-build.sh: This script verifies the effectiveness of the tree-shaking process by analyzing the generated CSS file size.

The project uses pnpm for package management and provides several scripts for development, building, and testing. The dist/ directory contains the final build output, which is intended to be included in a JupyterLab Docker image.

## How Has This Been Tested?

Played with the demo page in `dist/`. Did not try it live, want to get a conceptual ok from someone before I spend any more time polishing this.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #1029, opened 2025-04-25
RHOAIENG-24684: fix(jupyter/utils/addon) fix vertical centering of the PatternFly spinner

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-24684

Follows up on

* https://github.com/opendatahub-io/notebooks/pull/1018

## Description

After we switched to the css without global css reset, we no longer have the rule

```
  :where(html, body) {
    height: 100%;
  }
```

present, and it causes the spinner to be tucked to the top edge of screen.

image

## How Has This Been Tested?

Played with it a while.

I've checked that setting this does not visibly change how JupyterLab (with TensorBoard open) displays.

## Merge criteria:

### DIFF
--- a/jupyter/utils/addons/partial-head.html
+++ b/jupyter/utils/addons/partial-head.html
@@ -3,7 +3,11 @@
   type="text/css"
   href="{{page_config.fullStaticUrl}}/pf.css"
 />
+<!-- set height to 100% so that vertical centering works -->
 <style>
+  :where(html, body) {
+    height: 100%;
+  }
   #loading-container {
     width: 100%;
     height: 100%;


jstourac:
Okay, thank you for this. In general, this is LGTM. I have couple of questions:
1. wouldn't be better to track this as a separate issue (Jira) than attaching it to the existing and closed one? the fix for this will end up in 2.21 and not in 2.20 as the original issue is saying
2. does this have any affect #1024?

I'll give the lgtm, but I would rather see these questions above answered first.

/lgtm

jiridanek:
I guess I can open special Jira just for this

this has no impact on the tree shaking

PR #1033, opened 2025-05-04
RHOAIENG-23584: test(run-rstudio.sh): fix tests to more correctly check for exposed env variables

jiridanek:
https://issues.redhat.com/browse/RHOAIENG-23584

## Description

Previously, we were not checking env variables carefully enough. It is important to ensure test is passing because R/RStudio set the variable and not because the variable got inherited. RStudio blocks the normal parent/child inheritance.

Having considered multiple options of testing, the last option was chosen

1. check /usr/lib64/R/etc/Renviron.site it contains the variable
2. start rstudio, connect to it, switch to the terminal tab (needs selenium/etc) and check that the bash process has the env variable set, either directly (/proc/pid/environ, or type env in the browser)
3. run R, same as now, but before that, purge its env, so it is inheriting nothing and has to get it on startup

## How Has This Been Tested?

```
uv run pytest tests/containers/workbenches/rstudio/rstudio_test.py --image=ghcr.io/jiridanek/notebooks/workbench-images:rstudio-rhel9-python-3.11-_54e28ab963f7ebd77b6529a36a8d58493d9f0a0a --image=ghcr.io/jiridanek/notebooks/workbench-images:rstudio-rhel9-python-3.11-_ddf8a72c75ae675918f82760e6cb8b919f6c78d4
```

First image is current main, second is Diamond's PR.

## Merge criteria:

### DIFF
--- a/tests/containers/workbenches/rstudio/rstudio_test.py
+++ b/tests/containers/workbenches/rstudio/rstudio_test.py
@@ -107,6 +107,25 @@ class TestRStudioImage:
         finally:
             docker_utils.NotebookContainer(container).stop(timeout=0)
 
+    @allure.issue("RHOAIENG-23584")
+    def test_arbitrary_env_propagates_unchanged(self, rstudio_image: str) -> None:
+        """
+        Checks that environment variables are propagated into the RStudio environment.
+        """
+
+        container = WorkbenchContainer(image=rstudio_image, user=1000, group_add=[0])
+        container.with_env("SOME_VARIABLE", "Some Value")
+
+        try:
+            # We need to wait for the IDE to be completely loaded so that the envs are processed properly.
+            container.start(wait_for_readiness=True)
+
+            # Once the RStudio IDE is fully up and running, the processed envs should include the variable.
+            assert_has_env_variable(container, "SOME_VARIABLE", "Some Value")
+
+        finally:
+            docker_utils.NotebookContainer(container).stop(timeout=0)
+
     @allure.issue("RHOAIENG-16604")
     def test_http_proxy_env_propagates(self, rstudio_image: str, subtests: pytest_subtests.plugin.SubTests) -> None:
         """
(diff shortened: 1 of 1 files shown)


tests/containers/workbenches/rstudio/rstudio_test.py
```diff
@@ -120,9 +120,12 @@ class TestCase(NamedTuple):
             value: str
 
         test_cases: list[TestCase] = [
+            # RHOAIENG-16604: proxy
             TestCase("HTTP_PROXY", "http_proxy", "http://localhost:8080"),
             TestCase("HTTPS_PROXY", "https_proxy", "https://localhost:8443"),
             TestCase("NO_PROXY", "no_proxy", "google.com"),
+            # RHOAIENG-xxx: generic vars, this is not supposed to be passing
+            TestCase("some_random_variable", "some_random_variable", "some_random_value"),
```
jiridanek, line 128: We can be checking `/proc/$pid/environ` of the`/usr/lib/rstudio-server/bin/rsession` process.

Or we can try to make the rstudioapi thing work; is not working for me, but maybe... https://rstudio.github.io/rstudioapi/

jiridanek: If everything else fails, we can write e2e web tests.

jiridanek: the check for lowercase version of the variable is actually valuable, it does check something valuable.

so the test is only about 11% buggy, I'd estimate ;p

jiridanek:
```
# We need to wait for the IDE to be completely loaded so that the envs are processed properly.
container.start(wait_for_readiness=False)
```

is a bit of a mystery. SInce the `R -e` does not run in the RStudio session, why does it matter if RStudio did fully start or not? (And it indeed matters, I checked that for myself.)

jstourac: good catch about the test pass even if it shouldn't we should look into it

jiridanek: I'm trying to compare

```
--image=ghcr.io/jiridanek/notebooks/workbench-images:rstudio-rhel9-python-3.11-_54e28ab963f7ebd77b6529a36a8d58493d9f0a0a --image=ghcr.io/jiridanek/notebooks/workbench-images:rstudio-rhel9-python-3.11-_ddf8a72c75ae675918f82760e6cb8b919f6c78d4
```

the first one is current main and the second is with Diamond's fix. Looking for a test that would show difference between them.

jiridanek: > is a bit of a mystery

no mystery at all, the test is checking for the lowercase version of the variable, and that is only available after `/usr/lib64/R/etc/Renviron.site` has the thing written in. So we need the RStudio startup to finish. The uppercase version of the variable is available immediately on commandline, and never in rsession in RStudio.

jiridanek: Regarding more robust test, I don't have any good ideas. Some poor ones

* check /usr/lib64/R/etc/Renviron.site it contains the variable
* start rstudio, connect to it, switch to the terminal tab (needs selenium/etc) and check that the bash process has the env variable set, either directly (/proc/pid/environ, or type env in the browser)
* run R, same as now, but before that, purge its env, so it is inheriting nothing and has to get it on startup

I'll update PR to do the third option.

jstourac:
Thank you for this fix, Jiri - I like it! :slightly_smiling_face: 

There is the python linter fail, so if you could update this, please. But otherwise:

/lgtm

jiridanek:
Thanks, I'll merge after the all vars fix gets in

* https://github.com/opendatahub-io/notebooks/pull/1037

PR #1045, opened 2025-05-08
(WIP) RHOAIENG-24411: Upgrade Juptyer Minimum to Python 3.12

dibryant:
Fixes for https://issues.redhat.com/browse/RHOAIENG-24411
## Description

Upgrade Jupyter Minimum to Python 3.12. I put WIP because we want to work on testing before merging. 
## How Has This Been Tested?

$ make jupyter-minimal-ubi9-python-3.12
     -e RELEASE_PYTHON_VERSION=3.12
     -e IMAGE_REGISTRY="quay.io/${YOUR_USER}/workbench-images" \
     -e RELEASE="2025b" \
     -e CONTAINER_BUILD_CACHE_ARGS="" \
     -e PUSH_IMAGES="no"
ensuring the Notebook works properly:
$ export QUAY_IMAGE=$(podman images --format "{{.Repository}}:{{.Tag}}" | head -1)
$ podman run -it -p 8888:8888 $QUAY_IMAGE

after opening URL 
import sys
print(sys.version)

## Merge criteria:


jiridanek:
> RHOAIENG-24411: Upgrade Juptyer Minimum to Python 3.12

Is this the correct PR title? The PR is creating entirely new 3.12 image, it is not upgrading anything.

dibryant:
> > [RHOAIENG-24411](https://issues.redhat.com//browse/RHOAIENG-24411): Upgrade Juptyer Minimum to Python 3.12
> 
> Is this the correct PR title? The PR is creating entirely new 3.12 image, it is not upgrading anything.

Yes sorry changed the title would better word it next time

jiridanek:
```
make jupyter-minimal-ubi9-python-3.12 \
     -e RELEASE_PYTHON_VERSION=3.12
```

That's not a bad idea on its own, but sadly it does not work with Github actions in this repo. For now, let me just add -3.12 entry to Makefile.

jiridanek:
And we've got some CI tests!

```
======================================================================
FAIL: test_version (__main__.TestPythonVersion.test_version)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/ipykernel_38/3768342736.py", line 31, in test_version
    self.assertEqual(actual_major_minor, expected_major_minor, "incorrect version")
AssertionError: '3.12' != '3.11'
- 3.12
?    ^
+ 3.11
?    ^
 : incorrect version
```

That's actually somewhat nasty problem, similar to the Makefile issue above.

Now the question is, for how long do we intend to have the 3.11/3.12 Python versions in parallel? IIUC, for more than month, because Adriel wants to have at least some of the 3.12 images available on the side in RHOAI 2.21, ideally? Is that your understanding too?

If yes, we need to redo @andyatmiami 's test setup a little, so that we can handle building images for two Python versions in parallel.

andyatmiami:
Yes, the "parallel multiple python versions" was not something I considered when I parameterizing the python version in our targets 😢 

I was operating on the (apparently wrong!) assumption moving forward that we would do this type of thing _in totality_ on the `fast` channel... i.e. 3.11 would be totally swapped out for 3.12 within the repo...

Under that (again, wrong) assumption - not having the python version hardcoded in 80+ places seemed like a great idea...

I still do like the idea of the parameterization.. but supporting dual versions gets more complicated...

Clearly the base image defined in our all our various `Dockerfile` files becomes an issue (not sure if that is what @jiridanek was referencing w.r.t.  GHA problems)...

and then our image streams (which is where the python version is pulled from) gets screwing too:

```
{"name": "Python", "version": "v3.11"}
```

I guess I must admit I'm very confused on the technical direction of where we want to go - I had though all the discussions earlier in the year around "streamlining our build process" meant we did **NOT** need to "dual-support" images anymore... but if we are adding a new `ubi9-python-3.12` image folder to live alongside `ubi9-python-3.11`... how is this any different than what we have historically been doing??

jiridanek:
> I was operating on the (apparently wrong!)

It was correct assumption at the time! We had no way of knowing what Adriel will come up with in few months.

> (not sure if that is what @jiridanek was referencing w.r.t. GHA problems)...

github actions run papermill tests and those take the sw versions to check for from the manifests, so now they end up expecting python 3.11

> I guess I must admit I'm very confused on the technical direction of where we want to go

yeah, right! we have to clarify what's gotta go into 2.21, that's why I asked Diamond whether she already knows anything...

Adriel wants to "release in the fast channel a few Python 3.12 images". That scares me.

daniellutz:
my line of thinking is: if it's not supporting multiple scenarios, why our directories are named `ubiX-python-Y` ?

we could then simply name them `datascience/Dockerfile`, `minimal/Dockerfile`, etc, assuming the same Python version, same base OS, etc, across all of the images, it would be better

and this is not, in my understanding, a `knowing what Adriel will come up with in few months`, this is regarding the enterprise software that is delivered with the work done here.

we can't simply "change Python version and release them all at once", as the development takes some effort, testing takes some effort, other teams might depend on our work to also do theirs and, in a point of time, some teams will need to deliver some work before, even if it's internally, so they could support others.

I believe that this is not yet a matter of supporting two versions, it's more of having one version under development/testing for a future release, provide it to other teams to test as well, so we can plan the release accordingly.

andyatmiami:
@daniellutz 

Very fair point.. and I cannot claim to be right here - so this is good discussion...

But from my view point - all the motivation/rationale you described I was attributing to being "the old way" - and that by streamlining our release process... we could do away this all that extra effort...

If someone wants Python 3.12 (today) - I had thought (and clearly I could be wrong) - with our new release/support strategy - we **would** simply replace `Python 3.11` with `Python 3.12` and release it on the `fast` channel.

As to why our folders **still** have `python-3.11` in the name.. its because (in my possibly faulty world view) we had a lot of logic that parsed those folder names (particularly in Makefile) - and as we recently switched over to this "new release model" - it wasn't worth the effort to try and clean that up.

But, yes, in a perfect world... changing python versions should not requiring adding/updating dozens and dozens of files, folders, etc - it should be some type of parameter in our release process that "just works" with minimal configuration.

jiridanek:
> And this isn't, as I see it, about trying to 'know what Adriel will come up with in a few months'. This relates to the enterprise software that is delivered based on this work.

I was alluding to the "Agile values" there. The software should be in such a shape that whatever reasonable request "the business" comes up with, the wishes can be accommodated without excessive time spent on rework. The "Bill of Rights" things that were popular once.

https://www.informit.com/articles/article.aspx?p=2990402&seqNum=3#:~:text=Customer%20Bill%20of%20Rights&text=You%20have%20the%20right%20to%20see%20progress%20in%20a%20running,priorities%20without%20paying%20exorbitant%20costs.

jiridanek:
> It's more about having one version under development/testing for a future release

and under continuous integration, that's the one thing this PR is missing.

daniellutz:
@jiridanek and providing a way to build and test multiple Python versions would help on that, in a way that we can understand, test and ensure that the next upgrades/scenarios/requirements are in accordance (not yet talking about released versions on support maintenance, only the ability to build and test images)

I know that context is important and things were done the way they should back then, the same way that it's important to have discussions about the next steps, so let's figure out how this can happen in the best way possible :)

jiridanek:
Well, let's wait for the renaming pr from Harshad, and then we can create a new imagestream for this 3.12 image and we'd be pretty much done for now

jiridanek:
Next, we ask Adriel if he wants 3.12 version of every image in the repo, or about half, or just a few.

If just a few, then I'd just add makefile entry and imagestream in ad hoc way.

If he wants all, then we should think how to make doing that easi(er) and adopt such repo structure.

daniellutz:
For now, as far as I remember, the main idea was to build and test to see if the workbenches would work with Python 3.12 on the minimal and data science images, not the others, at least for now.

daniellutz (approved):

PR #1070, opened 2025-05-09
RHOAIENG-25226: Standardize the image name with image convention

harshad16:
Standardize the image name with image convention
- Change image name as per image name convention
- Split params.env into latest and previous
- Adjust runtime manifest in params format

## Description
Related-to: [RHOAIENG-25226](https://issues.redhat.com/browse/RHOAIENG-25226)

## How Has This Been Tested?

```
kustomize build manifests/base
```

## Merge criteria:


manifests/base/params-latest.env
```diff
@@ -0,0 +1,18 @@
+odh-workbench-jupyter-minimal-cpu-py311-ubi9-n=quay.io/opendatahub/workbench-images@sha256:36f48843b48b68ca8964758dad43dc0db4acb2b2d9c2f5936e1b54dc6da34019
```
atheo89, line 1: Need to update the lookup on the digest updater script to this new path and its filtering:
https://github.com/opendatahub-io/notebooks/blob/main/ci/sha-digest-updater.sh#L118

The same for the commit-latest.env

manifests/base/runtime-tensorflow-imagestream.yaml
```diff
@@ -16,7 +16,7 @@ spec:
   tags:
     - annotations:
         # language=json
-        opendatahub.io/runtime-image-metadata: |
+        opendatahub.io/runtime-image-metadata:
```
jiridanek, line 19: huh? this works?

manifests/base/runtime-minimal-imagestream.yaml
```diff
@@ -24,7 +24,7 @@ spec:
                   "minimal"
                 ],
                 "display_name": "Python 3.11 (UBI9)",
-                "image_name": "quay.io/opendatahub/workbench-images@sha256:e9cfdf68eb2b3150b4c4bf11f11efc8430172900f990b59d128ba830173137e3",
+                "image_name": odh-runtime-minimal-cpu-py311-ubi9-n_PLACEHOLDER,
```
atheo89, line 27: This will not work with parameters on the runtimes imagestreams.
The operator can not reconcile/translate the `odh-runtime-minimal-cpu-py311-ubi9-n_PLACEHOLDER` to `quay.io/opendatahub/workbench-images@sha256:e9cfdf68eb2b3150b4c4bf11f11efc8430172900f990b59d128ba830173137e3` (at least that happen in the older version on the kustomize)

manifests/base/runtime-tensorflow-imagestream.yaml
harshad16: json in yaml works, learned it today :)
https://alisoftware.github.io/yaml/2021/08/17/yaml-part1-json/#:~:text=cases%20of%20YAML.-,Back%20to%20Basics,-I%E2%80%99m%20sure%20most

manifests/base/runtime-minimal-imagestream.yaml
atheo89: I tried on a cluster with:

```
      devFlags:
        manifests:
          - contextDir: manifests
            sourcePath: ''
            uri: 'https://github.com/harshad16/odh-notebooks/archive/standardize-manifests.tar.gz'
```
and i get this issue on the DSC :

Screenshot 2025-05-09 at 16 41 30

ci/security-scan/weekly_commit_ids.env
```diff
@@ -1,30 +1,30 @@
 odh-minimal-notebook-image-main=quay.io/opendatahub/workbench-images@sha256:e9d6a6ee0e1ce6878d3a00dd9ffe85ffb536298ce1fe4d9c577ba0159c69a7f3
-odh-minimal-notebook-image-n=quay.io/opendatahub/workbench-images@sha256:cde20ac445d25c70d95042a546334c398ed3fca73e85530f0ffef3cbdb6ec746
-odh-minimal-notebook-image-n-1=quay.io/opendatahub/workbench-images@sha256:9323e689ec6ab1abb3cbdfd6258811bd57376c6b3e48f71838408cbb0b8b24a3
+odh-workbench-jupyter-minimal-cpu-py311-ubi9-n=quay.io/opendatahub/workbench-images@sha256:cde20ac445d25c70d95042a546334c398ed3fca73e85530f0ffef3cbdb6ec746
+odh-workbench-jupyter-minimal-cpu-py311-ubi9-n-1=quay.io/opendatahub/workbench-images@sha256:9323e689ec6ab1abb3cbdfd6258811bd57376c6b3e48f71838408cbb0b8b24a3
```
atheo89, line 3: As we don't support the n-1 should we include these on the weekly scan?

manifests/base/runtime-minimal-imagestream.yaml
harshad16: thanks for checking, i tried it too,
as you rightly pointed out, it doesnt work
researching on the fix, do share if there are any suggestions to fix this.

manifests/base/runtime-tensorflow-imagestream.yaml
jiridanek: sure, but it changes the value of `opendatahub.io/runtime-image-metadata` from string to an array, that's ok?

harshad16: you are absolutely right
i thought was alright, seems like operator does have problem with patching.
reverted it back.

ci/security-scan/weekly_commit_ids.env
harshad16: not relevant to this issue, lets track this separately and fix it later.

manifests/base/runtime-minimal-imagestream.yaml
harshad16: Removed the image_name from the string JSON, 
and adjusted the controller to take care of this case by override the url directly from imagestream url
https://github.com/opendatahub-io/kubeflow/pull/603

atheo89: I reviewed the other PR on nbc, it looks good, so If we go this way we need only a comment to inform that the image_url it gets populated by the nbc.  

Unfortunately, the Kustomize does not natively support parsing JSON strings within annotations, so direct variable substitution for image_name inside it doesn't work. There is a plugin that is called `transformers` where it can be used for this cases. 

like:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - runtime-minimal-imagestream.yaml
  - runtime-standard-imagestream.yaml
  - runtime-advanced-imagestream.yaml

transformers:
  - |-
    apiVersion: builtin
    kind: ExecTransformer
    metadata:
      name: image-name-transformer
    args:
      - ./image_name_transformer.py
    envs:
      - runtimes.env

configMapGenerator:
  - name: runtimes-parameters
    env: runtimes.env
generatorOptions:
  disableNameSuffixHash: true
```

The imageStream:
```
apiVersion: image.openshift.io/v1
kind: ImageStream
metadata:
  name: runtime-minimal
spec:
  tags:
    - name: minimal
      annotations:
        opendatahub.io/runtime-image-metadata: |
          [
            {
              "display_name": "Python 3.11 (UBI9)",
              "metadata": {
                "tags": ["minimal"],
                "display_name": "Python 3.11 (UBI9)",
                "image_name": "PLACEHOLDER",
                "pull_policy": "IfNotPresent"
              },
              "schema_name": "runtime-image"
            }
          ]
      from:
        kind: DockerImage
        name: PLACEHOLDER
      referencePolicy:
        type: Source
```

and the script that handles the translate:
```
#!/usr/bin/env python3
import sys
import yaml
import json
import os

IMAGE_MAP = {
    "runtime-minimal": "odh-runtime-minimal-cpu-py311-ubi9-n",
    "runtime-datascience": "odh-runtime-datascience-cpu-py311-ubi9-n",
}

def transform_image_name(obj):
    if obj.get("kind") != "ImageStream":
        return obj

    name = obj.get("metadata", {}).get("name")
    env_var = IMAGE_MAP.get(name)
    if not env_var:
        return obj  # Skip if no matching env var

    image_name = os.getenv(env_var)
    if not image_name:
        print(f"Warning: Environment variable '{env_var}' not found", file=sys.stderr)
        return obj

    tags = obj.get("spec", {}).get("tags", [])
    for tag in tags:
        annotations = tag.get("annotations", {})
        metadata_json = annotations.get("opendatahub.io/runtime-image-metadata")
        if metadata_json:
            try:
                metadata = json.loads(metadata_json)
                for entry in metadata:
                    if "metadata" in entry:
                        entry["metadata"]["image_name"] = image_name
                annotations["opendatahub.io/runtime-image-metadata"] = json.dumps(metadata, indent=2)
            except Exception as e:
                print(f"Error parsing JSON for {name}: {e}", file=sys.stderr)

        # Optional: update the tag `from.name` field too
        if "from" in tag:
            tag["from"]["name"] = image_name

    return obj

def main():
    resources = list(yaml.safe_load_all(sys.stdin))
    transformed = [transform_image_name(r) for r in resources]
    yaml.safe_dump_all(transformed, sys.stdout)

if __name__ == "__main__":
    main()

```

manifests/base/kustomization.yaml
```diff
@@ -671,7 +671,6 @@ replacements:
     targets:
       - fieldPaths:
           - spec.tags.0.from.name
-          - spec.tags.0.annotations.[opendatahub.io/runtime-image-metadata].0.metadata.image_name
```
jiridanek: * https://github.com/opendatahub-io/kubeflow/pull/603

yes, got it, we're getting quite far from the original intent/title of the PR, though

> Standardize the image name with image convention

edit: I now see why it's necessary to do it all in the same PR, so ok

jiridanek (commented):
Can you please add explanation into some readme or a comment on top of params.env about

* Change image name as per image name convention
* Split params.env into latest and previous

?

What are the image conventions, and why the env files have to be split?

# this file contains references to latest versions of workbench images that are  updated by konflux nudges. see params.env for the references to older versions

something like that

and

Image names follow the established RHOAI format odh-<component>...-ubi9. Or whatever the standard image name actually is.

jiridanek:
Is Konflux able to maintain the commit.envs, actually? I never saw it mentioned that nudges also update commit hashes!

ci/security-scan/weekly_commit_ids.env
atheo89: thnx @jiridanek for creating the tracker on odh,i will addit here for reference

manifests/base/params-latest.env
harshad16: Fixed this,please take a look.

manifests/base/runtime-minimal-imagestream.yaml
harshad16: Thank for sharing this, 
i kinda feel as having the change in JSON , still require NBC for full utilization in elyra.
I strongly feel, we should rely on NBC for updating image_name field based on imagestream image.
What do you say ?

manifests/base/kustomization.yaml
harshad16: I wasnt able to find a way to separate it out.
perhaps we can consider it for this one time.

harshad16:
Updated all the required changes
@atheo89 @jiridanek @jstourac please review

docs/workbenches-naming.md
```diff
@@ -0,0 +1,66 @@
...
+```
+odh-<image type>-<image-feature>-<image-scope>-<accelerator>-<python version>-<os-version>
+```
+
+- Image prefix: odh
+- Image type: workbench, pipeline
+- Image feature:
+  - jupyter
+  - codeserver
+  - Rstudio
```
jiridanek, line 16: lowercase R, so it matches examples below

manifests/base/kustomization.yaml
jiridanek: Its fine, the changes do belong together. And hopefully this PR will not drag too long, because there's actually urgency to getting all this sorted out, so it's not a problem 🙏🏽

jiridanek:
/lgtm

it seems to do the job, and solves the immediate problems

manifests/base/runtime-minimal-imagestream.yaml
atheo89: It is fine for me this way as well! Just posted my research on how we can do it explicitly on notebooks side.

.github/workflows/params-env.yaml
```diff
@@ -6,9 +6,11 @@ on:  # yamllint disable-line rule:truthy
     paths:
       - 'manifests/base/commit.env'
       - 'manifests/base/params.env'
+      - 'manifests/base/commit-latest.env'
+      - 'manifests/base/params-latest.env'
       - 'ci/check-params-env.sh'
-      - 'ci/check-runtime-images.sh'
-      - 'manifests/base/runtime-*-imagestream.yaml'
+      # - 'ci/check-runtime-images.sh'
```
jstourac, line 12: maybe we should delete this line completely now?

.github/workflows/params-env.yaml
```diff
@@ -6,9 +6,11 @@ on:  # yamllint disable-line rule:truthy
...
+      # - 'manifests/base/runtime-*-imagestream.yaml'
```
jstourac, line 13: hm, do we want this to be commented out? I think that it's still good to have it here or am I missing something? :thinking:

ci/check-params-env.sh
```diff
@@ -407,66 +460,75 @@ function check_image() {
...
-    }
-done < "${PARAMS_ENV_PATH}"
+        check_image "${IMAGE_VARIABLE}" "${IMAGE_URL}" || {
+            echo "ERROR: Image definition for '${IMAGE_VARIABLE}' isn't okay!"
+            echo "------------------------"
+            ret_code=1
+            continue
+        }
+    done < "${1}"
+}
```
jstourac, line 518: we should return value of the `ret_code`  variable before exiting the function; also, it would be better to initialize the `ret_code` variable to 0 at the beginning of the function, just for the clarity, wdyt?

ci/check-params-env.sh
```diff
@@ -407,66 +460,75 @@ function check_image() {
...
+
+# process_file "${PARAMS_ENV_PATH}" || {ret_code=1}
+# if test "${ret_code}" -eq 0; then
+#     echo "Validation of '${PARAMS_ENV_PATH}' was successful! Congrats :)"
+# else
+#     echo "The '${PARAMS_ENV_PATH}' file isn't valid, please check above!"
+# fi
```
jstourac, line 525: Is this commented out on purpose?

Maybe we should put this into a `for` cycle iteration over these two files to avoid code duplicity?

docs/workbenches-naming.md
```diff
@@ -0,0 +1,66 @@
...
+  - tag :v3-20250418-c0fa3b2
+- Image:
+  - name: runtime-images
+  - tag: rocm-runtime-tensorflow-ubi9-python-3.11-20250418-c0fa3b2
+
+## Split of params 
+
+Konflux would be help with build of workbenches. Konflux patch the sha of the built image into params file as a process of nudging.  
+To keep this file not getting overriden, we have split the params file into 2 files.
+- params-latest.env: Any workbench images that represent the nth image.
```
jstourac, line 63: Optional: the nth makes me think like it is 1, 2, 3, 4, 5...
```suggestion
- params-latest.env: Any workbench images that represent the image version N (the latest one).
```
but I don't insist

jstourac:
I put couple of comments, but in general LGTM.

I also checked the generated content like this:
```
git checkout main
./ci/kustomize.sh ; echo $?
git fetch harshad16 standardize-manifests
git checkout harshad16/standardize-manifests
./ci/kustomize.sh ; echo $?
meld /tmp/kustomize-uepU32pPb0/kustomize-5.0.3-stdout.yaml /tmp/kustomize-8FTQf41Vsc/kustomize-5.0.3-stdout.yaml
```
the diff seems reasonable and okay to me

.github/workflows/params-env.yaml
harshad16: as we moved the image quotient to params.env , realized running this workflow on change of runtime-imagestream.yaml should have no impact.

.github/workflows/params-env.yaml
harshad16: ack, i wanted too
however thought to confirm.
removing it.

.github/workflows/params-env.yaml
```diff
@@ -28,7 +30,8 @@ jobs:
...
 
-      - name: Validate references for runtime images
-        id: validate-runtime-images-references
-        run: |
-          bash ./ci/check-runtime-images.sh
+      # Validation of runtime images is included in the check-params-env.sh script
+      # - name: Validate references for runtime images
+      #   id: validate-runtime-images-references
+      #   run: |
+      #     bash ./ci/check-runtime-images.sh
```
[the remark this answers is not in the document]
harshad16: Yes, i believe so.
Removing it.

ci/check-params-env.sh
harshad16: Include the ret_code, as the ret_code was global var, 
didnt  return it from func, used it as it.
as the value gets updated one after another.

ci/check-params-env.sh
harshad16: Yes, made the change.

.github/workflows/params-env.yaml
jstourac: ah, right, we probably don't care about anything else there... I see, you are right, thanks!

ci/check-params-env.sh
jstourac: Hm, okay. If we returned the `ret_code` value, then we could use e.g. this:
```if process_file "${file_}"; then
```

ci/check-params-env.sh
```diff
@@ -407,66 +460,74 @@ function check_image() {
...
+    done < "${1}"
+}
 
-    check_image "${IMAGE_VARIABLE}" "${IMAGE_URL}" || {
-        echo "ERROR: Image definition for '${IMAGE_VARIABLE}' isn't okay!"
+for file_ in  "${PARAMS_ENV_PATH}" "${PARAMS_LATEST_ENV_PATH}"; do
+    echo "Checking file: '${file_}'"
+    process_file "${file_}"
+    if test "${ret_code}" -eq 0; then
+        echo "Validation of '${PARAMS_LATEST_ENV_PATH}' was successful! Congrats :)"
```
jstourac, line 525:
```suggestion
        echo "Validation of '${file_}' was successful! Congrats :)"
```

ci/check-params-env.sh
```diff
@@ -407,66 +460,74 @@ function check_image() {
...
-else
-    echo "The '${PARAMS_ENV_PATH}' file isn't valid, please check above!"
-fi
+    else
+        echo "The '${PARAMS_LATEST_ENV_PATH}' file isn't valid, please check above!"
+        echo "------------------------"
+    fi
+done
 
 exit "${ret_code}"
```
jstourac, line 533: we need to be sure that `ret_code` of the second file processing doesn't override the eventual error in the first file processing - so maybe we should store this value

ci/check-params-env.sh
harshad16: Absolutely, i have included the new changes

jstourac: I would use `local ret_code=0` to keep the variable just for the local scope.

Anyway, I'm not gonna block this anymore. Thank you for your changes, Harshad!

harshad16:
Thanks @jiridanek @jstourac @atheo89 for the review :100: 
/approve

harshad16:
Merging to move forward

jiridanek:
@coderabbitai are you sure this is not introducing any bugs you may've missed?

jiridanek:
@coderabbitai filled

* https://github.com/opendatahub-io/notebooks/issues/1786

we will fix it eventually

jiridanek:
@coderabbitai ofc. it is introduced in #1070, see change on line 66 in ci/check-params-env.sh

https://github.com/opendatahub-io/notebooks/pull/1070/files#diff-2617249dc2cd2c36044c0e68dd1cb60324b8338863a75660ebd3586bbfc67c39L66

PR #1084, opened 2025-05-13
NO-JIRA: tests(gha): fix `check-payload` to work with PRs and without `podman login` for root user

jiridanek:
## Description

Turns out, there were two problems with previous implementation, that were only revealed when

* the action was running on a push to a repository that has private github container registry
* the action was running on a pull request that modified some image

In the first case, action was unable to pull from repository that requires authentication.

In the second case, the image is not pushed anywhere, so attempts to pull are futile.

The solution appears to be what was implemented in

* https://github.com/openshift/check-payload/pull/154

That way we can run against a locally present image.

## How Has This Been Tested?

* The GHAs that runs on the PR, note that PR branch is in this repo, not on my fork, that is intentional

my fork

* https://github.com/jiridanek/notebooks/actions/runs/14992177148

tried running without the valgrind ignores to check that it will be picked up

* https://github.com/jiridanek/notebooks/actions/runs/14992549084 (failed as intended)

## Merge criteria:

### DIFF
--- a/scripts/check-payload/config.toml
+++ b/scripts/check-payload/config.toml
@@ -234,7 +234,15 @@ files = [
 
 # https://issues.redhat.com/browse/RHOAIENG-24290
 
+# when scanning with `scan image --spec`
 [[payload.python-311-container.ignore]]
+error = "ErrNotDynLinked"
+files = [
+    # executable is not dynamically linked
+    "/opt/app-root/bin/py-spy",
+]
+# when scanning with `scan local --path`, the above does not apply
+[[ignore]]
 error = "ErrNotDynLinked"
 files = [
     # executable is not dynamically linked
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -561,7 +561,7 @@ jobs:
 
       # endregion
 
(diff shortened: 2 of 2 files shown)


jiridanek:
/override ci/prow/images
not applicable

jiridanek:
/override "Code static analysis / check-generated-code (pull_request)"
that's due to konflux changes sooner today

scripts/check-payload/config.toml
```diff
@@ -234,12 +234,20 @@ files = [
...
 # https://issues.redhat.com/browse/RHOAIENG-24290
 
+# when scanning with `scan image --spec`
 [[payload.python-311-container.ignore]]
 error = "ErrNotDynLinked"
 files = [
     # executable is not dynamically linked
     "/opt/app-root/bin/py-spy",
 ]
+# when scanning with `scan local --path`, the above does not apply
```
jstourac, line 244: So - is the above still needed since we're not running it that way anymore?

jstourac: Okay then, I wouldn't know how to run it locally myself - maybe we could have some target in the Makefile? anyway, that would be a different PR.

jiridanek: * https://github.com/opendatahub-io/notebooks/issues/1086

PR #1087, opened 2025-05-13
Build mongocli instead of downloading it pre-built

grdryn:
* https://issues.redhat.com/browse/RHOAIENG-24348
* https://issues.redhat.com/browse/RHOAIENG-24291
* https://issues.redhat.com/browse/RHOAIENG-24340
* https://issues.redhat.com/browse/RHOAIENG-24344
* https://issues.redhat.com/browse/RHOAIENG-24343

### DIFF
--- a/jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
+++ b/jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
@@ -1,3 +1,16 @@
+######################################################
+# mongocli-builder (build stage only, not published) #
+######################################################
+FROM registry.access.redhat.com/ubi9/go-toolset:latest AS mongocli-builder
+
+ARG MONGOCLI_VERSION=2.0.3
+
+WORKDIR /tmp/
+RUN curl -Lo mongodb-cli-mongocli-v${MONGOCLI_VERSION}.zip https://github.com/mongodb/mongodb-cli/archive/refs/tags/mongocli/v${MONGOCLI_VERSION}.zip
+RUN unzip ./mongodb-cli-mongocli-v${MONGOCLI_VERSION}.zip
+RUN cd ./mongodb-cli-mongocli-v${MONGOCLI_VERSION}/ && \
+    CGO_ENABLED=1 GOOS=linux go build -a -tags strictfipsruntime -o /tmp/mongocli ./cmd/mongocli/
+
 ########################
 # base                 #
 ########################
@@ -66,10 +79,8 @@ USER root
 # Install useful OS packages
 RUN dnf install -y jq unixODBC postgresql git-lfs libsndfile && dnf clean all && rm -rf /var/cache/yum
 
-# Install MongoDB Client, We need a special repo for MongoDB as they do their own distribution
-COPY ${DATASCIENCE_SOURCE_CODE}/mongodb-org-6.0.repo-x86_64 /etc/yum.repos.d/mongodb-org-6.0.repo
-
-RUN dnf install -y mongocli && dnf clean all && rm -rf /var/cache/yum
+# Copy dynamically-linked mongocli built in earlier build stage
+COPY --from=mongocli-builder /tmp/mongocli /opt/app-root/bin/
 
 # Install MSSQL Client, We need a special repo for MSSQL as they do their own distribution
 COPY ${DATASCIENCE_SOURCE_CODE}/mssql-2022.repo-x86_64 /etc/yum.repos.d/mssql-2022.repo
@@ -95,11 +106,11 @@ RUN echo "Installing softwares and packages" && \
     # Remove default Elyra runtime-images \
     rm /opt/app-root/share/jupyter/metadata/runtime-images/*.json && \
     # Replace Notebook's launcher, "(ipykernel)" with Python's version 3.x.y \
-    sed -i -e "s/Python.*/$(python --version | cut -d '.' -f-2)\",/" /opt/app-root/share/jupyter/kernels/python3/kernel.json && \    
+    sed -i -e "s/Python.*/$(python --version | cut -d '.' -f-2)\",/" /opt/app-root/share/jupyter/kernels/python3/kernel.json && \
     # Disable announcement plugin of jupyterlab \
-    jupyter labextension disable "@jupyterlab/apputils-extension:announcements" && \    
+    jupyter labextension disable "@jupyterlab/apputils-extension:announcements" && \
     # Apply JupyterLab addons \
-    /opt/app-root/bin/utils/addons/apply.sh && \    
+    /opt/app-root/bin/utils/addons/apply.sh && \
     # Fix permissions to support pip in Openshift environments \
     chmod -R g+w /opt/app-root/lib/python3.11/site-packages && \
     fix-permissions /opt/app-root -P
--- a/scripts/check-payload/config.toml
+++ b/scripts/check-payload/config.toml
@@ -216,6 +216,8 @@ files = [
     "/usr/lib/rstudio-server/bin/quarto/bin/tools/x86_64/esbuild",
 ]
 
+# valgrind is waived since it doesn't do anything with crypto in a
+# security context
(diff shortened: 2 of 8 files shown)


jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -1,3 +1,16 @@
+######################################################
+# mongocli-builder (build stage only, not published) #
+######################################################
+FROM registry.access.redhat.com/ubi9/go-toolset:latest AS mongocli-builder
+
+ARG MONGOCLI_VERSION=2.0.3
```
grdryn, line 6: Not sure where best to store the mongocli version. Maybe the ARG shouldn't have a value set here in the actual Dockerfiles, and be set as part of the build commands? I presume that can be set in Makefiles and/or tekton pipelinerun files as appropriate, but not 100% sure where, so LMK what you think!

jiridanek:
What's causing this build failure?

> [1/4] STEP 7/7: RUN CGO_ENABLED=1 GOOS=linux go build -a -tags strictfipsruntime -o /tmp/mongocli ./cmd/mongocli/
go: go.mod file not found in current directory or any parent directory; see 'go help modules'
Error: building at STEP "RUN CGO_ENABLED=1 GOOS=linux go build -a -tags strictfipsruntime -o /tmp/mongocli ./cmd/mongocli/": while running runtime: exit status 1

jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -1,3 +1,16 @@
...
+
+WORKDIR /tmp/
+RUN curl -Lo mongodb-cli-mongocli-v${MONGOCLI_VERSION}.zip https://github.com/mongodb/mongodb-cli/archive/refs/tags/mongocli/v${MONGOCLI_VERSION}.zip
+RUN unzip ./mongodb-cli-mongocli-v${MONGOCLI_VERSION}.zip
+RUN cd ./mongodb-cli-mongocli-v${MONGOCLI_VERSION}/
+RUN CGO_ENABLED=1 GOOS=linux go build -a -tags strictfipsruntime -o /tmp/mongocli ./cmd/mongocli/
```
jctanner, line 12: This RUN directive will be run within `/` because the previous directive to `cd ./mongodb-cli-mongocli-v${MONGOCLI_VERSION}/` does not set the shell context for subsequent RUN directives.

Example:
```bash
jtanner@corsair:/tmp/deleteme$ cat Dockerfile 
FROM registry.access.redhat.com/ubi9/ubi:latest

RUN mkdir /foo
RUN cd /foo
RUN pwd
jtanner@corsair:/tmp/deleteme$ DOCKER_BUILDKIT=0 docker build --no-cache -t test:latest .
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.

Sending build context to Docker daemon  2.048kB
Step 1/4 : FROM registry.access.redhat.com/ubi9/ubi:latest
 ---> d53e845ae020
Step 2/4 : RUN mkdir /foo
 ---> Running in c5892a0d4619
 ---> Removed intermediate container c5892a0d4619
 ---> 9aa6b7c1aeca
Step 3/4 : RUN cd /foo
 ---> Running in b67ec4ac8cf4
 ---> Removed intermediate container b67ec4ac8cf4
 ---> 1ca53560efb8
Step 4/4 : RUN pwd
 ---> Running in 282ba45f86df
/
 ---> Removed intermediate container 282ba45f86df
 ---> d9869df57513
Successfully built d9869df57513
Successfully tagged test:latest
```

jctanner (changes requested):

jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
grdryn: Fixed, thanks :+1: I realised this when I saw @jiridanek's comment earlier, but was afk so couldn't fix until now :slightly_smiling_face:

jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
jiridanek: /lgtm

I don't really have an answer to this. `.tekton/` pipelines are undergoing overhaul currently, to be maintained by devtestops afterwards, so any change going forward would require raising jira ticket to them; that's impractical; cc @MohammadiIram , https://redhat-internal.slack.com/archives/C05TTTYG599/p1747115889072959?thread_ts=1747115307.125289&cid=C05TTTYG599

scripts/check-payload/config.toml
```diff
@@ -255,33 +255,3 @@ files = [
     # executable is not dynamically linked
     "/usr/lib/code-server/lib/vscode/node_modules/@vscode/ripgrep/bin/rg",
 ]
-
-# https://issues.redhat.com/browse/RHOAIENG-24340
```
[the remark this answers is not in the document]
grdryn: It was already listed here, but I've updated the comments to explicitly say that it's to be waived

jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
grdryn: I'm going to resolve this since it's an existing pattern in these Dockerfiles to have "version" ARGs specified with a value directly in the Dockerfile, so it's not something that is specific to this PR. See for example `CODESERVER_VERSION`, `ROCM_VERSION`, `CUDA_VERSION`, etc.

jstourac:
@grdryn the changes LGTM in general. Thanks for this change!

/lgtm

We should go and do some check that the mongocli binary works as expected for us. AFAIK we don't have any automated test as of today.

jupyter/datascience/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -1,3 +1,16 @@
...
+######################################################
+FROM registry.access.redhat.com/ubi9/go-toolset:latest AS mongocli-builder
+
+ARG MONGOCLI_VERSION=2.0.3
+
+WORKDIR /tmp/
+RUN curl -Lo mongodb-cli-mongocli-v${MONGOCLI_VERSION}.zip https://github.com/mongodb/mongodb-cli/archive/refs/tags/mongocli/v${MONGOCLI_VERSION}.zip
+RUN unzip ./mongodb-cli-mongocli-v${MONGOCLI_VERSION}.zip
+RUN cd ./mongodb-cli-mongocli-v${MONGOCLI_VERSION}/ && \
+    CGO_ENABLED=1 GOOS=linux go build -a -tags strictfipsruntime -o /tmp/mongocli ./cmd/mongocli/
```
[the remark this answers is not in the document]
jstourac: Hm, checked locally myself, the way I proposed above fails, would need to check what is wrong. Can get back to it once I have some time.

Anyway, I don't think this should block this. Though, it would be great if we could have the version build into the tool for anyone to check online on the running image.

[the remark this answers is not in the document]
grdryn: Good idea. I've figured it out, those flags need to be provided as part of `-ldflags` to `go build`. I'll try to update this now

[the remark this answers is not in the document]
grdryn: Actually, scratch that: while I can get my local Go to build it in a way that it'll make the `mongocli --version` output reasonable information, the RHEL/UBI9 Go doesn't even with the same command used. Let's go with your original thinking:

> Anyway, I don't think this should block this.

jiridanek:
> We should go and do some check that the mongocli binary works as expected for us. AFAIK we don't have any automated test as of today.

That should not be too hard, good luck to you with it.

Let's merge this and add tests separately.

/approve

PR #1085, opened 2025-05-13
Update references to kubeflow-training to match up to date version

Fiona-Waters:
The kubeflow-training version included in notebooks was not up to date, therefore I am updating it.

## Description

Updated references in 2025.1 for kubeflow-training to 1.9.2

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #1095, opened 2025-05-14
RHOAIENG-24981: Updated codeflare sdk versions and shas for v0.28.1

kryanbeane:
## Merge criteria:


kryanbeane:
fixing conflicts

kryanbeane:
@atheo89 thanks for the pointers, should be all up to date now!

manifests/base/jupyter-datascience-notebook-imagestream.yaml
```diff
@@ -36,7 +36,7 @@ spec:
             {"name": "Odh-Elyra", "version": "4.2"},
             {"name": "PyMongo", "version": "4.11"},
             {"name": "Pyodbc", "version": "5.2"},
-            {"name": "Codeflare-SDK", "version": "0.27"},
+            {"name": "Codeflare-SDK", "version": "0.28.1"},
```
jstourac, line 39: Something that was discussed yesterday in https://github.com/opendatahub-io/notebooks/pull/1085#issuecomment-2875933849. So we probably want to have `0.28` only there.

kryanbeane: Would this pick up the latest v0.28.1 if we just have 0.28? We need to release 0.28.1 as opposed to 0.28 as 0.28.0 has a bug

atheo89: This is only for presentation reasons on the UI. The image itself will include the `0.28.1` as it is defined on the pipfiles

kryanbeane: Ah I understand. Will update, ty!

atheo89:
> @atheo89 thanks for the pointers, should be all up to date now!

Great! Thank you for the fast update 🙂 
What about the tensorflow runtime? Is the only missing part. Does it have conflicts with the other packages?

kryanbeane:
> > @atheo89 thanks for the pointers, should be all up to date now!
> 
> Great! Thank you for the fast update 🙂 What about the tensorflow runtime? Is the only missing part. Does it have conflicts with the other packages?

Oops missed that, I was searching for `0.27.0` to find what to update. Weird that the tensor flow runtime wasn't updated in the past. Shouldn't be any conflicts with other packages, I will double check with the team and if not I will update it

manifests/base/jupyter-datascience-notebook-imagestream.yaml
kryanbeane: resolved in fc6c6dc

kryanbeane:
hey @atheo89 two question: 
- is there an easy way to run `pipenv lock` or `poetry lock` on a mac but for CUDA runtime images? I am trying to lock dependencies for the pytorch runtime image and only way I could get it going was to start a docker container and run it from inside that but had to mess with the torchvision package (change 0.21.0+cu126 to 0.21.0) to get it working which I'd imagine isnt recommended. Any way you guys do it?
- It turns out the tensorflow one wasn't updated because of an issue with tf2onnx package dependency resolution (see [here](https://github.com/project-codeflare/codeflare-sdk/blame/main/.github/workflows/odh-notebooks-sync.yml#L100C90-L100C127)). It looks like a workaround was added a while back (see [here](https://github.com/opendatahub-io/notebooks/blob/669be3c3d7f131a39147e8ece15b0e5619012c87/jupyter/tensorflow/ubi9-python-3.11/Pipfile#L7-L9)) so just looking for some confirmation that it should be okay to update codeflare sdk despite that? 

Thanks!

.github/workflows/params-env.yaml
```diff
@@ -27,4 +27,3 @@ jobs:
       - name: Validate the 'manifests/base/params.env' file content
         run: |
           bash ./ci/check-params-env.sh
-
```
kryanbeane, line 30: thanks for lgtm @jstourac, last update. Format check was failing because of an extra blank line🙄 There is one failing check that fails as it doesn't allow my user to run the check, which should be overriden I assume

PR #1132, opened 2025-05-25
RHOAISTRAT-169: add conditional package install in minimal Dockerfile.cpu for s390x

Nash-123:
https://issues.redhat.com/browse/RHOAISTRAT-169

This PR enhances the `Dockerfile.cpu` for the minimal runtime to conditionally install additional OS packages when building on the `s390x` architecture.

This python package `pyzmq` require native compilation and thus depend on development tools and libraries such as `gcc`, `g++`, `make`, `openssl-devel`, `autoconf`, `automake`, `libtool`, and `cmake`. These are typically not needed on x86_64 but are essential for s390x.

### Changes
- Added an `ARCH=$(uname -m)` check to detect the architecture.
- Installed the additional packages only when `ARCH == "s390x"`.

### Impact
This change ensures:
- notebook runtime minimal builds are successfull on IBM Z (s390x).
- No impact to x86_64 builds.

Tested on: s390x 

c.c: @modassarrana89 @vibhaKulka

## Summary by CodeRabbit

- **Chores**
	- Improved compatibility for s390x architecture by conditionally installing additional development tools during the build process. No changes for other architectures.

### DIFF
--- a/runtimes/minimal/ubi9-python-3.11/Dockerfile.cpu
+++ b/runtimes/minimal/ubi9-python-3.11/Dockerfile.cpu
@@ -9,7 +9,14 @@ WORKDIR /opt/app-root/bin
 USER 0
 
 # Install useful OS packages
-RUN dnf install -y mesa-libGL skopeo && dnf clean all && rm -rf /var/cache/yum
+RUN ARCH=$(uname -m) && \
+    echo "Detected architecture: $ARCH" && \
+    if [ "$ARCH" = "s390x" ]; then \
+        dnf install -y mesa-libGL skopeo gcc g++ make openssl-devel autoconf automake libtool cmake; \
+    else \
+        dnf install -y mesa-libGL skopeo; \
+    fi && \
+    dnf clean all && rm -rf /var/cache/yum
 
 # Other apps and tools installed as default user
 USER 1001


runtimes/minimal/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -9,7 +9,14 @@ WORKDIR /opt/app-root/bin
 USER 0
 
 # Install useful OS packages
-RUN dnf install -y mesa-libGL skopeo && dnf clean all && rm -rf /var/cache/yum
+RUN ARCH=$(uname -m) && \
+    echo "Detected architecture: $ARCH" && \
+    if [ "$ARCH" = "s390x" ]; then \
+        dnf install -y mesa-libGL skopeo gcc g++ make openssl-devel autoconf automake libtool cmake; \
```
jstourac, line 15: I'm thinking whether doing the build of `pyzmq` in a separate build container and only then copying the result in our minimal image wouldn't be a slightly cleaner solution. Something like we are doing for `mongocli` in #1087.

But let's get opinion from others too first.

jiridanek: It would be much cleaner, but also much work, sadly, if we wanted to make everything clean and tidy.

There are two options, either build the wheels in builder image and then install them in the output image, or install in the builder image and then copy over the python venv from the builder image. The first option is cleaner, but more work.

Going forward, there's also extra work ensuring that all necessary runtime deps are installed in the output image; but hopefully they are already installed because the x86_64 image hopefully worked, so maybe this is not really an issue.

@jstourac I suggest we create a jira ticket and explore going the separate builder image route first. We'd need to test on z390x but hopefully the qemu-user binary emulation will cover it.

Nash-123: Thanks @jstourac  and @jiridanek  for the comments!

I believe using QEMU binary emulation to build for s390x should work, but it will consume a lot of time to get build.
Given that, I'd like to confirm with you both, which approach would you recommend I proceed with for now, considering this is an initial enablement and we'd still want to keep things maintainable?

Thanks again!

jiridanek: @adrielparedes ^^^

@jstourac Something Matej said today got me thinking. Maybe on the z390x images we should have build tools preinstalled by default. The availability of prebuilt wheels for the less popular architectures is poorer, therefore our users may need build tools around so that they can build what they need as they go. The infamous
```
!pip install
```

JupyterLab notebook cells, is what I have in mind.

jiridanek: > Is there some tracking Jira issue for this change?
Here, https://issues.redhat.com/browse/RHOAISTRAT-169. It even has IDE component. LOL.

runtimes/minimal/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -9,7 +9,14 @@ WORKDIR /opt/app-root/bin
 USER 0
 
 # Install useful OS packages
-RUN dnf install -y mesa-libGL skopeo && dnf clean all && rm -rf /var/cache/yum
+RUN ARCH=$(uname -m) && \
+    echo "Detected architecture: $ARCH" && \
```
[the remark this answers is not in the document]
jiridanek: @coderabbitai please create a github issue for this, we will initially merge without the fix

jiridanek (approved):

runtimes/minimal/ubi9-python-3.11/Dockerfile.cpu
jiridanek: @coderabbitai create two issues for the suggestions from comment above: explore going the separate builder image route first. I hope it will not be too much work. We'd need to test on z390x but hopefully the qemu-user binary emulation will cover it.

jiridanek:
@coderabbitai create issue for the nitpick suggestion 14-18: Refactor conditional package lists to reduce duplication

PR #1258, opened 2025-07-02
RHOAIENG-28774: allow ODH codeserver image to be built for arm64

grdryn:
## Description

JIRA: https://issues.redhat.com/browse/RHOAIENG-28774

## How Has This Been Tested?

This has only been tested by verifying that the image can be built for both amd64 and arm64. The changes here shouldn't have any impact on the existing amd64 image.

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Improved build process to support multiple architectures when building the image.

### DIFF
--- a/codeserver/ubi9-python-3.11/Dockerfile.cpu
+++ b/codeserver/ubi9-python-3.11/Dockerfile.cpu
@@ -29,6 +29,8 @@ RUN curl -L https://mirror.openshift.com/pub/openshift-v4/$(uname -m)/clients/oc
 ####################
 FROM base AS codeserver
 
+ARG TARGETOS TARGETARCH
+
 ARG CODESERVER_SOURCE_CODE=codeserver/ubi9-python-3.11
 ARG CODESERVER_VERSION=v4.98.0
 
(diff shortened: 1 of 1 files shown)


jiridanek:
@coderabbitai full review

jiridanek:
@coderabbitai full

codeserver/ubi9-python-3.11/Dockerfile.cpu
```diff
@@ -29,6 +29,8 @@ RUN curl -L https://mirror.openshift.com/pub/openshift-v4/$(uname -m)/clients/oc
 ####################
 FROM base AS codeserver
 
+ARG TARGETOS TARGETARCH
```
atheo89, line 32: Thanks, Gerard, for this update! The changes look good to me as well.

That said, since we're moving our Dockerfiles toward multi-architecture support, I think it would be valuable to make their structure a bit more robust—**not necessarily in this PR, but as part of future enhancements**.

More specifically: 
We should consider defining `ARG TARGETOS TARGETARCH` in the `base stage` as well, to keep the architecture handling consistent across layers. Also, it would be more reliable to replace `$(uname -m)` with `${TARGETARCH}`. The reason is that `uname -m` reflects the architecture of the build machine, not the target. If you're building for a different architecture (e.g., `TARGETARCH=arm64` on an `amd64` host), this could result in pulling or using the wrong binary.

It might also be a good idea to add label annotations that explicitly indicate the target architecture and OS of the built image for clarity and compliance with OCI image metadata.

Let’s ask @coderabbitai to open a new issue to track these suggestions for future improvement.

/lgtm
/approve

PR #1493, opened 2025-07-23
Install torch CPU wheel rather than CUDA one in Trusty AI notebooks

grdryn:
## Description

The Trusty AI notebooks seem to be intended for CPU usage rather than CUDA usage, and they don't have the CUDA RPMs installed like the other CUDA workbenches, so I'm wondering if the installation of the CUDA version of torch is intentional or not.

Using the CPU wheel makes it possible to also build this image for aarch64, where the CUDA wheel causes problems because of some missing aarch64 versions of some of the nvidia deps.

## How Has This Been Tested?

I just built it locally and ensured that the container process started as expected when run. I didn't attempt to run a relevant notebook or anything like that.

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Updated dependencies to use CPU-only versions of PyTorch, removing all GPU (CUDA) related packages and repositories.
  * Simplified and reduced package hashes for certain dependencies.
  * Removed GPU-specific libraries to improve compatibility for CPU-based environments.
  * Upgraded several packages including grpcio, comm, huggingface-hub, and narwhals for improved stability and features.


harshad16:
The reason to have CUDA toolkit in pytorch image container, is to support users if they want to install and use packages, that maybe require CUDA support then it can really on cuda toolkit installed, as other packages cant rely on torch install cuda libraries out of box.

grdryn:
Thanks for the clarifications folks, I'll close this now then.

PR #2558, opened 2025-10-07
Use triton from pytorch in trustyai image

grdryn:
## Description

JIRA: https://issues.redhat.com/browse/RHOAIENG-34086

Triton here is a dependency of torch, and is just being explicitly re-declared so that the index to pull it from can be customized. Although torch was being pulled from the PyTorch index, all of its dependencies are still defaulting to coming from PyPI (or whatever default index is used for a build). In the triton case, PyPI only has x86_64 wheels for this version, but the PyTorch index has both x86_64 and aarch64 wheels, matching the arches that are available for torch.

Since it's a dependency of torch, I've also added same platform exclusion marker to omit for ppc64le.

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

- **Bug Fixes**
  - Broadened Triton compatibility to apply on more platforms (excluding ppc64le), improving installation reliability.

- **Chores**
  - Declared Triton explicitly as a dependency and aligned its package source with the PyTorch CUDA index.
  - Added arm64 (linux-m2xlarge/arm64) to the multi-arch build platforms for CI builds.

### DIFF
--- a/.tekton/odh-workbench-jupyter-trustyai-cpu-py312-ubi9-pull-request.yaml
+++ b/.tekton/odh-workbench-jupyter-trustyai-cpu-py312-ubi9-pull-request.yaml
@@ -36,6 +36,7 @@ spec:
   - name: build-platforms
     value:
     - linux/x86_64
+    - linux-m2xlarge/arm64
     - linux/ppc64le
   - name: dockerfile
     value: jupyter/trustyai/ubi9-python-3.12/Dockerfile.cpu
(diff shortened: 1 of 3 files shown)


grdryn:
> On upstream [-pull-request.yaml](https://github.com/opendatahub-io/notebooks/blob/main/.tekton/odh-workbench-jupyter-trustyai-cpu-py312-ubi9-pull-request.yaml) builds we need to add arm as well on the build platform.

Done, thanks!

PR #2768, opened 2025-12-12
NO-JIRA: chore(scripts/pylocks_generator.sh): regenerate locks with `--universal` option enabled

jiridanek:
This is due to changes in latest uv

* https://github.com/astral-sh/uv/pull/16956

that have caused rhds builds to fail with

```
+ uv pip install --strict --no-deps --no-cache --no-config --no-progress --verify-hashes --compile-bytecode --index-strategy=unsafe-best-match --requirements=./pylock.toml
warning: The `--pylock` option is experimental and may change without warning. Pass `--preview-features pylock` to disable this warning.
Using Python 3.12.9 environment at: /opt/app-root
error: Package `ray` can't be installed because it doesn't have a source distribution or wheel for the current platform
hint: You're on Linux (`manylinux_2_34_aarch64`), but `ray` (v2.47.1) only has wheels for the following platform: `manylinux2014_x86_64`; consider adding "sys_platform == 'linux' and platform_machine == 'aarch64'" to `tool.uv.required-environments` to ensure uv resolves to a version with compatible wheels
```

and other such fun messages when only x86-64 gets resolved.

## Description

Documentation at https://docs.astral.sh/uv/reference/cli/#uv-pip-compile--python-platform says that `--python-platform linux` is alias for `x86_64-unknown-linux-gnu`; we cannot use this to get a multiarch pylock.

So even though this used to work for us well in the past, it was not the right thing to do.

See also `--universal` discussion with @grdryn at https://redhat-internal.slack.com/archives/C0961HQ858Q/p1757935641975969?thread_ts=1757542802.032519&cid=C0961HQ858Q

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Updated Python dependency lock file generation configuration to support universal multi-architecture builds instead of platform-specific variants
  * Modified lock file generation pipeline to use universal platform specifications for broader compatibility
  * Added environment markers for setuptools and wheel packages to ensure correct dependency resolution on different Python implementations


atheo89 (approved):
Good catch! 🎉 

/lgtm
/approve

PR #2790, opened 2025-12-19
NO-JIRA: chore(workflows): refactor cache handling logic and consolidate make targets in build workflows

jiridanek:
## Description

Successfully unified the two make target steps into a single step in [`.github/workflows/build-notebooks-TEMPLATE.yaml`](.github/workflows/build-notebooks-TEMPLATE.yaml):

**Changes made:**
- Combined `push-make-target` and `pr-make-target` steps into unified `make-target` step
- Eliminated duplicate code (disk monitoring background process and make command)
- Used conditional expression for `PUSH_IMAGES` environment variable: sets to "no" for pull request events, "yes" otherwise
- Updated Playwright test step to reference the new `make-target` step ID
- Removed redundant if conditions - step now runs for all event types

**Benefits:**
- Reduced code duplication by ~15 lines
- Easier to maintain - single source of truth for the build step
- Clearer logic with conditional environment variable

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Unified container build workflow into a single build step for make targets.
  * Improved cache strategy that varies behavior by workflow event (schedule, push, pull_request, workflow_dispatch).
  * Updated image push policy to disable pushes for pull request events and enable otherwise.
  * Adjusted test gating so end-to-end tests run based on the new build step outcome.

### DIFF
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -192,50 +192,59 @@ jobs:
           fi
           echo "${{ secrets.AIPCC_QUAY_BOT_PASSWORD }}" | podman login quay.io/aipcc -u "${{ secrets.AIPCC_QUAY_BOT_USERNAME }}" --password-stdin
 
-      - name: Compute extra podman build args
-        id: extra-podman-build-args
-        run: |
-          set -Eeuxo pipefail
-
-          EXTRA_PODMAN_BUILD_ARGS=""
-          if [[ "${{ inputs.platform }}" == "linux/s390x" ]]; then
-            # workaround for known issue https://github.com/zeromq/libzmq/pull/4486
-            # In qemu-user, CACHELINE_SIZE probe is undefined
-            # Compiling pyzmq through UV needs this in CFLAGS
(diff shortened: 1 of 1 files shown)


ide-developer (approved):

.github/workflows/build-notebooks-TEMPLATE.yaml
jiridanek: @coderabbitai create a github issue for this problem; my suggestion there is to store the command options in a python list, that way we avoid problems with spaces when concatting; using shlex.quote can also help

🧹 Nitpick comments (1)
.github/workflows/build-notebooks-TEMPLATE.yaml (1)
198-234: Python script uses type annotation but the type is enforced only at static analysis time.

The Literal type annotation on line 215 provides documentation but won't catch unexpected event names at runtime since Python doesn't enforce type hints. The else branch with ValueError handles this correctly. However, consider a minor improvement for robustness.

Also, on line 234, there's a potential issue: if extra_podman_build_args is empty, you'll have a leading space in CONTAINER_BUILD_CACHE_ARGS. This is unlikely to cause problems with shell parsing but is worth noting.

🔎 Optional: Avoid leading space when extra_podman_build_args is empty
-          with open(os.environ["GITHUB_OUTPUT"], "a") as f:
-             f.write(f"CONTAINER_BUILD_CACHE_ARGS={extra_podman_build_args} {cache_args}\n")
+          with open(os.environ["GITHUB_OUTPUT"], "a") as f:
+             args = f"{extra_podman_build_args} {cache_args}".strip()
+             f.write(f"CONTAINER_BUILD_CACHE_ARGS={args}\n")

PR #2846, opened 2026-01-20
chore(makefile): improve macOS compatibility for `.RECIPEPREFIX`

jiridanek:
## Description

Added automatic fallback to `gmake` on macOS if the default `make` version does not support `.RECIPEPREFIX`. Simplified error handling and eliminated redundancy.

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Improved build system compatibility across macOS and systems without certain GNU Make versions
  * Enhanced build reliability with stricter shell execution standards and improved environment detection

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,13 @@
+# On macOS, the default /usr/bin/make is GNU Make 3.81 which doesn't support .RECIPEPREFIX.
+# Transparently re-exec with gmake if available.
+ifeq ($(origin .RECIPEPREFIX), undefined)
+	ifeq ($(shell command -v gmake 2>/dev/null),)
+		$(error This Make does not support .RECIPEPREFIX. If on macOS, install GNU Make 4.0+ (`gmake`) via 'brew install make')
+	endif
+.DEFAULT:
+	@exec gmake $(MAKECMDGOALS)
+endif
+
 # https://stackoverflow.com/questions/18136918/how-to-get-current-relative-directory-of-your-makefile
 ROOT_DIR := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
 SELF ::= $(firstword $(MAKEFILE_LIST))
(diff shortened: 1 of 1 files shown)


jiridanek:
This is an excellent and highly detailed breakdown of the "transparent re-exec" problem in GNU Make. You have correctly identified that the stateless nature of Make, combined with its two-phase operation (parsing vs. execution) and Unix process hierarchy, makes this seemingly simple task surprisingly difficult.

Here is a visual analysis of the core failure mode you encountered, along with a confirmation of your conclusion.

### Visualizing the Core Failure (Attempt 4 & 6)

The most subtle and "promising" failure you described (Attempt 4) failed because of the Unix process model. When Make runs a recipe, it doesn't run it inside the Make process itself; it forks a child process (the shell).

When you run `exec gmake` inside a recipe:

1. **Parent:** The original `make` (3.81) waits for the child to finish.
2. **Child:** The subshell is *replaced* by the new `gmake` process via `exec`.
3. **Result:** The new `gmake` runs, finishes, and exits.
4. **Return:** The original `make` sees the child exit successfully (exit code 0) and thinks "Recipe finished." It then proceeds to run the *next* line of its own internal logic, often resulting in the duplicate execution you observed.

### Critique of Solutions

* **The Error Message (Your Solution):**
This is the most robust approach. It respects the principle of "Fail Fast." It avoids "magic" behavior that can be impossible to debug later. By forcing the user to correct their environment (installing `gmake` and aliasing it), you ensure that their environment is consistent for *all* projects, not just this one.
* **The Wrapper Method (Alternative):**
The `GNUmakefile` vs `Makefile` precedence is a clever workaround.
* **Pros:** Truly transparent to the user.
* **Cons:** It bifurcates your build entry point. You now have to maintain `Makefile` solely as a trampoline. If someone accidentally edits `Makefile` thinking it's the real build logic, things break.



### Summary of Syntax Quirks

Your glossary is accurate. A few nuances to reinforce:

* **`.DEFAULT`:** As you noted, this is a "last resort" handler, not a "pre-execution" handler. It is only consulted if Make cannot find *any* other way to make a target.
* **`$(eval)`:** This happens strictly during the **Parsing Phase**. By the time Make starts the **Execution Phase** (running recipes), all `$(eval)`s have already finished. This is why you can't use `$(eval)` to react to runtime events.

### Next Step

Since you have settled on the "Clear Error Message" approach, would you like me to generate a cross-platform snippet that detects the Make version and `OS` (to only error on macOS, for example) to minimize disruption for Linux users who likely already have a compatible Make?

PR #3092, opened 2026-03-13
fix(codeserver): fix 502 bad gateway for culling probe

### LINKED ISSUE #3091: Code Server Culling 502 Bad Gateway: Missing mod_unixd and DirectoryIndex
## Description
Currently, when accessing `/api/kernels/` on the code-server image to retrieve the `last_activity` for culling purposes, the NGINX proxy returns a `502 Bad Gateway` and the backend Apache HTTPD CGI server fails to start.

This is caused by two missing dependencies/configurations in the `httpd.conf`:
1. **Missing `mod_unixd`**: HTTPD fails to start with `AH00136: Server MUST relinquish startup privileges before accepting connections.`
2. **Missing `mime.types`**: HTTPD fails with `AH01597: could not open mime types config file` because `mod_mime` is loaded but the container doesn't have the file.
3. **Missing `DirectoryIndex`**: Even after starting HTTPD, accessing `/api/kernels/` returns a 404 because Apache does not know to serve `access.cgi` by default for the directory.

## Expected Behavior
The culler should be able to successfully retrieve the JSON payload containing the `last_activity` from the code-server.

## Proposed Fix
In `codeserver/ubi9-python-3.12/httpd/httpd.conf`:
- Add `LoadModule unixd_module modules/mod_unixd.so`
- Remove `LoadModule mime_module modules/mod_mime.so`
- Add `DirectoryIndex access.cgi` to the `/opt/app-root` Directory block.

EdisonSu768:
Fixes #3091

## Summary

Fix 502 Bad Gateway error that occurs during Activity Culling probe in Code Server workbench.

### Changes

- Add `mod_unixd` module to httpd.conf (required for CGI execution on Unix systems)
- Remove `mod_mime` module (not needed for this use case)
- Add `DirectoryIndex access.cgi` to ensure proper routing to the CGI script

### Problem

The culling probe was returning 502 errors because the Apache httpd server wasn't properly configured to execute CGI scripts. The missing `mod_unixd` module prevented the CGI server from starting correctly.

### Solution

This fix ensures the `/api/kernels` endpoint correctly routes to `access.cgi` for Activity Culling functionality.

### DIFF
--- a/codeserver/ubi9-python-3.12/httpd/httpd.conf
+++ b/codeserver/ubi9-python-3.12/httpd/httpd.conf
@@ -4,8 +4,8 @@ Listen 8080
 
 # Load modules
 LoadModule mpm_prefork_module modules/mod_mpm_prefork.so
+LoadModule unixd_module modules/mod_unixd.so
 LoadModule authz_core_module modules/mod_authz_core.so
-LoadModule mime_module modules/mod_mime.so
 LoadModule dir_module modules/mod_dir.so
 LoadModule cgi_module modules/mod_cgi.so
 LoadModule rewrite_module modules/mod_rewrite.so
@@ -32,6 +32,7 @@ DocumentRoot "/opt/app-root"
     Options +ExecCGI
     Require all granted
     AddHandler cgi-script .cgi
+    DirectoryIndex access.cgi
 </Directory>
 
 # Error and access logs


jiridanek:
@EdisonSu768 Greetings! The fix has been taken up by @daniellutz in https://github.com/opendatahub-io/notebooks/pull/3683 so please see here. I will close your PR as a duplicate. Thanks for alerting us to the problem and providing both an issue report and a fix!

PR #3352, opened 2026-04-11
build: optimize hermetic prefetch with arch filtering and parallel downloads

### LINKED ISSUE #3250: Local hermetic builds: cachi2/output/ is a global singleton, breaks multi-image workflows
## Problem

The local hermetic build workflow requires a two-step process (prefetch, then build), but `cachi2/output/` is a single shared directory for the entire repo. This creates several problems when building multiple images locally.

### The two-step process

Every component with a `prefetch-input/` directory requires prefetching before building:

```bash
# Step 1: prefetch
./scripts/lockfile-generators/prefetch-all.sh --component-dir jupyter/minimal/ubi9-python-3.12

# Step 2: build (Makefile auto-detects cachi2/output/ and sets LOCAL_BUILD=true)
gmake jupyter-minimal-ubi9-python-3.12
```

Running `gmake` without step 1 fails with an error pointing to `prefetch-all.sh`. This is documented. What is **not** documented is the interaction between prefetch runs for different images.

### Shared output directory

All prefetch runs write to the same `cachi2/output/deps/` tree:

| Dep type | Output path | Behavior across runs |
|----------|------------|---------------------|
| RPMs | `cachi2/output/deps/rpm/` | `rm -rf` + replace (hermeto-fetch-rpm.sh:247) |
| Pip wheels | `cachi2/output/deps/pip/` | Accumulates (skips existing files) |
| Generic artifacts | `cachi2/output/deps/generic/` | Accumulates |
| NPM | `cachi2/output/deps/npm/` | Only codeserver uses this |
| Go modules | `cachi2/output/deps/gomod/` | Merges into existing tree |

### What breaks

**Scenario: build jupyter-minimal, then jupyter-pytorch**

1. Prefetch for `jupyter/minimal` -- downloads minimal's pip wheels into `cachi2/output/deps/pip/`
2. Build jupyter-minimal -- works fine
3. Prefetch for `jupyter/pytorch` -- downloads pytorch's pip wheels into the same `cachi2/output/deps/pip/`. Minimal's wheels remain (skip-if-exists). RPMs are replaced entirely.
4. Build jupyter-pytorch -- works fine
5. **Re-build jupyter-minimal without re-prefetching** -- builds with pytorch's RPM set and a mix of both images' pip wheels. May work by accident or may fail/produce a wrong image.

The pip accumulation behavior (step 3) means the directory grows unboundedly and contains wheels from all previously-prefetched images. While this doesn't cause build failures (extra wheels are ignored by `pip install --find-links`), it wastes disk space and makes the cache non-reproducible.

The RPM replacement behavior is more dangerous: after prefetching pytorch, minimal's build uses pytorch's RPM set. Since all jupyter images currently share the same `rpms.in.yaml` (via the `prefetch-input` symlink), this happens to work today, but it's fragile.

### What's not possible

- **Building multiple images in parallel** -- concurrent writes to `cachi2/output/` would corrupt it
- **Keeping prefetch cached across image switches** -- you must re-run `prefetch-all.sh` every time you switch which image you're building (at least for pip wheels to be correct)
- **Validating that the prefetch matches the build** -- the Makefile checks that `cachi2/output/` exists, but not that it was prefetched for the correct component

## Possible solutions

1. **Per-component output directories**: `cachi2/output/<component>/deps/` instead of `cachi2/output/deps/`. The Makefile would mount the correct one based on the build target.

2. **Prefetch-and-build in one step**: A wrapper that runs prefetch + make atomically, removing the need to manage the shared directory manually. Could be a make target like `gmake hermetic-jupyter-minimal-ubi9-python-3.12`.

3. **Document the limitation**: At minimum, add a warning to `scripts/lockfile-generators/README.md` and the `prefetch-all.sh` output explaining the singleton behavior.

## Related

- #2076 -- Use caching in our Dockerfiles
- [RHAIENG-2061](https://redhat.atlassian.net/browse/RHAIENG-2061) -- Jira tracker for Dockerfile caching
- #2669 -- Optimize DNF package manager performance

jiridanek:
## Problem

Currently, `scripts/lockfile-generators/prefetch-all.sh` downloads dependencies for all supported architectures (`x86_64`, `aarch64`, `ppc64le`, `s390x`) sequentially before the build begins. This results in a massive waste of network bandwidth, disk space, and CI compute time (e.g., downloading ~5.8 GB of wheels and RPMs when the target architecture only needs ~1.4 GB).

Furthermore, `cachi2/output/` was hardcoded as a global singleton. When building multiple different images sequentially or in parallel, the prefetch scripts would overwrite or pollute this shared directory, leading to corrupted or incorrect builds (Fixes #3250).

## Solution

This PR introduces massive optimizations for the local and CI hermetic prefetch process (Phase 1):
1. **Architecture filtering (`--arch` flag):** Added to `prefetch-all.sh` to prevent downloading GBs of unused pip wheels and RPMs for other architectures.
2. **Parallel Downloads:** Parallelized NPM and Pip downloads using `xargs` and `concurrent.futures`.
3. **Namespaced `cachi2/output/`:** The output directory is now dynamically namespaced by the component hash (e.g., `cachi2/output/<hash>/deps/`). This prevents directory collisions and enables safe parallel local builds. The `Makefile` has been updated to auto-detect and mount this new directory structure.
4. **Documentation:** Updated all related hermetic build documentation to reflect the new architecture.

cc @coderabbitai

## Summary by CodeRabbit

* **New Features**
  * Added architecture-aware build option for arch-specific dependency selection.
  * Switched to per-component hashed dependency output directories to avoid cross-component conflicts.

* **Performance Improvements**
  * Parallelized npm and pip downloads for faster dependency fetching.

* **Documentation**
  * Added detailed hermetic build architecture guide and updated build docs to reflect per-component outputs and arch handling.


ysok:
I think some files here are outdated and no longer used in the prefetch-all.sh script, eg. helper/download-pip-packages.py.

For pip wheel, I directly wget the URL specified in requirements.txt instead of using Hermeto like [gomod](https://github.com/opendatahub-io/notebooks/blob/main/scripts/lockfile-generators/helpers/hermeto-fetch-gomod.sh), rpm, or npm. It was a quick solution when I first started working on it, but for a proper solution, I recommend switching to Hermeto and reading the value from the .tekton/*yaml file to determine which architecture it needs to fetch. So ideally, we should replicate the way it was setup and being used in Konflux. Eg. we define arch here in .tekton [file](https://github.com/opendatahub-io/notebooks/blob/main/.tekton/odh-workbench-jupyter-minimal-cpu-py312-ubi9-odh-main-push.yaml#L61).

For RPM, we defined those [arches](https://github.com/opendatahub-io/notebooks/blob/main/prefetch-input/rhds/rpms.in.yaml#L25) in rpms.in.yaml 

Note: Hermeto creates a random and unique temporary directory when used, preventing it from polluting with other directories.

jiridanek:
@ysok Thanks for the review! A few clarifications:

**`download-pip-packages.py` is actively used** — this PR actually wired it into the prefetch chain. The call path is `prefetch-all.sh` → `create-requirements-lockfile.sh --download` → `download-pip-packages.py`. The old inline shell download loop (73 lines parsing pylock.toml URLs directly) was replaced with a call to this script, which reads from `requirements.txt` instead. That's actually a simplification — it means we could eventually drop pylock.toml generation entirely and just use `requirements.txt`.

**Hermeto for pip (Phase 2)** — Agreed that switching pip to Hermeto (like gomod/rpm/npm) would be the proper long-term solution, including reading arch from `.tekton/*.yaml`. This PR takes the pragmatic incremental approach (parallel wget + `--arch` filtering) since it works now and cuts download size from ~5.8 GB to ~1.4 GB. Happy to track the Hermeto migration as a follow-up issue.

**Arch from `.tekton/*.yaml`** — Good pointer to the [platform definitions](https://github.com/opendatahub-io/notebooks/blob/main/.tekton/odh-workbench-jupyter-minimal-cpu-py312-ubi9-odh-main-push.yaml#L61). For local dev, `--arch` defaults to the host architecture (`uname -m`), which is the common case. Reading from Tekton YAML would make sense when we move to Hermeto.

**Hermeto temp dirs** — Right, Hermeto already namespaces its output. The MD5-hashed `cachi2/output/<hash>/` we added here is for the non-Hermeto steps (pip wget, generic artifacts) that don't get that for free. Once everything goes through Hermeto this namespacing could be simplified.

I've also addressed the @coderabbitai review findings in the latest commit (efd17a41e) — npm failure propagation, arch filtering false positives, `CACHI2_OUT_DIR` not being used by `hermeto-fetch-gomod.sh`/`download-rpms.sh`, trailing slash normalization, etc.

ide-developer:
@ysok Update on the prefetch architecture — PR #3653 rewrites `prefetch-all.sh` to use a single hermeto invocation for all ecosystems (matching the Konflux `prefetch-dependencies` task pattern). This addresses the disk exhaustion issue (#3641) and gives us parallel downloads via hermeto's asyncio.

However, we hit a hermeto limitation on ppc64le/s390x: hermeto's pip `binary` filter ignores environment markers like `platform_machine != "ppc64le"`, causing `PackageRejected` errors for packages that should be skipped on those architectures (filed upstream as hermetoproject/hermeto#1570).

The current workaround in #3653 disables the binary filter for ppc64le/s390x, which means hermeto downloads sdists instead of binary wheels — a regression from the old wget-based approach which DID fetch AIPCC binary wheels for all arches.

This means **there may still be value in a custom parallel pip downloader** (the `download-pip-packages.py` from this PR) for ppc64le/s390x specifically, until hermeto fixes the environment marker handling upstream. The custom downloader can:
1. Respect environment markers before downloading (skip packages excluded for the target arch)
2. Fetch binary wheels from the AIPCC index (not just sdists)
3. Run in parallel with `ThreadPoolExecutor`

For x86_64/aarch64, hermeto works perfectly with the binary filter. The question is whether to keep the custom downloader as a ppc64le/s390x fallback or wait for the upstream fix.

@coderabbitai

ide-developer:
Superseded by #3661 which takes the same approach (parallel arch-aware pip downloads) on a fresh branch from main.

This branch had drifted too far from main (551 files diverged) to rebase cleanly. The key improvements (parallel ThreadPoolExecutor, should_keep_for_arch, xargs -P 10 for npm) have been ported to the new PR with additional fixes (BUILD_ARCH for QEMU cross-builds, Tekton variant lookup, sdist skipping for AIPCC).

ide-developer:
@coderabbitai please take a look at https://github.com/opendatahub-io/notebooks/pull/3352#issuecomment-4475844418 as well.

ide-developer:
@coderabbitai please take a look at https://github.com/opendatahub-io/notebooks/pull/3352#issuecomment-4478260111 as well.

PR #3419, opened 2026-04-21
[RHAIENG-4591] ci(piplock-renewal) updates piplock-renewal to skip pr it changes are only timestamps

mtchoum1:
## Description

- piplock-renewal workflow 1: After make test, detect when the working tree differs from HEAD only because of ephemeral UV/pylock metadata (e.g. --exclude-newer=… timestamps). If so, restore the worktree and do not open a PR, avoiding noisy multi-file renewals and rebase friction when package pins are unchanged.
- scripts/ci/lockfile_renewal_diff_is_metadata_only.py (+ unit tests): Implements that normalization/compare logic and documents exit codes for shell use.
- manifests/tools/update_imagestream_annotations_from_pylock.py: Expand abbreviated commit SHAs to full OIDs via the GitHub REST API when fetching from github.com, so git fetch <url> <short-sha> does not fail treating short hex as a ref name; wire GITHUB_TOKEN into the ImageStream refresh step for API auth/rate limits. Documentation updated for fork CI and RHDS-only commits.

## How Has This Been Tested?

./uv run pytest tests/unit/scripts/ci/test_lockfile_renewal_diff_is_metadata_only.py (unit coverage for metadata-only detection)

GitHub Actions: Ran Lock Files Renewal Action on the fork; job refresh-lock-files completed successfully — [run log](https://github.com/mtchoum1/notebooks/actions/runs/24741226422/job/72381155884).

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Improved automation to detect and skip non-substantive timestamp/metadata-only updates, preventing creation of unnecessary pull requests and reducing noise in the project workflow.

### DIFF
--- a/.github/workflows/piplock-renewal.yaml
+++ b/.github/workflows/piplock-renewal.yaml
@@ -112,6 +112,12 @@ jobs:
             exit 0
           fi
 
+          # Skip PR creation if only timestamp changes
+          if git diff --cached -I '^#.*--exclude-newer' --quiet; then
+            echo "Only timestamp changes. Skipping PR creation."
+            exit 0
+          fi
+
           BRANCH_NAME="lockfile-update-$(date +%Y%m%d-%H%M)"
           git checkout -b "$BRANCH_NAME"
           git commit -m "Update lock files and ImageStream dependency annotations"


scripts/ci/lockfile_renewal_diff_is_metadata_only.py
jiridanek, line 1: @coderabbitai can we do this in fewer lines of code? read suggestion below; then suggest 3 more alternatives

## Proposed fix: `git diff -I` regex

Git (2.36+) supports `-I <regex>` which **ignores hunks where all removed/added lines match the regex**. ubuntu-latest (24.04) ships Git 2.43+, so this is available.

Replace the current check in the "Create Pull Request" step with:

```yaml
      - name: Create Pull Request
        env:
          GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
        run: |
          git add .
          if git diff --cached --quiet; then
            echo "No changes to commit."
            exit 0
          fi

          # Skip PR when the only changes are --exclude-newer timestamps in pylock headers
          if git diff --cached -I '^#.*--exclude-newer' --quiet; then
            echo "Only --exclude-newer timestamp changes detected, skipping PR."
            exit 0
          fi

          BRANCH_NAME="lockfile-update-$(date +%Y%m%d-%H%M)"
          ...
```

The logic becomes:
1. `git diff --cached --quiet` -- if no changes at all, bail (existing check)
2. `git diff --cached -I '^#.*--exclude-newer' --quiet` -- if changes exist but ALL of them are the timestamp comment line, bail (new check)
3. Otherwise, real changes exist, proceed with commit + PR

## Why this works

- The `--exclude-newer` timestamp only appears in a `#`-prefixed comment line in each `pylock.toml`
- If actual packages changed, there will be additional hunks with `[[packages]]` / hash / URL changes that do NOT match the regex
- ImageStream annotations, Dockerfiles, and kustomization only change when actual package versions change, so those would also survive the filter
- No changes needed to the lock generator itself

jiridanek: @coderabbitai the thing is, two files may have substantive updates, three others just modify timestamp; what we want is undo the three other files and commit the two with substantive update

ysok: Yeah agreed with @jiridanek, using git diff is good enough to find the diff, make code changes smaller.

jiridanek: @coderabbitai what we actually want here is to stop the https://github.com/opendatahub-io/notebooks/blob/main/.github/workflows/code-quality.yaml#L17 `check-generated-code` ci/generate_code.sh from pulling in random new build tags of a package that were created after the pylock.toml has been locked; maybe the whole approach requires rethinking

maybe we should use the

```
      --custom-compile-command <CUSTOM_COMPILE_COMMAND>
          The header comment to include at the top of the output file generated by `uv pip compile` [env:
          UV_CUSTOM_COMPILE_COMMAND=]
```

switch and maybe keep timestamp in a .timestamp file? alternatives?

jiridanek: @coderabbitai the timestamp written to file  should not be the timestamp when the locking happened, but actually the timestamp when the most recent package in the lockfile was published

that's enough to keep the churn out; and that means we actually don't need to store the timestamp; it's enough if the ci script extracts/computes it and then uses it

I remember that lockfiles did not have the timestamps before I added the --before flag... but maybe that's actually something that has to do with the url, `upload-time = 2026-04-01T00:22:14Z`, that stuff was missing, maybe it's because now we say we want json in pylocks_generator.py.

jiridanek: @coderabbitai any reason not to load the lockfile as toml? why parse regex?

ysok:
@mtchoum1  I am not sure if you are still working on this PR, but for this PR it shouldn't be this complicated, here is my untested version.` My Cursor prompt: Read the Jira https://redhat.atlassian.net/browse/RHAIENG-4591 and propose code changes.`

image

atheo89 (approved):

ysok (approved):

scripts/ci/lockfile_renewal_diff_is_metadata_only.py
jiridanek: @coderabbitai create github issue to implement this

jiridanek: @coderabbitai I created the issue manually since you couldn't reply to the outdated diff:

#3449 — `pylocks_generator: use max(upload-time) from pylock.toml as --exclude-newer cutoff`

PR #3470, opened 2026-04-29
Update lock files and ImageStream dependency annotations

ide-developer:
Automated lock file update, Dockerfile/kustomization regeneration, and ImageStream `notebook-python-dependencies` / `notebook-software` annotations from pylocks.

**Auto-merge policy:** This PR will be automatically merged after 1 working day unless:
- Moved to draft status
- Labeled with `do-not-merge/*`
- Manually merged or closed


jiridanek (approved):

atheo89:
🔴 There are issues with the builds:  

### **For PyTorch (ODH):** 
co-pilot assisstant: 👇 

I'll fetch the logs for this failing job to understand what went wrong.Now let me get the Dockerfile and Makefile to better understand the build context:## Problem Analysis

The build is failing during the Docker image build for `cuda-jupyter-pytorch-ubi9-python-3.12`. The error occurs when trying to install packages using `dnf`:

```
Problem 1: package perl-devel-4:5.32.1-483.el9.x86_64 from appstream requires redhat-rpm-config, but none of the providers can be installed
- package glibc-2.34-168.el9_6.24.x86_64 from @System requires (glibc-gconv-extra(x86-64) = 2.34-168.el9_6.24 if redhat-rpm-config), but none of the providers can be installed
- package libomp-19.1.7-2.el9.x86_64 from @System requires llvm-libs(x86-64) = 19.1.7-2.el9, but none of the providers can be installed
```

This is a **package dependency conflict** in the UBI9 repository. The `dnf install` command is failing to resolve dependencies for `perl`, `mesa-libGL`, `skopeo`, `compat-openssl11`, and `openshift-clients`.

## Solution

The issue is in the `dnf-helper.sh` script. While it already handles the `upgrade` case with `--nobest --skip-broken`, the `install` command doesn't have the same fallback options. Update the script to include these flags for the `install` case:

**File: `base-images/utils/dnf-helper.sh`**

```bash
#!/usr/bin/env bash
set -Eeuxo pipefail

# Verified against dnf 4.14.0 on UBI9/RHEL 9.7 (dnf config-manager --dump)
DNF_OPTS=(
    -y
    --nodocs
    # do not set --noplugins, we do need subscription-manager plugin
    --setopt=install_weak_deps=0
    --setopt=max_parallel_downloads=10
    --setopt=keepcache=0
    --setopt=deltarpm=0
)

COMMAND="${1:-}"
shift || true

case "$COMMAND" in
    upgrade)
        # Problem: The operation would result in removing the following protected packages: systemd
        #  (try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages)
        # Solution: --best --skip-broken does not work either, so use --nobest
        dnf upgrade --refresh --nobest --skip-broken "${DNF_OPTS[@]}" "$@"
        ;;
    install)
        # Use --nobest --skip-broken to handle repository dependency conflicts
        dnf install --nobest --skip-broken "${DNF_OPTS[@]}" "$@"
        ;;
    *)
        echo "Usage: $0 {upgrade|install} [packages...]"
        exit 1
        ;;
esac

dnf clean all
rm -rf /var/cache/yum /var/cache/dnf
```

## Key Changes

- Added `--nobest --skip-broken` flags to the `dnf install` command (line 26)
- This allows dnf to skip problematic packages and install what it can, rather than failing the entire operation
- The flags tell dnf to:
  - `--nobest`: Don't require the best candidate version, use any available version that satisfies dependencies
  - `--skip-broken`: Skip packages with broken dependencies instead of failing

This matches the existing pattern already used for the `upgrade` command and should resolve the repository conflict during the Docker build.

----

🔴  
### **Minimal [RHDS]**
co-pilot assisstant: 👇 

Now let me get the Dockerfile referenced in the build to understand the context better:## Analysis

The **CUDA Jupyter Minimal [RHOAI] build failure** is caused by **SSL certificate verification failures when downloading Red Hat repository metadata for ARM64 (aarch64) architecture**.

### Root Cause

The error occurs during the Docker build at **STEP 5/9** when attempting to install packages via `dnf`:

```
Curl error (91): SSL server certificate status verification FAILED for 
https://cdn.redhat.com/content/dist/rhel9/9.6/aarch64/baseos/os/repodata/...
[No OCSP response received]
```

The issue is that the OCSP (Online Certificate Status Protocol) response is not being received from Red Hat's CDN, preventing the SSL certificate validation from completing successfully.

### Key Issues

1. **ARM64 Repository SSL/OCSP Issues**: The build is targeting `linux/arm64` but the Red Hat CDN may be experiencing connectivity or OCSP responder issues for the ARM64 mirror.

2. **Missing RPM GPG Key**: Line 20 in the Dockerfile attempts to import a GPG key that appears to be missing:
   ```dockerfile
   RUN rpm --import /cachi2/output/deps/generic/RPM-GPG-KEY-CentOS-Official || true
   ```
   The error message shows: `error: /cachi2/output/deps/generic/RPM-GPG-KEY-CentOS-Official: import read failed(2).`

### Solution

**Update the Dockerfile to handle OCSP certificate validation issues:**

```dockerfile
# Replace lines 20-21 with:
RUN rpm --import /cachi2/output/deps/generic/RPM-GPG-KEY-CentOS-Official || true
RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release

# Add DNS and SSL configuration to bypass OCSP issues
RUN echo -e '[main]\nsslverify=false\nccl_ca_cert_dir=/etc/pki/tls/certs' >> /etc/yum.repos.d/redhat-rhui.conf || true

# Or, set environment variables for the dnf command:
ENV CURL_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt

# Update line 28-29 to handle certificate issues:
RUN --mount=type=bind,source=base-images/utils/dnf-helper.sh,target=/utils/dnf-helper.sh,ro \
    CURL_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt /utils/dnf-helper.sh install perl mesa-libGL skopeo compat-openssl11 openshift-clients
```

**Alternative: Update dnf-helper.sh script to disable OCSP checks:**

If the `base-images/utils/dnf-helper.sh` script exists, add OCSP bypass flags:

```bash
# Add to dnf-helper.sh before running dnf install
dnf install -y \
  --setopt=ssl_verify_host=0 \
  --setopt=ssl_verify_peer=0 \
  perl mesa-libGL skopeo compat-openssl11 openshift-clients
```

**Recommended immediate fix (least disruptive):**

Modify the Dockerfile to skip OCSP validation temporarily:

```dockerfile
# Add after line 17 (USER 0):
RUN echo "sslverify=false" >> /etc/yum.conf || true

# Or wrap the dnf call with environment variable:
RUN --mount=type=bind,source=base-images/utils/dnf-helper.sh,target=/utils/dnf-helper.sh,ro \
    REQUESTS_CA_BUNDLE="" CURL_CA_BUNDLE="" /utils/dnf-helper.sh install perl mesa-libGL skopeo compat-openssl11 openshift-clients
```

The root issue is infrastructure-related (Red Hat CDN SSL/OCSP), so the build should be retried once the CDN connectivity is restored, or the Dockerfile should be configured to handle certificate validation more gracefully for builds in restricted networks.

jiridanek:
Rerun
* https://github.com/opendatahub-io/notebooks/actions/runs/25169224066

jiridanek:
@coderabbitai what do you think about it?

jiridanek:
@coderabbitai recreated the PR in
* https://github.com/opendatahub-io/notebooks/pull/3475

Issue #3449, opened 2026-04-28
pylocks_generator: use max(upload-time) from pylock.toml as --exclude-newer cutoff


jiridanek:
## Problem

The `--exclude-newer` value used by `pylocks_generator.py` is the wall-clock time when `uv pip compile` runs (`utc_now_iso()`). This timestamp drifts on every renewal run, causing the pylock.toml header comment to change even when no actual packages changed.

PR #3419 added a `git diff -I '^#.*--exclude-newer' --quiet` guard in the piplock-renewal workflow to skip PRs when the only changes are timestamp comments. However, two gaps remain:

1. **All-or-nothing filtering**: the current `git diff -I` check applies to the entire staged tree. If some files have substantive package changes and others only have timestamp changes, all files are committed — including the timestamp-only ones.

2. **Root cause unaddressed**: the `--exclude-newer` value still drifts on every run. The `git diff -I` guard is a workaround, not a fix.

### Primary: use `max(upload-time)` as `--exclude-newer` cutoff

Since pylock.toml (PEP 751) already contains `upload-time` on every wheel entry (e.g. `upload-time = 2026-04-01T00:22:14Z`), the tightest stable cutoff is `max(upload-time)` across all wheel entries in the existing lockfile. This value is content-derived and only advances when a genuinely newer package lands in the lockfile.

Add a new function in `scripts/pylocks_generator.py`:

```python
import tomllib
from datetime import datetime
from pathlib import Path

def parse_max_upload_time_from_lockfile(path: Path) -> datetime | None:
    """Return the latest upload-time across all wheel entries in a pylock.toml."""
    if not path.is_file():
        return None
    try:
        with path.open("rb") as f:
            data = tomllib.load(f)
    except (OSError, tomllib.TOMLDecodeError):
        return None

    times: list[datetime] = []
    for pkg in data.get("packages", []):
        for wheel in pkg.get("wheels", []):
            t = wheel.get("upload-time")
            if isinstance(t, datetime):
                times.append(t)
    return max(times) if times else None
```

Then update `resolve_exclude_newer()` to prefer this value in CI check mode:

```python
def resolve_exclude_newer(lockfile, *, ci_check, live_timestamp):
    if not ci_check:
        return live_timestamp
    max_t = parse_max_upload_time_from_lockfile(lockfile)
    if max_t is not None:
        return max_t.strftime("%Y-%m-%dT%H:%M:%SZ")
[...]

PR #3555, opened 2026-05-07
RHAIENG-4443: ci: add lenient check for generated files in PRs

daniellutz:
This PR aims to implement the changes from https://redhat.atlassian.net/browse/RHAIENG-4443

## Description

Implements a two-mode check for ci/generate_code.sh output:
  - STRICT mode (main branch): Fails on any uncommitted files
  - LENIENT mode (pull requests): Allows lock file changes in directories not touched by the PR, preventing false-positive failures when AIPCC index publishes new wheel builds with same version but different hashes

The lenient mode still fails if:
  - Lock changes are in directories modified by the PR
  - Generator entrypoints (dockerfile_fragments.py, pylocks_generator.py, etc.) were touched
  - Non-lock files are dirty

This prevents CI from blocking valid PRs due to external index churn while maintaining strict validation for actual code generation changes.

## How Has This Been Tested?
Tests have been written and local tests are being executed

- [X] Ensure that you have run `make test` (`gmake` on macOS) before asking for review
- [X] Changes to everything except `Dockerfile.konflux` files should be done in `odh/notebooks` and automatically synced to `rhds/notebooks`. For Konflux-specific changes, modify `Dockerfile.konflux` files directly in `rhds/notebooks` as these require special attention in the downstream repository and flow to the upcoming RHOAI release.

## Merge criteria:

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -18,6 +18,13 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          # Fetch full history so the lenient check can diff against PR base
+          fetch-depth: 0
+
+      - name: Fetch PR base ref for diff
+        if: github.event_name == 'pull_request'
+        run: git fetch origin ${{ github.event.pull_request.base.ref }}
 
       - name: Setup uv and Python
         uses: ./.github/actions/setup-uv
@@ -26,16 +33,10 @@ jobs:
         run: bash ci/generate_code.sh
 
       - name: Check there aren't any modified files present
-        run: |
-          clean=$(git status --porcelain)
-          if [[ -z "$clean" ]]; then
-              echo "Empty git status --porcelain: $clean"
-          else
-              echo "::error::Please run 'bash ci/generate_code.sh' (the command from the previous step), commit the changed files locally, and push again."
-              echo "Uncommitted file changes detected: $clean"
-              git diff
-              exit 1
-          fi
+        run: python3 ci/check_generated_tree_clean.py
+        env:
+          GITHUB_EVENT_NAME: ${{ github.event_name }}
+          GITHUB_BASE_SHA: ${{ github.event.pull_request.base.sha }}
 
   pytest-tests:
     runs-on: ubuntu-latest
--- a/ci/check_generated_tree_clean.py
+++ b/ci/check_generated_tree_clean.py
@@ -1,0 +1,296 @@
+#!/usr/bin/env python3
+"""
+Check that the working tree is clean after running code generators.
+
+RHAIENG-4443: Handle AIPCC index churn gracefully in CI.
+
+This script implements a two-mode check:
+- STRICT mode (main branch pushes): Fails on ANY modified files
+- LENIENT mode (pull requests): Allows lock file changes in directories
+  not touched by the PR, but fails if lock changes are in modified directories
+  or if any generator entrypoint was touched.
+
+The lenient mode prevents false-positive CI failures when the AIPCC index
+publishes new wheel builds (same version, different hash) that fall within
+the --exclude-newer window, causing lock regeneration unrelated to PR changes.
+"""
+
+import os
+import re
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import ClassVar
+
+
+@dataclass
+class CheckConfig:
+    """Configuration for the generated tree cleanliness check."""
+
+    # Patterns matching lock artifact files (regex)
+    LOCK_ARTIFACT_PATTERNS: ClassVar[list[str]] = [
+        r".*\/uv\.lock\.d\/pylock\.(cpu|cuda|rocm)\.toml$",
+        r".*\/requirements\.(cpu|cuda|rocm)\.txt$",
+        r".*\/uv\.lock$",
+    ]
+
+    # Paths that, if modified in a PR, trigger strict checking
+    # (any of these touched = treat all dirty files as failures)
+    GENERATOR_ENTRYPOINTS: ClassVar[list[str]] = [
+        "scripts/dockerfile_fragments.py",
+        "manifests/tools/generate_kustomization.py",
+        "scripts/pylocks_generator.py",
+        "scripts/lockfile-generators/",
+        "dependencies/",
+    ]
+
+
+class GeneratedTreeChecker:
+    """Check if the working tree is clean after running code generators."""
+
+    def __init__(self, config: CheckConfig):
+        self.config = config
+        self.repo_root = self._get_repo_root()
+
+    def _get_repo_root(self) -> Path:
+        """Get the repository root directory."""
+        result = subprocess.run(
+            ["git", "rev-parse", "--show-toplevel"],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+        return Path(result.stdout.strip())
+
+    def _run_git_command(self, args: list[str]) -> str:
+        """Run a git command and return stdout."""
+        result = subprocess.run(
+            ["git", *args],
+            capture_output=True,
+            text=True,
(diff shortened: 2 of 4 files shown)


.github/workflows/code-quality.yaml
jiridanek, line 1: @coderabbitai compare this with the approach suggested in https://github.com/opendatahub-io/notebooks/pull/3419#discussion_r3125715721 and decide which to implement

<details>

https://redhat.atlassian.net/browse/RHAIENG-4443

Problem

The check-generated-code job fails on a PR when git status is dirty after ci/generate_code.sh, even when the only differences come from external changes on the AIPCC / RHOAI Python index (e.g. new torch wheel builds with new hashes/URLs), not from the PR’s own dependency or source changes.

Example

Workflow: [https://github.com/opendatahub-io/notebooks/actions/runs/24272366603/job/70879763402?pr=3341|GitHub Actions job (PR #3341)]

PR context: Hermetic builds for Runtime images

CI message: run bash ci/generate_code.sh, commit, push.

Uncommitted changes were only:

jupyter/trustyai/ubi9-python-3.12/requirements.cpu.txt — extra --hash= lines for torch==2.10.0

jupyter/trustyai/ubi9-python-3.12/uv.lock.d/pylock.cpu.toml — additional wheel URLs (e.g. torch-2.10.0-7-... from packages.redhat.com pulp) and matching hashes

Contributors perceive this as noise: the PR did not change those dependencies; the index moved underneath them.

Desired behavior

On pull requests: do not fail the PR solely because regenerated lockfiles/index artifacts drift due to AIPCC publishing new wheels for unchanged version pins (or equivalent external index churn). The PR check should focus on whether the PR’s edits are consistent with generated code.

On main (or the appropriate post-merge / release build path): still enforce that generated code and locks are current — i.e. if generate_code.sh would change tracked files, that should fail until locks are updated and committed, so main never ships stale generated artifacts.

Notes

Implementation may require splitting workflows, scoping the dirty-tree check, or only running full regeneration against a frozen index on PRs — details for the notebooks maintainers.

</details>

jiridanek, line 1: @coderabbitai what do you think about the approach in https://github.com/opendatahub-io/notebooks/issues/3449 as an alternative approach to the one in this PR?

<details>

## Problem

The `--exclude-newer` value used by `pylocks_generator.py` is the wall-clock time when `uv pip compile` runs (`utc_now_iso()`). This timestamp drifts on every renewal run, causing the pylock.toml header comment to change even when no actual packages changed.

PR #3419 added a `git diff -I '^#.*--exclude-newer' --quiet` guard in the piplock-renewal workflow to skip PRs when the only changes are timestamp comments. However, two gaps remain:

1. **All-or-nothing filtering**: the current `git diff -I` check applies to the entire staged tree. If some files have substantive package changes and others only have timestamp changes, all files are committed — including the timestamp-only ones.

2. **Root cause unaddressed**: the `--exclude-newer` value still drifts on every run. The `git diff -I` guard is a workaround, not a fix.

## Proposed Solution

### Primary: use `max(upload-time)` as `--exclude-newer` cutoff

Since pylock.toml (PEP 751) already contains `upload-time` on every wheel entry (e.g. `upload-time = 2026-04-01T00:22:14Z`), the tightest stable cutoff is `max(upload-time)` across all wheel entries in the existing lockfile. This value is content-derived and only advances when a genuinely newer package lands in the lockfile.

Add a new function in `scripts/pylocks_generator.py`:

```python
import tomllib
from datetime import datetime
from pathlib import Path

def parse_max_upload_time_from_lockfile(path: Path) -> datetime | None:
    """Return the latest upload-time across all wheel entries in a pylock.toml."""
    if not path.is_file():
        return None
    try:
        with path.open("rb") as f:
            data = tomllib.load(f)
    except (OSError, tomllib.TOMLDecodeError):
        return None

    times: list[datetime] = []
    for pkg in data.get("packages", []):
        for wheel in pkg.get("wheels", []):
            t = wheel.get("upload-time")
            if isinstance(t, datetime):
                times.append(t)
    return max(times) if times else None
```

Then update `resolve_exclude_newer()` to prefer this value in CI check mode:

```python
def resolve_exclude_newer(lockfile, *, ci_check, live_timestamp):
    if not ci_check:
        return live_timestamp
    max_t = parse_max_upload_time_from_lockfile(lockfile)
    if max_t is not None:
        return max_t.strftime("%Y-%m-%dT%H:%M:%SZ")
    parsed = parse_exclude_newer_from_lockfile(lockfile)
    return parsed if parsed is not None else live_timestamp
```

`tomllib` is stdlib from Python 3.11+, which this repo already requires.

### Secondary: per-file `git diff -I` filtering in the workflow

Upgrade the all-or-nothing check in `.github/workflows/piplock-renewal.yaml` to a per-file loop that restores timestamp-only files before committing:

```bash
while IFS= read -r f; do
  if git diff HEAD -I '^#.*--exclude-newer' --quiet -- "$f"; then
    git restore --source=HEAD -- "$f"
    echo "Restored (metadata-only): $f"
  fi
done < <(git diff --name-only HEAD)
```

This ensures that when some lockfiles have real changes and others only have timestamp drift, only the substantive changes are committed.

## Key Files

- `scripts/pylocks_generator.py` — `resolve_exclude_newer()` (line ~290), `parse_exclude_newer_from_lockfile()` (line ~270)
- `.github/workflows/piplock-renewal.yaml` — "Create Pull Request" step (lines 115-118)

## References

- PR #3419 review discussion: https://github.com/opendatahub-io/notebooks/pull/3419#discussion_r3125558534
- pylock.toml `upload-time` field is present on all wheel entries (both RH-index and public-index lockfiles)

</details>

daniellutz: @jiridanek we could absolutely use the git -I pattern used in 3449 here as well, it just wasn't my main idea in the first moment, but I'm really open to changing it as well if you all think would be better

jiridanek: @coderabbitai this one should also help https://github.com/pulp/pulp_python/issues/1232

jiridanek: @coderabbitai https://github.com/pulp/pulp_python/issues/1232 says

<details>
PEP 691 JSON API: upload-time should reflect when package was added to repository, not pulp_created #1232

## Summary

The PEP 691 JSON Simple API response includes an `upload-time` field for each file. Currently, this field uses the content unit's `pulp_created` timestamp (when the package was first uploaded to Pulp), rather than the date the package was added to the specific repository being queried.

## Problem

When a package is uploaded to Pulp and later added to a different repository (e.g. via copy or modify), the `upload-time` in the JSON API still reflects the original upload date, not when it became available in the repository serving the request.

### Example

- `ray-2.54.1-cp312-cp312-manylinux2014_aarch64.whl` was uploaded to Pulp on **2026-03-26**
- It was added to the `rhoai-3.5-EA1-cpu-ubi9` repository on **2026-05-12** (repository version 370)
- The PEP 691 JSON response for that repository reports `upload-time` as **2026-03-26** (the `pulp_created` date)

## Impact

Tools like `uv` use `--exclude-newer` to filter packages by `upload-time` for reproducible dependency resolution. When `upload-time` reflects the original upload rather than when the package appeared in the repository, the resolver may include packages that were not yet available in the repository at the intended cutoff time, or exclude packages that were.

This is particularly relevant for workflows that copy packages between repositories (e.g. from a staging repo to a release repo) where the repository addition date is the meaningful timestamp for consumers.

## Expected Behavior

The `upload-time` field in the PEP 691 JSON API response should reflect when the package was added to the repository being queried, not the global `pulp_created` timestamp of the content unit.

## Reference

- [PEP 691 – JSON-based Simple API for Python Package Indexes](https://peps.python.org/pep-0691/)
- [PEP 700 – Additional Fields for the Simple API for Package Indexes](https://peps.python.org/pep-0700/) (defines `upload-time`)
</details>

jiridanek: @coderabbitai even if `max(upload-time)` is now unreliable, we can still use the lockfile's last modification time from git for `uv pip compile --exclude-newer`

PR #3677, opened 2026-05-19
RHOAIENG-62436: Integrate Kale into Notebooks

hmtosi:
## Description
This PR is for the integration of Kale into RHOAI notebook images for the first time. The epic is [here](https://redhat.atlassian.net/browse/RHOAIENG-62436)
All relevant changes will be made on this branch - multiple child issues within the epic will be included on this one PR.

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Enabled Kubeflow Kale integration across notebook runtimes with automatic Elyra-to-Kale pipeline configuration at startup when runtime settings are present.
  * Disables the JupyterLab Kale extension during image builds.
* **Bug Fixes**
  * Improved startup configuration to safely apply Kale/KFP settings only when valid Elyra runtime configuration is available.
* **Dependency Updates**
  * Added Kale runtime dependencies (including Kubeflow Kale and notebook-related packages) and updated pinned requirement sets across images.
* **Tests**
  * Expanded/streamlined unit tests covering Elyra→Kale mapping, auth handling, SSL CA selection, and failure/no-save scenarios.


runtimes/datascience/ubi9-python-3.12/utils/requirements-kale.txt
```diff
@@ -0,0 +1,6 @@
+# Originally copied from: https://github.com/kubeflow/kale
+# This is a comprehensive list of python dependencies that Kale requires to execute Jupyter notebooks.
+
+# This (now effectively blank) requirements.txt file is left here for the exclusive reason to escape falling into
+# similar dependency injection issues as with Elyra. In case the details are needed, please follow the
+# comprehensive list of python dependencies present in odh-notebooks-meta-runtime-kale-deps.
```
jesuino, line 6: We may probably have to add kale dependency here as well:
```
kubeflow-kale==2.0.0
```
This dependency will surely be used for all pipelines that were built with Kale, but we need to confirm that this file is being used on the runtime image

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,105 @@
...
+    traceback.print_exc()
+EOF
+  fi
+fi
+
+# Set environment variables for KFP authentication
+export KF_PIPELINES_SA_TOKEN_PATH="/var/run/secrets/kubernetes.io/serviceaccount/token"
+export KF_PIPELINES_SSL_SA_CERTS="${KF_PIPELINES_SSL_SA_CERTS:-/var/run/secrets/kubernetes.io/serviceaccount/ca.crt}"
+
+# Disable Kale extension by default - users can enable it via JupyterLab UI
```
jesuino: This is something that we need to test: will enabling Kale on the UI work?

dependencies/odh-notebooks-meta-runtime-kale-deps/pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[project]
+name = "odh-notebooks-meta-runtime-kale-deps"
```
jesuino, line 2: We need to confirm if this file is used at all and how it is used. I don't see reference of it anywhere else, but I see `odh-notebooks-meta-runtime-elyra-deps` being references on runtime images toml

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,105 @@
...
+  # Get the first Elyra runtime config file
+  ELYRA_RUNTIME_CONFIG=$(ls /opt/app-root/runtimes/..data/*.json 2>/dev/null | head -1)
+
+  if [ -n "$ELYRA_RUNTIME_CONFIG" ] && [ -f "$ELYRA_RUNTIME_CONFIG" ]; then
+    # Configure Kale KFP server connection by mapping Elyra config to Kale config
+    python3 << 'EOF'
+import json
+import os
+
+elyra_config_path = os.environ.get('ELYRA_RUNTIME_CONFIG')
```
jesuino: Later we can put this script in an independent file to independently test it (make it sure that Elyra runtime info is correctly translated into Kale runtime info.

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,105 @@
...
+set -x
+
+# RHOAIENG-57465: Setup script for Kubeflow Kale JupyterLab extension
+# This script configures Kale to connect to KFP by reading Elyra runtime config
+
+# Read Elyra config and copy the relevant information to Kale config
+# Extract KFP configuration from Elyra runtime configs if available
+if [ "$(ls -A /opt/app-root/runtimes/ 2>/dev/null)" ]; then
+  # Get the first Elyra runtime config file
+  ELYRA_RUNTIME_CONFIG=$(ls /opt/app-root/runtimes/..data/*.json 2>/dev/null | head -1)
```
jesuino: This is getting the first runtime configuration that it can find. I wonder if we should use the default runtime configuration instead and hardcode its name here. What is the runtime configuration created by notebooks when we first launch a workbench?

jesuino (changes requested):
Adding an initial review with points that we need to research/explore next

Overall it is looking good for the extension setup, but I don't know about the runtime configuration, which we will be sure only when we fire a pipeline from Kale integrated into notebooks.

dependencies/odh-notebooks-meta-runtime-kale-deps/pyproject.toml
jiridanek: @coderabbitai as it is added by this pr and not referenced from anywhere else, it is not being used

what was the motivation adding it?

dependencies/odh-notebooks-meta-runtime-kale-deps/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+    "nbconvert",
+    "nbformat",
+    "jinja2",
+    "dill",
+    "networkx",
+    "autopep8",
+    "pyflakes",
+    "astor",
+    "packaging",
+    "progress",
```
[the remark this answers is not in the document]
daniellutz: @coderabbitai we are always using latest version on purpose, as we have also our internal index and when we check on it, latest will already be compliant with our dependencies

so this is done in the index level, not in the application level

for now, let's keep it that way until some problem arises from it

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
hmtosi: Users will have to first enable the extension in the JupyterLab workbench using the command `jupyter labextension enable jupyterlab-kubeflow-kale`. They will then be able to see the Kale icon on the left hand side, open a Jupyter Notebook, and enable kale with the toggle.

hmtosi: I tested this on the UI and it works

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
hmtosi: The default name is "Pipeline" - I have hardcoded this. I updated the fallback to use the first image found, but it uses the proposed fix from coderabbit so it is more robust.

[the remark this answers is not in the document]
hmtosi: addressed

dependencies/odh-notebooks-meta-runtime-kale-deps/pyproject.toml
hmtosi: I have wired this into the same places as the elyra one.

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
hmtosi: I have moved the wrapped python to its own file, `jupyter/datascience/ubi0-python-3.12/utils/configure_kale_from_elyra.py`, but I have not tested it yet

runtimes/datascience/ubi9-python-3.12/utils/requirements-kale.txt
jesuino: I see now that Kale runtime dependencies seem to be resolved in `odh-notebooks-meta-runtime-kale-deps/pyproject.toml` file and authoring (workbench) files in `dependencies/odh-notebooks-meta-jupyterlab-datascience-deps`

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,35 @@
...
+    shopt -u nullglob
+  fi
+
+  if [ -n "$ELYRA_RUNTIME_CONFIG" ] && [ -f "$ELYRA_RUNTIME_CONFIG" ]; then
+    # Configure Kale KFP server connection by mapping Elyra config to Kale config
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    python3 "${SCRIPT_DIR}/utils/configure_kale_from_elyra.py"
+  fi
+fi
+
```
jesuino, line 34: Now that we have Kale 2.1.x we should set the following env vars here:

- `KALE_DEFAULT_BASE_IMAGE` - use the ubi Python Image
- `KALE_SECURITY_CONTEXT_RUN_AS_USER`: Use a valid user ID **OR**
- `KALE_SECURITY_CONTEXT_ENABLE=false`: To completely remove the security context by default


Note about the security context setup: If we don't find the correct answer for this setup (valid user id or disable it) we should document to users that they should adjust it before trying to run their pipelines in RHOAI.

jesuino (changes requested):
Thanks for the changes, all the other comments are resolved, just a final suggest change about the new features on Kale 2.1.x

runtimes/datascience/ubi9-python-3.12/utils/requirements-kale.txt
hmtosi: sounds good, thank you!

jupyter/rocm/pytorch/ubi9-python-3.12/uv.lock.d/pylock.rocm.toml
```diff
@@ -8,1914 +8,1942 @@ requires-python = ">=3.12"
...
 [[packages]]
 name = "kube-authkit"
 version = "0.4.0"
 marker = "python_full_version >= '3.12' and implementation_name == 'cpython' and platform_machine != 'ppc64le' and platform_machine != 's390x' and sys_platform == 'linux'"
-wheels = [{ url = "https://packages.redhat.com/api/pulp-content/public-rhai/rhoai/3.5-EA2/rocm7.1-ubi9-test/kube_authkit-0.4.0-12-py3-none-any.whl", upload-time = 2026-04-09T00:57:19Z, size = 33177, hashes = { sha256 = "b497539784b4a1a2be0c93fe30a7a74501a00b07d52ec0311b8abf601f3be22f" } }]
+wheels = [{ url = "https://packages.redhat.com/api/pulp-content/public-rhai/rhoai/3.5-EA2/rocm7.1-ubi9-test/kube_authkit-0.4.0-12-py3-none-any.whl", upload-time = 2026-05-13T16:27:21Z, size = 33177, hashes = { sha256 = "b497539784b4a1a2be0c93fe30a7a74501a00b07d52ec0311b8abf601f3be22f" } }]
+
+[[packages]]
+name = "kubeflow-kale"
+version = "2.0.0"
```
jesuino: The Kale version is not up-to-date - It should be resolved to 2.1.1

jesuino (changes requested):
Just a comment about the Kale version - it is not the latest so the env vars from the setup-kale.sh will not have effect.

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
hmtosi: done

dependencies/odh-notebooks-meta-runtime-kale-deps/pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[project]
+name = "odh-notebooks-meta-runtime-kale-deps"
+version = "0.1.0"
+requires-python = "==3.12.*"
+
+# This is a comprehensive list of python dependencies that Kale requires to execute Jupyter notebooks.
+dependencies = [
```
atheo89, line 7: Hey Hannah!
Did you try installing only `kubeflow-kale[jupyter]` through `pyproject.toml` first and verify whether all of its dependencies are installed automatically, without explicitly listing them?

Is there a specific reason for introducing a dedicated file for these dependencies? I ran a quick experiment and found that all dependencies were properly resolved except for `packaging`, which was not locked.

Instead of maintaining a separate file, could we try adding only packaging alongside `kubeflow-kale[jupyter]` and include a comment explaining that it is a companion dependency? This would reduce the number of files we need to maintain and keep the dependency management simpler.

runtimes/datascience/ubi9-python-3.12/pyproject.toml
```diff
@@ -9,6 +9,7 @@ dependencies = [
 
     "odh-notebooks-meta-db-connectors-deps",
     "odh-notebooks-meta-runtime-elyra-deps",
+    "odh-notebooks-meta-runtime-kale-deps",
```
atheo89, line 12: Why is Kale being installed into the runtimes? Since Kale is intended to replace Elyra, in theory it should not be included in the runtimes, but only in workbenches and specifically greater than minimal. For example, Elyra itself is not installed there.

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,43 @@
...
+# Set environment variables for KFP authentication
+export KF_PIPELINES_SA_TOKEN_PATH="/var/run/secrets/kubernetes.io/serviceaccount/token"
+export KF_PIPELINES_SSL_SA_CERTS="${KF_PIPELINES_SSL_SA_CERTS:-/var/run/secrets/kubernetes.io/serviceaccount/ca.crt}"
+
+# Configure Kale security context settings
+# Disable security context enforcement (leave RUN_AS_USER and RUN_AS_GROUP undefined)
+export KALE_SECURITY_CONTEXT_ENABLED=false
+
+# Set default image
+export KALE_DFAULT_BASE_IMAGE=ubi9/python-312
```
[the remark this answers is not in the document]
hmtosi: fixed

runtimes/datascience/ubi9-python-3.12/pyproject.toml
jiridanek: @coderabbitai Why is Kale being installed into the runtimes? Since Kale is intended to replace Elyra, in theory it should not be included in the runtimes, but only in workbenches and specifically greater than minimal. For example, Elyra itself is not installed there.

hmtosi: Thank you @atheo89 and @jiridanek for you question and follow up. As coderabbit is saying, I was trying to follow the elyra structure exactly with the new kale changes. I believe the code pattern follows elyra's, but if you see somewhere specifically that kale is installed where elyra is not, please let me know. 

We are not ready to directly replace elyra with kale, so we will have both for now.

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,105 @@
...
+    with open(elyra_config_path, 'r') as f:
+        elyra_config = json.load(f)
+
+    metadata = elyra_config.get('metadata', {})
+
+    # Build Kale configuration from Elyra metadata
+    kale_config = {}
+
+    # 1. host: Use public_api_endpoint if available, otherwise api_endpoint
+    kale_config['host'] = metadata.get('public_api_endpoint') or metadata.get('api_endpoint', '')
```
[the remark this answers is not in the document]
jiridanek: @coderabbitai is this fixed? and how realistic the scenario is?

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,105 @@
...
+                'env_var_password': 'KF_PIPELINES_PASSWORD'
+            }
+        else:
+            kale_config['auth_config'] = {}
+
+    # 4. ssl_ca_cert: Check if KF_PIPELINES_SSL_SA_CERTS env var is set
+    ssl_cert_path = os.environ.get('KF_PIPELINES_SSL_SA_CERTS',
+                                     '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt')
+    if os.path.exists(ssl_cert_path):
+        kale_config['ssl_ca_cert'] = ssl_cert_path
```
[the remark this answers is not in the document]
jiridanek: @coderabbitai is this fixed; anyways it does not make much sense, the whole thing runs as one single-user container, there is nowhere the single user can't look already

tests/unit/test_configure_kale_from_elyra.py
```diff
@@ -0,0 +1,547 @@
...
+
+import pytest
+
+# Import the module under test
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_UTILS_PATH = _REPO_ROOT / "jupyter/datascience/ubi9-python-3.12/utils"
+if str(_UTILS_PATH) not in sys.path:
+    sys.path.insert(0, str(_UTILS_PATH))
+
+import configure_kale_from_elyra as kale_config
```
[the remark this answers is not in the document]
jiridanek: @coderabbitai the suggested change is convoluted; what about either disabling the warning on the import, or rearranging things so that ruff/pyright can resolve it?

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,43 @@
...
+      ELYRA_RUNTIME_CONFIG=""
+    fi
+    shopt -u nullglob
+  fi
+
+  if [ -n "$ELYRA_RUNTIME_CONFIG" ] && [ -f "$ELYRA_RUNTIME_CONFIG" ]; then
+    # Configure Kale KFP server connection by mapping Elyra config to Kale config
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    export ELYRA_RUNTIME_CONFIG
+    python3 "${SCRIPT_DIR}/configure_kale_from_elyra.py"
```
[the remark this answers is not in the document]
jiridanek: @coderabbitai this is questionable advice, see https://www.martinfowler.com/ieeeSoftware/failFast.pdf, https://www.jamesshore.com/v2/blog/2005/failfastanduserinteraction

runtimes/datascience/ubi9-python-3.12/pyproject.toml
daniellutz: this is what I was going to add, @hmtosi,that for a while, both will be there, not a full-replacement happening right now

thanks for raising it up already

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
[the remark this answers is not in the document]
ide-developer: @codeerabbitai One nuance after checking upstream Kale 2.1.x: the fail-fast concern here is more concrete than a generic `|| true` debate. `configure_kale_from_elyra.py` currently writes `auth_config={"env_var_username","env_var_password"}` for `auth_type="dex"`, but Kale only accepts cookie refs (`env_var` / `file_path`) for DEX. So a valid Elyra DEX config should make `save_config()` throw during notebook startup, before the user has even opted into Kale. I left the concrete auth-shape finding inline on the converter rather than reopening the broader fail-fast argument.

ide-developer (commented):
@codeerabbitai Leaving only the findings I could not find already reported in this PR. I skipped duplicating the existing SSRF/path-traversal discussions and focused on two auth-path regressions plus one rollout gap.

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
```diff
@@ -0,0 +1,105 @@
...
+            kale_config['auth_config'] = {
+                'token_path': '/var/run/secrets/kubernetes.io/serviceaccount/token'
+            }
+
+        elif elyra_auth_type == 'EXISTING_BEARER_TOKEN':
+            kale_config['auth_type'] = 'existing_bearer_token'
+            # Use environment variable reference instead of storing the actual token
+            if metadata.get('api_password'):
+                # If api_password exists, reference it via env var
+                kale_config['auth_config'] = {'env_var': 'KF_PIPELINES_TOKEN'}
```
ide-developer, line 64: @codeerabbitai `EXISTING_BEARER_TOKEN` looks incomplete here. This only stores a reference to `KF_PIPELINES_TOKEN`, but nothing in this PR exports that env var from Elyra's `api_password` field or writes a token file that Kale can read later. So this branch saves a config that still cannot authenticate at runtime.

ide-developer (commented):
@codeerabbitai Adding the two follow-up inline comments with corrected quoting.

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
```diff
@@ -0,0 +1,105 @@
...
+            else:
+                kale_config['auth_config'] = {}
+
+        elif elyra_auth_type in ['DEX_STATIC_PASSWORDS', 'DEX_LDAP', 'DEX_LEGACY']:
+            kale_config['auth_type'] = 'dex'
+            # DEX uses username/password from api_username and api_password
+            # Kale expects these to be resolved at runtime, not stored in config
+            if metadata.get('api_username') and metadata.get('api_password'):
+                # Store reference to environment variables instead of actual credentials
+                kale_config['auth_config'] = {
```
ide-developer: @codeerabbitai I think this DEX mapping is incompatible with upstream Kale 2.1.x. `kfp_server_config.save_config()` only accepts `auth_config={"env_var"|"file_path"}` for `auth_type="dex"`, and the Dex authenticator consumes a cookie string, not username/password env vars. With `env_var_username` / `env_var_password` here, a DEX-backed Elyra config should raise during `save_config()` instead of producing a usable Kale config. Because `setup-kale.sh` is sourced under `set -e`, that also turns into a notebook-startup failure for DEX-backed configs.

dependencies/odh-notebooks-meta-jupyterlab-datascience-deps/pyproject.toml
```diff
@@ -5,6 +5,7 @@ requires-python = "==3.12.*"
 
 dependencies = [
     "odh-elyra",
+    "kubeflow-kale[jupyter]",
```
ide-developer, line 8: @codeerabbitai This dependency lands in a shared meta-package, but the PR only regenerates `jupyter/datascience/ubi9-python-3.12/requirements.cpu.txt` / `uv.lock.d/pylock.cpu.toml` and only updates the datascience Dockerfile to copy `setup-kale.sh`. The leaf workbench images (`jupyter/pytorch`, `jupyter/tensorflow`, `jupyter/trustyai`, `jupyter/rocm-*`) still build from their own unchanged lockfiles and still copy only `setup-elyra.sh`, so Kale will not actually ship there yet.

jiridanek: @coderabbitai This dependency lands in a shared meta-package, but the PR only regenerates `jupyter/datascience/ubi9-python-3.12/requirements.cpu.txt` / `uv.lock.d/pylock.cpu.toml` and only updates the datascience Dockerfile to copy `setup-kale.sh`. The leaf workbench images (`jupyter/pytorch`, `jupyter/tensorflow`, `jupyter/trustyai`, `jupyter/rocm-*`) still build from their own unchanged lockfiles and still copy only `setup-elyra.sh`, so Kale will not actually ship there yet.

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
jiridanek: @coderabbitai I think this DEX mapping is incompatible with upstream Kale 2.1.x. `kfp_server_config.save_config()` only accepts `auth_config={"env_var"|"file_path"}` for `auth_type="dex"`, and the Dex authenticator consumes a cookie string, not username/password env vars. With `env_var_username` / `env_var_password` here, a DEX-backed Elyra config should raise during `save_config()` instead of producing a usable Kale config. Because `setup-kale.sh` is sourced under `set -e`, that also turns into a notebook-startup failure for DEX-backed configs.

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
jiridanek: @coderabbitai `EXISTING_BEARER_TOKEN` looks incomplete here. This only stores a reference to `KF_PIPELINES_TOKEN`, but nothing in this PR exports that env var from Elyra's `api_password` field or writes a token file that Kale can read later. So this branch saves a config that still cannot authenticate at runtime.

jupyter/rocm/pytorch/ubi9-python-3.12/uv.lock.d/pylock.rocm.toml
jiridanek: @coderabbitai check if this is fixed

dependencies/odh-notebooks-meta-runtime-kale-deps/pyproject.toml
jiridanek: @coderabbitai please assess

runtimes/datascience/ubi9-python-3.12/pyproject.toml
atheo89: I would like to bring @harshad16 into that discussion for final confirm. I remember that the runtimes does not wire the odh-elyra package (now kale). 

This is a AI discussion when i try it to double check it. 

# ODH-Elyra wiring and package installation
_Exported on 6/16/2026 at 20:26:08 GMT+2 from Cursor (3.7.27)_

---

**User**

How odh-elyra is wired on the notebooks give me the wiring on jupyter/datascience workbench and runtime/datascience runtime how these two arw connected? and where the elyra packages are installed? 


---

**Cursor**

Tracing how odh-elyra connects the datascience workbench and runtime in this repo.


Here is how **odh-elyra** is wired between the **datascience workbench** and the **datascience runtime** in this repo.

## Short answer

They are **not linked as Python dependencies of each other**. The workbench installs the full **`odh-elyra`** package (JupyterLab UI + pipeline submission). The runtime installs **Elyra execution dependencies only** (papermill, ipykernel, etc.) plus a **`bootstrapper.py`** script. At **cluster runtime**, the **notebook controller** connects them by mounting JSON configs that tell Elyra which runtime image to use when a pipeline runs.

---

## 1. Where Elyra packages are installed

### Workbench (`jupyter/datascience/ubi9-python-3.12`)

| What | Where it comes from | Installed to |
|------|---------------------|--------------|
| **`odh-elyra`** (full Elyra fork) | `odh-notebooks-meta-jupyterlab-datascience-deps` → declares `odh-elyra` | `/opt/app-root/lib/python3.12/site-packages` during Docker build |
| **`kfp`** (pipeline client) | `odh-notebooks-meta-workbench-datascience-deps` | same venv |
| Elyra config + startup wiring | `setup-elyra.sh`, `utils/jupyter_elyra_config.py` | `/opt/app-root/bin/utils/` |

Dependency chain in workbench `pyproject.toml`:

```6:16:jupyter/datascience/ubi9-python-3.12/pyproject.toml
dependencies = [
    # Hermetic cpu-base: prefetched with the rest of the stack (see repo-root prefetch-input/ and jupyter/minimal pyproject)
    "uv",
    "micropipenv[toml]",

    # Datascience and useful extensions
    "skl2onnx",
    "odh-notebooks-meta-workbench-datascience-deps",
    "odh-notebooks-meta-db-connectors-deps",
    "odh-notebooks-meta-jupyterlab-deps",
    "odh-notebooks-meta-jupyterlab-datascience-deps",
```

The meta package that pulls in `odh-elyra`:

```6:12:dependencies/odh-notebooks-meta-jupyterlab-datascience-deps/pyproject.toml
dependencies = [
    "odh-elyra",
    "jupyter-bokeh",
    "jupyterlab-lsp",
    "jupyterlab-widgets",
    "jupyter-resource-usage",
]
```

Locked version today: **`odh-elyra==5.0.0`** in `requirements.cpu.txt`.

### Runtime (`runtimes/datascience/ubi9-python-3.12`)

| What | Where it comes from | Installed to |
|------|---------------------|--------------|
| **NOT `odh-elyra`** | — | — |
| Elyra **execution** deps (papermill, ipykernel, minio, nbconvert, …) | `odh-notebooks-meta-runtime-elyra-deps` | `/opt/app-root/lib/python3.12/site-packages` |
| DS libs (numpy, pandas, sklearn, …) | `odh-notebooks-meta-runtime-datascience-deps` | same venv |
| **`bootstrapper.py`** (KFP component entrypoint) | prefetched from `prefetch-input/elyra-v4.3.1/...` | `/opt/app-root/bin/utils/bootstrapper.py` |

Runtime `pyproject.toml`:

```6:11:runtimes/datascience/ubi9-python-3.12/pyproject.toml
dependencies = [
    # Datascience and useful extensions
    "odh-notebooks-meta-runtime-datascience-deps",

    "odh-notebooks-meta-db-connectors-deps",
    "odh-notebooks-meta-runtime-elyra-deps",
```

Runtime Dockerfile explicitly sets:

```63:65:runtimes/datascience/ubi9-python-3.12/Dockerfile.cpu
COPY --chown=1001:0 prefetch-input/elyra-v4.3.1/elyra/kfp/bootstrapper.py /opt/app-root/bin/utils/bootstrapper.py

ENV ELYRA_INSTALL_PACKAGES="false"
```

`utils/requirements-elyra.txt` is intentionally **empty** — deps are baked into the image via the meta package instead of installed at pipeline-run time.

---

## 2. How the two images connect

There is **no direct import or shared Dockerfile stage** between workbench and runtime. Connection happens at **OpenShift/K8s runtime** through ImageStreams + notebook controller mounts.

```mermaid
flowchart TB
    subgraph build [Build time - this repo]
        WB[jupyter/datascience workbench image]
        RT[runtimes/datascience runtime image]
        IS[ImageStream runtime-datascience<br/>opendatahub.io/runtime-image: true]
        RT --> IS
    end

    subgraph cluster [Cluster runtime]
        NC[Notebook Controller webhook]
        CM1[ConfigMap: pipeline runtime images]
        CM2[Secret: DSPA connection]
        POD[Workbench pod]
        EL[Elyra UI in JupyterLab]
        KFP[KFP / Argo pipeline pods]
    end

    WB --> POD
    IS --> NC
    NC --> CM1
    NC --> CM2
    CM1 -->|mount /opt/app-root/pipeline-runtimes/| POD
    CM2 -->|mount /opt/app-root/runtimes/| POD
    POD -->|setup-elyra.sh at startup| EL
    EL -->|submit pipeline, pick runtime image| KFP
    KFP -->|launch pod with| RT
```

### Step-by-step

1. **This repo publishes** the runtime image as an ImageStream (`manifests/odh/base/runtime-datascience-imagestream.yaml`) with label `opendatahub.io/runtime-image: "true"` and metadata annotation describing the Elyra runtime.

2. **Notebook controller** (outside this repo) watches those ImageStreams and mounts into every workbench pod:
   - `/opt/app-root/pipeline-runtimes/` — runtime **image** JSON (which container image to run each pipeline node on)
   - `/opt/app-root/runtimes/` — DSPA **connection** JSON (where to submit pipelines)

3. **At workbench startup**, `start-notebook.sh` sources `setup-elyra.sh`:

```7:9:jupyter/minimal/ubi9-python-3.12/start-notebook.sh
if [ -f "${SCRIPT_DIR}/utils/setup-elyra.sh" ]; then
  source ${SCRIPT_DIR}/utils/setup-elyra.sh
fi
```

4. **`setup-elyra.sh`** copies controller-provided configs into Elyra metadata dirs:

```13:21:jupyter/datascience/ubi9-python-3.12/setup-elyra.sh
# Set elyra runtime config from volume mount
if [ "$(ls -A /opt/app-root/runtimes/)" ]; then
  cp -r /opt/app-root/runtimes/..data/*.json $(jupyter --data-dir)/metadata/runtimes/
fi

# Set elyra runtime images json from volume mount
if [ "$(ls -A /opt/app-root/pipeline-runtimes/)" ]; then
  cp -r /opt/app-root/pipeline-runtimes/..data/*.json /opt/app-root/share/jupyter/metadata/runtime-images/
fi
```

5. **User runs a pipeline** in Elyra → workbench submits to KFP/Argo using `kfp` + mounted DSPA config → KFP launches **separate pods** using the **runtime image** selected in the pipeline node (e.g. `Runtime | Datascience | CPU | Python 3.12`).

6. **Runtime pod** executes notebook nodes via `bootstrapper.py` + preinstalled papermill/ipykernel stack — no Elyra JupyterLab extension needed there.

---

## 3. Workbench image prep for Elyra (build time)

The datascience Dockerfile also prepares mount points and clears default runtime-image JSON (controller fills them at runtime):

```155:157:jupyter/datascience/ubi9-python-3.12/Dockerfile.cpu
mkdir /opt/app-root/runtimes
mkdir /opt/app-root/pipeline-runtimes
rm /opt/app-root/share/jupyter/metadata/runtime-images/*.json
```

Elyra is configured for KFP-only processing:

```10:10:jupyter/datascience/ubi9-python-3.12/utils/jupyter_elyra_config.py
c.PipelineProcessorRegistry.runtimes = ['kfp']
```

---

## 4. Mental model

| Role | Image | Has `odh-elyra`? | Purpose |
|------|-------|------------------|---------|
| **Editor / submitter** | `jupyter/datascience` workbench | **Yes** | Pipeline editor in JupyterLab, submit to DSPA |
| **Executor** | `runtimes/datascience` runtime | **No** | Run individual pipeline node notebooks in KFP pods |

The "wiring" between them is:
- **Build/registry**: runtime ImageStream metadata in `manifests/`
- **Cluster**: notebook controller ConfigMap/Secret mounts
- **Startup**: `setup-elyra.sh` in the workbench
- **Execution**: KFP/Argo scheduling runtime pods with `bootstrapper.py`

Other workbench flavors (PyTorch, TensorFlow, TrustyAI, etc.) also copy `setup-elyra.sh` and pull `odh-notebooks-meta-jupyterlab-datascience-deps`, so they get Elyra too — but the **runtime side** has matching images under `runtimes/pytorch`, `runtimes/tensorflow`, etc., each with their own `odh-notebooks-meta-runtime-elyra-deps`.

atheo89: Also to give some more clarity, check this repo tag: https://github.com/opendatahub-io/notebooks/blob/v1.40.0/runtimes/datascience/ubi9-python-3.12/pyproject.toml 
this tag is a snapshot before we introduce the dependencies directory layering (for better package managing)
On the linked file you may see what **exactly** we install on a runtime (this one is the datascience runtime pyproject) From that moment we don't have add anything new on the runtime architecture

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+set -Eeuxo pipefail
```
ide-developer: @coderabbitai Because this file is sourced from `start-notebook.sh`, these strict shell options leak into the caller and force unrelated `${VAR:-}` guards later in notebook startup. A simpler shape here would be to snapshot/restore shell options (or execute this helper as a subprocess) so the Kale setup stays self-contained.

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
```diff
@@ -0,0 +1,105 @@
...
+        elyra_config_path: Path to Elyra runtime config JSON file
+
+    Returns:
+        True if configuration was saved successfully, False otherwise
+    """
+    if not elyra_config_path:
+        print("No Elyra runtime config found")
+        return False
+
+    try:
```
ide-developer, line 28: @coderabbitai This function would be much easier to reason about if the Elyra->Kale translation were a pure helper that returned a dict, and the CLI wrapper handled file I/O, `save_config()`, and exit status. Right now one broad `except Exception` covers parsing, env lookup, mapping, import, and save, which makes the control flow and tests more complex than they need to be.

ide-developer (commented):
@coderabbitai Adding simplify-focused inline notes.

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Runtime configuration for Kubeflow Kale JupyterLab extension
+# This script configures Kale to connect to KFP by reading Elyra runtime config
+# Note: The extension is disabled by default at build time (see Dockerfile)
+
+# Read Elyra config and copy the relevant information to Kale config
+# Extract KFP configuration from Elyra runtime configs if available
+if [ "$(ls -A /opt/app-root/runtimes/ 2>/dev/null)" ]; then
```
ide-developer, line 10: @coderabbitai One simplify-oriented follow-up here: since the script already prefers `Pipeline.json`, the discovery path can probably collapse to `ELYRA_RUNTIME_CONFIG=/opt/app-root/runtimes/..data/Pipeline.json`, then fall back to the first `*.json`, then a single `[[ -f ... ]]` guard. That removes the leading `ls`, the empty-string sentinel, and the extra `-n` check without changing behavior.

ide-developer (commented):
@coderabbitai Adding the runtime-config simplification note as a standalone inline comment.

dependencies/odh-notebooks-meta-runtime-kale-deps/pyproject.toml
hmtosi: @atheo89 thank you for your suggestion.
As coderabbit says, the reason for introducing this file for the dependencies is because there are two architecture layers - the Jupyterlab layer and the runtime layer. 

The current pattern follows Elyra's implementation and gives kale more independence, as it is not reliant on transitive dependencies.

However, I believe I could remote the meta package and just have `kubeflow-kale[jupyter]` rely on transitive dependencies. This would diverge from Elyra's architecture and could cause us to lose a dependency in the future.

Which implementation do you prefer?

runtimes/datascience/ubi9-python-3.12/pyproject.toml
hmtosi: @daniellutz and @harshad16 what are your thoughts?

jupyter/minimal/ubi9-python-3.12/start-notebook.sh
```diff
@@ -8,28 +8,32 @@ if [ -f "${SCRIPT_DIR}/utils/setup-elyra.sh" ]; then
...
+if [ -n "${NOTEBOOK_ROOT_DIR:-}" ]; then
     NOTEBOOK_PROGRAM_ARGS+="--ServerApp.root_dir=${NOTEBOOK_ROOT_DIR} "
 else
     NOTEBOOK_PROGRAM_ARGS+="--ServerApp.root_dir=${HOME} "
 fi
 
 # Add additional arguments if NOTEBOOK_ARGS variable is defined
-if [ -n "${NOTEBOOK_ARGS}" ]; then
+if [ -n "${NOTEBOOK_ARGS:-}" ]; then
     NOTEBOOK_PROGRAM_ARGS+=${NOTEBOOK_ARGS}
```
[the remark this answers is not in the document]
hmtosi: @coderabbitai I believe I have addressed this, please check

tests/unit/test_configure_kale_from_elyra.py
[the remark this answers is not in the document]
hmtosi: @coderabbitai I believe I have addressed this using Option A, please check

jupyter/datascience/ubi9-python-3.12/setup-kale.sh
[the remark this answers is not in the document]
hmtosi: @coderabbitai — I looked at this with Claude and we have concerns about adding URL validation here:

The legitimate use case requires internal IPs:

KFP endpoints are typically cluster-internal: http://ml-pipeline:8080, http://10.x.x.x:8888, http://ds-pipeline.namespace.svc.cluster.local
Blocking private IP ranges would break the primary use case
Allowing internal IPs defeats SSRF protection
The threat model doesn't justify validation:

api_endpoint comes from DSPA CR / ds-pipeline-config Secret (admin-controlled via RBAC)
An attacker with write access to DSPA already has cluster admin privileges → can directly access metadata endpoints, deploy pods, read secrets
SSRF via api_endpoint is not their easiest or most direct attack path
Minimal validation that doesn't break functionality:
At most, we could add scheme validation to catch malformed URLs (accidents, not attacks):


from urllib.parse import urlparse
parsed = urlparse(kale_config.get('host', ''))
if parsed.scheme not in ('http', 'https'):
    return False  # Invalid URL format
But this prevents accidents, not attacks. An attacker with DSPA write access can still inject http://169.254.169.254 (which has a valid http:// scheme).

Can you re-review whether SSRF protection makes sense here, given that:

The security boundary is RBAC (who can write DSPA), not URL validation
Legitimate internal cluster IPs need to be allowed
The attack requires privileges higher than the payload itself

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
hmtosi: @coderabbitai I believe I have addressed this, please confirm

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
hmtosi: @coderabbitai I believe I have addressed this, please confirm

jiridanek: @coderabbitai recheck

runtimes/datascience/ubi9-python-3.12/pyproject.toml
jiridanek: @coderabbitai what do _you_ think

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
hmtosi: @coderabbitai is this in scope for this PR? Is the PR acceptable without making this change?

jupyter/minimal/ubi9-python-3.12/start-notebook.sh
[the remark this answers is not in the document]
hmtosi: @coderabbitai I believe I have addressed this, please check

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
jiridanek: @coderabbitai what about adding some approval tests? An approval-style version would make the Elyra -> Kale mapping explicit by snapshotting the full transformed output, instead of asserting one field at a time.

Today the transformation is encoded here:

```37:83:jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
        # 1. host: Use api_endpoint (required field)
        kale_config['host'] = metadata.get('api_endpoint', '')

        # 2. namespace: Map user_namespace
        if metadata.get('user_namespace'):
            kale_config['namespace'] = metadata['user_namespace']

        # 3. auth_type and auth_config: Map Elyra auth_type to Kale auth_type
        elyra_auth_type = metadata.get('auth_type', 'KUBERNETES_SERVICE_ACCOUNT_TOKEN')

        if elyra_auth_type == 'NO_AUTHENTICATION':
            kale_config['auth_type'] = None
            kale_config['auth_config'] = {}

        elif elyra_auth_type == 'KUBERNETES_SERVICE_ACCOUNT_TOKEN':
            kale_config['auth_type'] = 'kubernetes_service_account_token'
            kale_config['auth_config'] = {
                'token_path': '/var/run/secrets/kubernetes.io/serviceaccount/token'
            }

        elif elyra_auth_type == 'EXISTING_BEARER_TOKEN':
            kale_config['auth_type'] = 'existing_bearer_token'
            if metadata.get('api_password'):
                kale_config['auth_config'] = {'env_var': 'KF_PIPELINES_TOKEN'}
            else:
                kale_config['auth_config'] = {}

        elif elyra_auth_type in ['DEX_STATIC_PASSWORDS', 'DEX_LDAP', 'DEX_LEGACY']:
            kale_config['auth_type'] = 'dex'
            if metadata.get('api_username') and metadata.get('api_password'):
                kale_config['auth_config'] = {
                    'env_var_username': 'KF_PIPELINES_USERNAME',
                    'env_var_password': 'KF_PIPELINES_PASSWORD'
                }
            else:
                kale_config['auth_config'] = {}

        ssl_cert_path = os.environ.get('KF_PIPELINES_SSL_SA_CERTS',
                                         '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt')
        if os.path.exists(ssl_cert_path):
            kale_config['ssl_ca_cert'] = ssl_cert_path
```

What I would write is one approval test that captures the whole mapping matrix in a single approved JSON file:

```python
from __future__ import annotations

import importlib.util
import json
import sys
from pathlib import Path
from unittest.mock import MagicMock

import pytest

_REPO_ROOT = Path(__file__).resolve().parents[2]
_MODULE_PATH = (
    _REPO_ROOT
    / "jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py"
)
_SPEC = importlib.util.spec_from_file_location("configure_kale_from_elyra", _MODULE_PATH)
assert _SPEC is not None
assert _SPEC.loader is not None
kale_config = importlib.util.module_from_spec(_SPEC)
_SPEC.loader.exec_module(kale_config)

_APPROVED = Path(__file__).with_name("test_configure_kale_from_elyra.approved.json")


def write_elyra_config(tmp_path: Path, metadata: dict) -> Path:
    config_file = tmp_path / "Pipeline.json"
    config_file.write_text(json.dumps({"metadata": metadata}, indent=2), encoding="utf-8")
    return config_file


def run_case(
    case_dir: Path,
    monkeypatch: pytest.MonkeyPatch,
    metadata: dict,
    *,
    cert_exists: bool = True,
) -> dict:
    saved = {"config": None}

    def save_config(config):
        saved["config"] = config

    mock_kfp_server_config = MagicMock()
    mock_kfp_server_config.save_config = save_config

    mock_kale_config = MagicMock()
    mock_kale_config.kfp_server_config = mock_kfp_server_config

    monkeypatch.setitem(sys.modules, "kale", MagicMock())
    monkeypatch.setitem(sys.modules, "kale.config", mock_kale_config)
    monkeypatch.setattr("builtins.print", lambda *args, **kwargs: None)
    monkeypatch.setattr(
        "os.path.exists",
        lambda path: cert_exists
        and path == "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
    )

    case_dir.mkdir()
    elyra_config = write_elyra_config(case_dir, metadata)
    success = kale_config.configure_kale_from_elyra(str(elyra_config))

    return {
        "success": success,
        "saved_config": saved["config"],
    }


def test_elyra_to_kale_transformation_approval(
    tmp_path: Path,
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    actual = {
        "kubernetes_service_account": run_case(
            tmp_path / "kubernetes_service_account",
            monkeypatch,
            {
                "api_endpoint": "http://ml-pipeline:8080",
                "user_namespace": "team-a",
                "auth_type": "KUBERNETES_SERVICE_ACCOUNT_TOKEN",
            },
        ),
        "no_authentication": run_case(
            tmp_path / "no_authentication",
            monkeypatch,
            {
                "api_endpoint": "http://ml-pipeline:8080",
                "user_namespace": "team-a",
                "auth_type": "NO_AUTHENTICATION",
            },
        ),
        "existing_bearer_token": run_case(
            tmp_path / "existing_bearer_token",
            monkeypatch,
            {
                "api_endpoint": "http://ml-pipeline:8080",
                "user_namespace": "team-a",
                "auth_type": "EXISTING_BEARER_TOKEN",
                "api_password": "secret-token",
            },
        ),
        "dex_static_passwords": run_case(
            tmp_path / "dex_static_passwords",
            monkeypatch,
            {
                "api_endpoint": "http://ml-pipeline:8080",
                "user_namespace": "team-a",
                "auth_type": "DEX_STATIC_PASSWORDS",
                "api_username": "admin",
                "api_password": "password123",
            },
        ),
        "missing_api_endpoint": run_case(
            tmp_path / "missing_api_endpoint",
            monkeypatch,
            {
                "user_namespace": "team-a",
                "auth_type": "NO_AUTHENTICATION",
            },
        ),
    }

    approved = json.loads(_APPROVED.read_text(encoding="utf-8"))
    assert actual == approved
```

And the approved file would be the real documentation of the transformation:

```json
{
  "kubernetes_service_account": {
    "success": true,
    "saved_config": {
      "host": "http://ml-pipeline:8080",
      "namespace": "team-a",
      "auth_type": "kubernetes_service_account_token",
      "auth_config": {
        "token_path": "/var/run/secrets/kubernetes.io/serviceaccount/token"
      },
      "ssl_ca_cert": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
    }
  },
  "no_authentication": {
    "success": true,
    "saved_config": {
      "host": "http://ml-pipeline:8080",
      "namespace": "team-a",
      "auth_type": null,
      "auth_config": {},
      "ssl_ca_cert": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
    }
  },
  "existing_bearer_token": {
    "success": true,
    "saved_config": {
      "host": "http://ml-pipeline:8080",
      "namespace": "team-a",
      "auth_type": "existing_bearer_token",
      "auth_config": {
        "env_var": "KF_PIPELINES_TOKEN"
      },
      "ssl_ca_cert": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
    }
  },
  "dex_static_passwords": {
    "success": true,
    "saved_config": {
      "host": "http://ml-pipeline:8080",
      "namespace": "team-a",
      "auth_type": "dex",
      "auth_config": {
        "env_var_username": "KF_PIPELINES_USERNAME",
        "env_var_password": "KF_PIPELINES_PASSWORD"
      },
      "ssl_ca_cert": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
    }
  },
  "missing_api_endpoint": {
    "success": false,
    "saved_config": null
  }
}
```

Why this is clearer than the current test:
- It shows the whole transformation in one place.
- It documents current behavior, not intended behavior.
- Reviewers can diff one approved JSON file and immediately see what changed in the mapping logic.

If I were doing this for real, I’d keep the current small edge-case tests, but replace most of the field-by-field auth mapping tests with this one approval snapshot. That would make the transformation much easier to audit, @coderabbitai.

jiridanek: @coderabbitai The transform reads an Elyra runtime config file whose useful data lives under a top-level `metadata` object, and turns that into the Kale KFP server config that gets saved by `kfp_server_config.save_config()` in `jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py`.

A representative input JSON would look like this:

```json
{
  "metadata": {
    "api_endpoint": "http://ml-pipeline:8080",
    "user_namespace": "my-namespace",
    "auth_type": "KUBERNETES_SERVICE_ACCOUNT_TOKEN"
  }
}
```

With the current implementation, the likely saved Kale config would look like this:

```json
{
  "host": "http://ml-pipeline:8080",
  "namespace": "my-namespace",
  "auth_type": "kubernetes_service_account_token",
  "auth_config": {
    "token_path": "/var/run/secrets/kubernetes.io/serviceaccount/token"
  },
  "ssl_ca_cert": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
}
```

A bearer-token example would be:

Input:

```json
{
  "metadata": {
    "api_endpoint": "http://ml-pipeline:8080",
    "user_namespace": "my-namespace",
    "auth_type": "EXISTING_BEARER_TOKEN",
    "api_password": "secret-token-value"
  }
}
```

Likely output:

```json
{
  "host": "http://ml-pipeline:8080",
  "namespace": "my-namespace",
  "auth_type": "existing_bearer_token",
  "auth_config": {
    "env_var": "KF_PIPELINES_TOKEN"
  },
  "ssl_ca_cert": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
}
```

And a DEX-style example would be:

Input:

```json
{
  "metadata": {
    "api_endpoint": "http://ml-pipeline:8080",
    "user_namespace": "my-namespace",
    "auth_type": "DEX_STATIC_PASSWORDS",
    "api_username": "admin",
    "api_password": "password123"
  }
}
```

Likely output from the current code:

```json
{
  "host": "http://ml-pipeline:8080",
  "namespace": "my-namespace",
  "auth_type": "dex",
  "auth_config": {
    "env_var_username": "KF_PIPELINES_USERNAME",
    "env_var_password": "KF_PIPELINES_PASSWORD"
  },
  "ssl_ca_cert": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
}
```

A couple of important details:
- If `user_namespace` is missing, `namespace` is omitted.
- If `api_endpoint` is missing or empty, the script returns `False` and does not save config.
- `ssl_ca_cert` is only included if the cert path exists.
- The current code uses `api_endpoint`, not `public_api_endpoint`.

If you want, @coderabbitai I can also show the mapping as a compact “input field -> output field” cheat sheet.

jiridanek (approved):
this has great promise and I believe that when you're confident with the code, you should go ahead and merge it

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
jiridanek: @coderabbitai is https://pypi.org/project/inline-snapshot/ any good?

jiridanek: @coderabbitai should @hmtosi read https://blog.janestreet.com/the-joy-of-expect-tests/?

runtimes/datascience/ubi9-python-3.12/pyproject.toml
harshad16: Sharing my understanding , 
Runtime-image are used a image for pipeline task container runs, as these are subset unit, which executes the task step detail, the module required are only necessary.
with that said, in past we didn't install `elyra` in runtime image, as we didn't know of any scenario, where a user is interested in running a pipeline from within a pipeline using elyra.

As the pipeline project matured, we noticed if user has a plan to do pipeline from within a pipeline, they rely on `kfp` as they are already pretty well-versed given the complexity they have selected.

Hope this helps with discussion, i cant be make the decision if kale should be installed in runtime.
it can only be answered by finding out, do we have scenario:
 - where a user can run another pipeline from within pipeline using kale 
 or 
 - if user can leverage it in their pipeline step 
 or 
 - if kale requires itself in each pipeline task step run container.


Thanks for the patience with my comment.

dependencies/odh-notebooks-meta-jupyterlab-datascience-deps/pyproject.toml
hmtosi: I have updated the PR to include kale in all Dockerfiles that have elyra present.

jesuino (approved):
From our tests the work done so far are feasible for putting making Kale work on Elyra images. 

It reuses Elyra default configuration and it is connecting automatically right now for the default token authentication used by Elyra. There are multiple other scenarios that the auto connection may not work, but users can always connect using the terminal (which should be good enough for the dev preview). Kale also has plan to include Runtime Configuration to the UI, but ATM it is only possible using its API.

jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py
hmtosi: I have updated the code so that DEX is disabled in dev preview and `EXISTING_BEARER_TOKEN` is not written directly to disk.

hmtosi:
@coderabbitai review

atheo89:
No other obsession by me to keep this pr open let's include it and see the kale in new adventures! 
/lgtm
/approve

PR #3803, opened 2026-06-05
RHAIENG-5321: Align system python3 RPM version across all architectures in runtime CPU images

ysok:
[RHAIENG-5321](https://redhat.atlassian.net/browse/RHAIENG-5321): Align system python3 RPM version across all architectures in runtime CPU images

## Description
Add python3 to runtime CPU BASE_PKGS so all arches upgrade to the lockfile version and fix Conforma rpm_packages.unique_version (s390x vs. others from base image).

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Updated Python runtime container images to include Python 3 in the base package set by default.

### DIFF
--- a/runtimes/datascience/ubi9-python-3.12/Dockerfile.konflux.cpu
+++ b/runtimes/datascience/ubi9-python-3.12/Dockerfile.konflux.cpu
@@ -25,7 +25,7 @@ COPY prefetch-input/rhds/rpms.lock.yaml /dev/null
 # hadolint ignore=DL3041
 RUN /bin/bash <<'EOF'
 set -Eeuxo pipefail
-BASE_PKGS=(perl mesa-libGL skopeo compat-openssl11 openshift-clients)
+BASE_PKGS=(perl mesa-libGL skopeo compat-openssl11 openshift-clients python3)
 if [ "$(uname -m)" = "s390x" ]; then
     BASE_PKGS+=(gcc gcc-c++ make openssl-devel autoconf automake libtool cmake python3-devel pybind11-devel openblas-devel)
 fi
--- a/runtimes/minimal/ubi9-python-3.12/Dockerfile.konflux.cpu
+++ b/runtimes/minimal/ubi9-python-3.12/Dockerfile.konflux.cpu
@@ -25,7 +25,7 @@ COPY prefetch-input/rhds/rpms.lock.yaml /dev/null
 # hadolint ignore=DL3041
 RUN /bin/bash <<'EOF'
 set -Eeuxo pipefail
-BASE_PKGS=(perl mesa-libGL skopeo compat-openssl11 openshift-clients)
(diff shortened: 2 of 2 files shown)


runtimes/datascience/ubi9-python-3.12/Dockerfile.konflux.cpu
jiridanek, line 1: @coderabbitai if a future base image on some arch ships a python3 version newer than what's in rpms.lock.yaml, then dnf install python3 from the lockfile will try to downgrade it, which dnf install won't do
  silently. It would either no-op (keep the newer version) or fail, and either way the arches stay mismatched.

The lockfile pins a specific NVR. If s390x base ships python3-3.12.10 but the lockfile has python3-3.12.9, the install won't downgrade. If x86_64 base already has 3.12.9, it stays. Conforma still sees a
  mismatch.

This AFAIK only works when the base image is behind the lockfile, not ahead of it.

jiridanek (approved):
80% fix is better than no fix at all, so, I don't complain...

although, shouldn't we dnf install the entire `dnf -qa` list? any package could get mismatched in the https://redhat-internal.slack.com/archives/C0B4V6RBWLU/p1780681349382369?thread_ts=1780624091.804719&cid=C0B4V6RBWLU way

runtimes/datascience/ubi9-python-3.12/Dockerfile.konflux.cpu
jiridanek: shouldn't we dnf install the entire dnf -qa list? any package could get mismatched

ysok: we don't want --allow-downgrade, if base image is ahead of the lockfile, then regen the lockfile. We'll need to treat rpms.lock.yaml just like other lock files, need to keep them up todate.

The potential issue in RHDS, that I could see, is that create-rpm-lockfile.sh is using UBI9_IMAGE registry.redhat.io/ubi9:9.6 vs. AIPCC base image is RHEL. We probably need to update create-rpm-lockfile.sh to whatever is used by AIPCC.

jiridanek: @coderabbitai and aipcc should be running the multiarch version mismatch check against their bases, and don't publish on quay.io/aipcc bases with mismatched versions of packages cross archs

atheo89 (approved):

ysok:
The failures on CI are not related. I will merge this PR and let it sync to the 3.5-ea2 branch.

runtimes/datascience/ubi9-python-3.12/Dockerfile.konflux.cpu
ysok: I'll resolve this comment, will follow up if any remained actions.

PR #3812, opened 2026-06-08
RHAIENG-4892: chore(build-args): drive base image updates from versions_config.yml file

atheo89:
Related to: https://redhat.atlassian.net/browse/RHAIENG-4892 

## Description

Introduces a centralized base-image version sync flow driven by `versions_config.yml`. Maintainers edit one config file, run `make sync-build-args-from-versions`, and the repo updates managed `build-args/*.conf` files plus root `Makefile` release defaults.

This replaces manual, per-image BASE_IMAGE edits with a validated, policy-driven workflow for RHDS (Konflux) and ODH base images across CPU, CUDA, and ROCm families.

**What operators can do**
Single entry point for version updates
- Edit `versions_config.yml` at the repo root.
- Run `make sync-build-args-from-versions` to apply changes across jupyter/, runtimes/, and codeserver/.
- Use `--dry-run` to preview diffs or `--check` to fail when the repo is out of sync.

For more info read its documentation file: [base_image_versions_update_configuration.md](https://github.com/atheo89/notebooks/blob/cetralized-version/docs/base_image_versions_update_configuration.md)
## How Has This Been Tested?

**manual user check/understanding:**
Update the `versions_config.yml` version command with desired versions and run the `make sync-build-args-from-versions` 

**embedded test:**
`./uv run pytest tests/unit/scripts/test_update_build_args_from_versions.py -q`

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added `gmake sync-build-args-from-versions` to sync `BASE_IMAGE`/release build-args from a central `versions_config.yml`, with argument validation and `--dry-run`/`--check` diff output.
* **Documentation**
  * Expanded the README with an “Update base image versions” section and added a dedicated how-to guide and troubleshooting notes.
* **Chores**
  * Added/updated `versions_config.yml`, refreshed guidance comments across many build-arg configs, and updated one ROCm runtime `BASE_IMAGE`.
* **Tests**
  * Added unit tests for the sync/update flow and adjusted image-variant checks to skip accelerator validation when GPU stable versions can’t be determined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -439,12 +439,15 @@ refresh-lock-files:
 #   - Registry access for quay.io/aipcc/base-images when syncing RHDS build args
 # ======================================================================================
 SYNC_BUILD_ARGS_ARGS ?=
+SYNC_BUILD_ARGS_ALLOWED := --dry-run --check
+SYNC_BUILD_ARGS_INVALID := $(strip $(filter-out $(SYNC_BUILD_ARGS_ALLOWED),$(value SYNC_BUILD_ARGS_ARGS)))
 .PHONY: sync-build-args-from-versions
 sync-build-args-from-versions:
+	$(if $(SYNC_BUILD_ARGS_INVALID),$(error Invalid SYNC_BUILD_ARGS_ARGS token(s): $(SYNC_BUILD_ARGS_INVALID) (allowed: $(SYNC_BUILD_ARGS_ALLOWED))))
 	@echo "==================================================================="
 	@echo "🔁 Syncing build-args BASE_IMAGE values from versions_config.yml"
 	@echo "==================================================================="
-	@cd $(ROOT_DIR) && ./uv run scripts/update_build_args_from_versions.py $(SYNC_BUILD_ARGS_ARGS)
+	@cd "$(ROOT_DIR)" && ./uv run scripts/update_build_args_from_versions.py $(value SYNC_BUILD_ARGS_ARGS)
 
 # ======================================================================================
 #   gmake update-imagestream-annotations
--- a/Makefile
+++ b/Makefile
@@ -434,12 +434,13 @@ refresh-lock-files:
 #   gmake sync-build-args-from-versions
 #   gmake sync-build-args-from-versions SYNC_BUILD_ARGS_ARGS=--check
 #   gmake sync-build-args-from-versions SYNC_BUILD_ARGS_ARGS=--dry-run
+#   gmake sync-build-args-from-versions SYNC_BUILD_ARGS_ARGS=--rhds-stable-repo-override=cuda=quay.io/<org>/cuda-el9.6
 # Prerequisites:
 #   - skopeo on PATH (RHDS tag resolution uses skopeo list-tags)
 #   - Registry access for quay.io/aipcc/base-images when syncing RHDS build args
 # ======================================================================================
 SYNC_BUILD_ARGS_ARGS ?=
-SYNC_BUILD_ARGS_ALLOWED := --dry-run --check
+SYNC_BUILD_ARGS_ALLOWED := --dry-run --check --rhds-stable-repo-override=%
 SYNC_BUILD_ARGS_INVALID := $(strip $(filter-out $(SYNC_BUILD_ARGS_ALLOWED),$(value SYNC_BUILD_ARGS_ARGS)))
 .PHONY: sync-build-args-from-versions
 sync-build-args-from-versions:
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ endef
 #######################################        Build helpers                 #######################################
 
 # https://stackoverflow.com/questions/78899903/how-to-create-a-make-target-which-is-an-implicit-dependency-for-all-other-target
-skip-init-for := all-images deploy% undeploy% test% validate% refresh-lock-files sync-commit-env-files update-imagestream-annotations refresh-imagestream-metadata scan-image-vulnerabilities print-release
+skip-init-for := all-images deploy% undeploy% test% validate% refresh-lock-files sync-build-args-from-versions sync-commit-env-files update-imagestream-annotations refresh-imagestream-metadata scan-image-vulnerabilities print-release
 ifneq (,$(filter-out $(skip-init-for),$(MAKECMDGOALS) $(.DEFAULT_GOAL)))
 $(SELF): bin/buildinputs
 endif
@@ -427,6 +427,24 @@ refresh-lock-files:
 		export PIP_LOCK_EXTRA_INDEX_URL="$(PIP_EXTRA_INDEX_URL)" && \
 		unset UV_EXTRA_INDEX_URL PIP_EXTRA_INDEX_URL && \
 		uv run scripts/pylocks_generator.py "$(INDEX_MODE)" $(DIR)
+
+# ======================================================================================
+# Sync build-args BASE_IMAGE values from versions_config.yml
+# Usage examples:
+#   gmake sync-build-args-from-versions
+#   gmake sync-build-args-from-versions SYNC_BUILD_ARGS_ARGS=--check
+#   gmake sync-build-args-from-versions SYNC_BUILD_ARGS_ARGS=--dry-run
+# Prerequisites:
+#   - skopeo on PATH (RHDS tag resolution uses skopeo list-tags)
+#   - Registry access for quay.io/aipcc/base-images when syncing RHDS build args
+# ======================================================================================
+SYNC_BUILD_ARGS_ARGS ?=
+.PHONY: sync-build-args-from-versions
+sync-build-args-from-versions:
+	@echo "==================================================================="
+	@echo "🔁 Syncing build-args BASE_IMAGE values from versions_config.yml"
+	@echo "==================================================================="
+	@cd $(ROOT_DIR) && ./uv run scripts/update_build_args_from_versions.py $(SYNC_BUILD_ARGS_ARGS)
 
 # ======================================================================================
 #   gmake update-imagestream-annotations
--- a/scripts/update_build_args_from_versions.py
+++ b/scripts/update_build_args_from_versions.py
@@ -332,7 +332,9 @@ def normalize_gpu_flavor_config(
     release: ReleaseConfig,
 ) -> tuple[dict[str, Any], str]:
     context = f"root.artifacts.base_image.{accelerator}.{flavor}"
-    mapping = validate_expected_mapping_keys(flavor_policy, {"acc_version", "rhds", "odh"}, context, required_keys={"rhds", "odh"})
+    mapping = validate_expected_mapping_keys(
+        flavor_policy, {"acc_version", "rhds", "odh"}, context, required_keys={"rhds", "odh"}
+    )
 
     rhds_policy = mapping["rhds"]
     odh_policy = mapping["odh"]
@@ -367,7 +369,9 @@ def normalize_gpu_flavor_config(
         ]
         if "acc_version" in normalized_rhds:
             shared_versions.append(
-                validate_version_value(normalized_rhds["acc_version"], release, f"{context}.rhds.acc_version", "acc_version")
+                validate_version_value(
+                    normalized_rhds["acc_version"], release, f"{context}.rhds.acc_version", "acc_version"
+                )
             )
 
         normalized_versions = {normalize_stream_version(version) for version in shared_versions}
@@ -390,7 +394,9 @@ def normalize_gpu_flavor_config(
             "acc_version",
         )
         if normalize_stream_version(distribution_version) != shared_version_normalized:
-            raise ValueError(f"Nested {distribution} acc_version at {context}.{distribution} must match shared acc_version")
+            raise ValueError(
+                f"Nested {distribution} acc_version at {context}.{distribution} must match shared acc_version"
+            )
 
     if scalar_to_string(normalized_rhds.get("channel", "")) == "fast":
         normalized_rhds["acc_version"] = shared_version
@@ -594,11 +600,15 @@ def extract_image_labels(config_payload: dict[str, Any]) -> dict[str, str]:
     if isinstance(config, dict):
         config_labels = config.get("Labels")
         if isinstance(config_labels, dict):
-            labels.update({key: value for key, value in config_labels.items() if isinstance(key, str) and isinstance(value, str)})
+            labels.update(
+                {key: value for key, value in config_labels.items() if isinstance(key, str) and isinstance(value, str)}
+            )
 
     root_labels = config_payload.get("Labels")
     if isinstance(root_labels, dict):
-        labels.update({key: value for key, value in root_labels.items() if isinstance(key, str) and isinstance(value, str)})
+        labels.update(
+            {key: value for key, value in root_labels.items() if isinstance(key, str) and isinstance(value, str)}
+        )
     return labels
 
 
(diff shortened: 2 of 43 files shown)


Makefile
```diff
@@ -428,6 +428,24 @@ refresh-lock-files:
...
+#   - skopeo on PATH (RHDS tag resolution uses skopeo list-tags)
+#   - Registry access for quay.io/aipcc/base-images when syncing RHDS build args
+# ======================================================================================
+SYNC_BUILD_ARGS_ARGS ?=
+.PHONY: sync-build-args-from-versions
+sync-build-args-from-versions:
+	@echo "==================================================================="
+	@echo "🔁 Syncing build-args BASE_IMAGE values from versions_config.yml"
+	@echo "==================================================================="
+	@cd $(ROOT_DIR) && ./uv run scripts/update_build_args_from_versions.py $(SYNC_BUILD_ARGS_ARGS)
```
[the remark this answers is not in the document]
jiridanek: > As per coding guidelines: **/Makefile: “Quote shell variables in targets” and **: “REVIEW PRIORITIES: Security vulnerabilities … first.”

@coderabbitai what coding guidelines? where do we have any such guideline?

scripts/update_build_args_from_versions.py
jiridanek, line 1: @coderabbitai can we use pydantic to make this script shorter? and to have fewer things to test in tests/unit/scripts/test_ for it?

jiridanek: @coderabbitai any other opportunities for making the script more succinct? This is a project utility, it should be easy to understand.

jiridanek: @coderabbitai also, ansi color formatting is I think duplicated with something else in the project

and the image parsing, like jupyter-minimal- and so on is already somewhere else too, but possibly there it is in bash

atheo89: I looked into a Pydantic rewrite and it is not so easy as it is proposed in the thread :( 

The easy part of this utility is basic schema-shape validation. The harder part is the repo-specific normalization and cross-field rules: shared GPU `acc_version`, RHDS `fast` vs `stable`, and preserving the current validation semantics. A realistic Pydantic model still needs custom parent-level validators for those rules, so the core complexity does not really go away; it mostly moves.
Lets keep the lower-risk path: keep behavior the same and simplify the structure in place. I the latest updated there are some of the feasible suggestions from this thread.

atheo89:
Any Build failures doesn't related to this PR as this code doesn't touch any image functionality

atheo89:
This GHA is out of scope on this PR [Code static analysis / check-generated-code (pull_request)](https://github.com/opendatahub-io/notebooks/actions/runs/27602977886/job/81608141899?pr=3812)  a follow up will fix it

tests/test_main.py
```diff
@@ -267,6 +267,8 @@ def test_image_pyprojects(subtests: pytest_subtests.plugin.SubTests, manifests_d
                             continue
                         elif s.get("name") in ("CUDA", "ROCm"):
                             expected_version = get_accelerator_version_for_directory(directory, s["name"])
+                            if expected_version is None:
+                                continue
```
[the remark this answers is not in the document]
ide-developer: @coderabbitai I agree with this one after rereading the latest head and the ticket intent.

Once a GPU target moves to a `*-stable-*` ref, both call sites stop asserting the manifest's `CUDA`/`ROCm` version against the configured shared `acc_version` (`continue` in the main loop, `pytest.skip()` in the variant loop). That leaves exactly the kind of operator-visible drift this sync flow is supposed to prevent.

I don't think the fact that stable tags omit the accelerator stream is enough reason to stop validating it; the expected version still exists in `versions_config.yml`.

scripts/update_build_args_from_versions.py
```diff
@@ -0,0 +1,1273 @@
...
+import re
+import subprocess
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+ROOT_DIR = Path(__file__).resolve().parents[1]
+DEFAULT_CONFIG_PATH = ROOT_DIR / "versions_config.yml"
```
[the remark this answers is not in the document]
ide-developer: @coderabbitai rechecked on the latest head: this one is still live.

`parse_args()` still binds `--config` to `DEFAULT_CONFIG_PATH`, and `main()` still unconditionally does `load_versions_config(args.config)`, so `--root /tmp/other-tree` without `--config` will still apply this checkout's `versions_config.yml` to that other tree.

The new unit coverage only exercises `--root` together with an explicit `--config`, so the broken default remains untested too.

scripts/update_build_args_from_versions.py
ide-developer: @coderabbitai after reviewing this against the actual ticket goals in `RHAIENG-4892` and `RHAIENG-3973`, my take is that the core feature could have been achieved materially simpler.

The justified core is pretty small: validate `versions_config.yml`, discover the managed `build-args/*.conf` files, rewrite `BASE_IMAGE`, and resolve the latest published RHDS tag for each configured family. Most of the bulk seems to come from three extra layers:
1. repo-state-driven bundle/phase inference,
2. stable GPU image introspection to recover CUDA/ROCm stream from container metadata,
3. compatibility normalization for multiple config shapes in a brand-new `schema_version: 1`.

If phase/family selection were explicit in config, or each target simply resolved independently from published tags, I think this likely collapses by something like 3-5x in code/test/docs while staying aligned with the ticket. Right now it reads less like a sync utility and more like a release-policy engine embedded in the repo.

scripts/update_build_args_from_versions.py
```diff
@@ -0,0 +1,1443 @@
...
+                rhds_bundle_phase,
+                stable_repo_overrides,
+            )
+        case "odh":
+            return resolve_odh_base_image(state, release)
+        case _:
+            raise ValueError(f"Unsupported distribution in {state.target.path}: {state.target.distribution}")
+
+
+def read_conf_assignments(text: str) -> dict[str, str]:
```
ide-developer, line 1220: @coderabbitai another simplification/reuse point: this PR grows a new `read_conf_assignments()` / `classify_*()` stack even though the repo already has nearby helpers for almost the same jobs (`scripts/index_url_resolver.py:read_conf_file()/parse_accelerator()/parse_release()`, `manifests/tools/commit_env_refs.py:parse_env_file()`, `tests/manifests.py:extract_metadata_from_path()`).

Pulling one shared build-args / path-metadata helper up would cut code here and reduce the risk that repo-layout knowledge drifts in two places.

scripts/update_build_args_from_versions.py
```diff
@@ -0,0 +1,1443 @@
...
+    if policy.version is None:
+        raise ValueError(f"Missing {policy_version_key(target.accelerator)} for odh origin in {target.path}")
+    if policy.mode == "in-house":
+        return build_odh_in_house_image(target.accelerator, policy.version, release)
+    if policy.mode == "midstream":
+        return build_odh_midstream_image(target.accelerator, policy.version, release)
+    raise ValueError(f"Unsupported odh origin in {target.path}: {policy.mode}")
+
+
+def build_target_base_image(
```
ide-developer, line 1194: @coderabbitai new simplification concern after reading the full ticket context: this is the seam where the implementation starts looking like a release-policy engine rather than a `versions_config.yml -> build-args` sync utility.

Between `determine_rhds_fast_bundle_phase()`, forward/rollback handling, and peer-state inference, the resolver is solving a materially broader problem than `RHAIENG-4892` spells out. If the desired phase/family were explicit in config, or if each target simply resolved independently to the latest published tag for its configured family, this whole path could be much shorter and easier to validate.

Makefile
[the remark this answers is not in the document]
ide-developer: @coderabbitai follow-up after rereading the latest head and the full PR discussion: I don't think this one is actually addressed yet.

The later allowlist only blocks extra whitespace-separated tokens. It still allows shell metacharacters inside the single override token, because `--rhds-stable-repo-override=%` is just a Make pattern and the recipe still expands `$(value SYNC_BUILD_ARGS_ARGS)` directly in the shell.

A concrete example that still passes the filter is:

`SYNC_BUILD_ARGS_ARGS='--rhds-stable-repo-override=cuda=quay.io/example/cuda-el9.6;id'`

That still matches the allowed pattern, and the trailing `;id` is interpreted by the shell at the current recipe line before Python validates the argv.

scripts/update_build_args_from_versions.py
```diff
@@ -0,0 +1,1273 @@
...
+    shared_acc_version: str | None = None
+
+
+def scalar_to_string(value: object) -> str:
+    if isinstance(value, str):
+        return value.strip()
+    if isinstance(value, int | float):
+        return str(value)
+    raise TypeError(f"Unsupported scalar value: {value!r}")
+
```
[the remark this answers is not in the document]
ide-developer: @coderabbitai rechecked on `fe260ce`: the numeric-scalar concern is still real.

`scalar_to_string()` still accepts `int | float`, so unquoted YAML scalars are silently coerced. The concrete failure mode here is not just `25.10 -> 25.1`; it is also `python_version: 3.10 -> \"3.1\"`, which would change path/repository selection instead of failing fast.

Because `versions_config.yml` is meant to be the operator entrypoint, I would still strongly prefer rejecting numeric version scalars entirely and forcing quoted strings.

ide-developer (approved):
@coderabbitai after reading the full PR conversation, the Jira scope in [RHAIENG-4892](https://redhat.atlassian.net/browse/RHAIENG-4892) / [RHAIENG-3973](https://redhat.atlassian.net/browse/RHAIENG-3973), and the latest `fe260ce` head, I think the direction here is good and the operator story is meaningfully better than today's manual fan-out, so I'm approving this.

The part I do want to capture for follow-up is simplification. The ticket-aligned core is small and valuable: make `versions_config.yml` the operator entrypoint, validate it, find the managed `build-args/*.conf` files, rewrite `BASE_IMAGE`, and resolve the latest published RHDS ref for the configured family. That bridge from declared versions to concrete build inputs is exactly the missing piece for the broader automation envisioned in [RHAIENG-3973](https://redhat.atlassian.net/browse/RHAIENG-3973), and I think this PR establishes it.

My main design feedback is that the implementation currently solves a broader problem than [RHAIENG-4892](https://redhat.atlassian.net/browse/RHAIENG-4892) strictly asked for. In addition to the core sync flow, it now owns peer-state-driven bundle/phase inference, stable GPU image introspection to rediscover CUDA/ROCm stream from container metadata, config-shape normalization / compatibility handling, a second repo-layout classification layer, and a very large test/docs surface. That makes the code read less like a sync utility and more like a repo-local release-policy engine.

I think the biggest future win would be to move intent into data instead of inferring it from current repo state. If phase / family / stable-vs-fast selection were explicit in `versions_config.yml` (or in a small fixed target table), each target could resolve independently from published tags with a handful of cached `skopeo list-tags` calls. That would likely remove most of the `determine_rhds_fast_bundle_phase()` / `build_rhds_pinned_tag()` / `resolve_rhds_base_image()` machinery and shrink the associated test matrix substantially. Likewise, if stable GPU refs remain ambiguous, I would rather make that ambiguity explicit in config than inspect candidate images until one exposes the expected accelerator version.

There is also some reuse left on the table. The repo already has Renovate coverage for `build-args/konflux.*.conf` base image updates in `.github/renovate.json5`, and there are nearby helpers for build-args parsing and path metadata in `scripts/index_url_resolver.py`, `manifests/tools/commit_env_refs.py`, and `tests/manifests.py`. Pulling those together over time would make this utility narrower and reduce drift.

So: approval on the overall direction and value, with the inline comments as follow-up cleanup / simplification suggestions. I think this can land and still be made materially smaller and more direct in subsequent iterations.

jiridanek (approved):
/lgtm

scripts/update_build_args_from_versions.py
```diff
@@ -0,0 +1,1443 @@
...
+        log_warning("skopeo inspect --config returned invalid JSON for %s: %s", image, exc, color=warning_color)
+        return None
+
+    if not isinstance(payload, dict):
+        log_warning("skopeo inspect --config returned unexpected payload for %s", image, color=warning_color)
+        return None
+    return payload
+
+
+def inspect_rhds_stable_acc_version(image: str, accelerator: str) -> str | None | object:
```
ide-developer, line 675: @coderabbitai simplification follow-up: this stable-image probing path is a strong sign that the config contract is still too implicit.

At this point the updater is not just mapping declared versions to concrete refs; it is probing candidate images and reverse-engineering CUDA/ROCm stream from env, labels, and history because the tag family itself is underspecified. That is a lot of machinery for a sync utility.

If stable GPU selection needs to stay, I would seriously consider making the needed discriminator explicit in `versions_config.yml` (or a tiny target metadata table) so the updater does not need to inspect container internals to decide which published tag it meant.

scripts/update_build_args_from_versions.py
```diff
@@ -0,0 +1,1443 @@
...
+    return int(phase.removeprefix("ea."))
+
+
+def build_rhds_seed_tag(release_version: str, phase: str | None) -> str:
+    if phase is None:
+        return f"{release_version}-0"
+    return f"{release_version}-{phase}-0"
+
+
+def determine_rhds_fast_bundle_phase(
```
ide-developer, line 960: @coderabbitai this is the other major simplification seam for me.

From here through `build_rhds_pinned_tag()` / `resolve_rhds_base_image()`, the implementation is inferring bundle phase and forward/rollback behavior from peer `BASE_IMAGE` state already checked into the repo. That is materially broader than `RHAIENG-4892`'s stated job of syncing managed build-args from `versions_config.yml` and resolving the latest published RHDS tag.

A more direct design would make the intended family / phase policy explicit in config and then resolve each target independently, instead of deriving policy from the current repo state.

scripts/update_build_args_from_versions.py
```diff
@@ -0,0 +1,1443 @@
...
+                    state.original_text,
+                    build_conf_replacements(
+                        read_conf_assignments(state.original_text), resolved_base_image, config.release
+                    ),
+                ),
+                target=state.target,
+            )
+        )
+
+    makefile = root_dir / "Makefile"
```
ide-developer, line 1351: @coderabbitai scope/simplification note: once `plan_updates()` also starts rewriting root `Makefile` release defaults, the feature grows beyond "sync build-args from `versions_config.yml`" into mutating repo-global release metadata.

That may still be useful, but it widens the blast radius and the amount of behavior this script owns. If the main objective is the `build-args/*.conf` sync flow, I would consider whether Makefile release defaults should stay a separate update step.

tests/unit/scripts/test_update_build_args_from_versions.py
```diff
@@ -0,0 +1,1980 @@
...
+import textwrap
+from typing import TYPE_CHECKING
+
+import pytest
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+def load_updater():
```
ide-developer, line 14: @coderabbitai simplification signal from the test side: this file is now almost a mini framework for the updater, with helper DSLs, stubs, and many permutations of the policy engine.

That is not inherently wrong, but it does suggest the implementation has grown well past a straightforward config->rewrite tool. If the resolver surface were more declarative and narrower, I would expect this test file to collapse substantially rather than needing nearly 2k lines to pin down behavior.

tests/unit/scripts/test_update_build_args_from_versions.py
```diff
@@ -0,0 +1,1980 @@
...
+    config = tmp_path / "versions_config.yml"
+    write_versions_config(config)
+
+    loaded = updater.load_versions_config(config)
+
+    assert loaded.release.full_version == "3.6.0"
+    assert loaded.release.python_version == "3.12"
+
+
+def test_load_versions_config_accepts_shared_gpu_acc_version_layout(tmp_path: Path) -> None:
```
ide-developer, line 328: @coderabbitai another thing the tests make visible: supporting both the legacy nested GPU `acc_version` shape and the new shared flavor-level `acc_version` shape adds a lot of code and test surface for a brand-new `schema_version: 1` interface.

If there isn't a strong migration requirement yet, requiring a single config shape would simplify both the implementation and the tests quite a bit.

atheo89:
Thank for the review :) Yes, I will follow up with simplifications. I will introduce helper functions so they can be used from all the subcomponents. I will start with skopeo tho because we use it in many places, and i will continue with an analysis to cover the rests

PR #3858, opened 2026-06-17
GHA: Update ODH rpms.lock.yaml lock files

ide-developer:
Automated regeneration of `rpms.lock.yaml` from `rpms.in.yaml` for the **ODH** variant.

Updated paths:
- `prefetch-input/odh/rpms.lock.yaml`
- `codeserver/ubi9-python-3.12/prefetch-input/odh/rpms.lock.yaml`


atheo89:
obsolete, this pr should be close. The base images have been updated, so the pylocks should be refresh and eventually the rpms after

PR #3976, opened 2026-06-29
RHAIENG-5136: chore(ci) Add fail_ci_if_error to Codecov actions

mtchoum1:
## Description

Greenify push CI on rhoai-3.5 by fixing the Code static analysis workflow 

pytest-tests failures ([example run](https://github.com/red-hat-data-services/notebooks/actions/runs/28377434702/job/84114767675)):
gen_gha_matrix_jobs tests failed because buildinputs_runner.py tried to podman run ghcr.io/.../buildinputs:main without podman/GHCR auth (exit 125).
Fix: build bin/buildinputs locally in the job and set CI: 'false' so tests use the local binary (same as local dev).
Codecov upload noise on protected branches:

codecov/test-results-action failed with Token required because branch is protected even when tests passed.
Fix: add fail_ci_if_error: false to both Python and Go test-results upload steps (matching coverage upload behavior).

## How Has This Been Tested?

https://github.com/mtchoum1/notebooks/actions/runs/28391710484/job/84120186161

Self checklist (all need to be checked):
- [x] Ensure that you have run `make test` (`gmake` on macOS) before asking for review
- [ ] Changes to everything except `Dockerfile.konflux` files should be done in `odh/notebooks` and automatically synced to `rhds/notebooks`. For Konflux-specific changes, modify `Dockerfile.konflux` files directly in `rhds/notebooks` as these require special attention in the downstream repository and flow to the upcoming RHOAI release.

## Merge criteria:

- [x] The commits are squashed in a cohesive manner and have meaningful messages.
- [x] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [x] The developer has manually tested the changes and verified that the changes work

PR #3977, opened 2026-06-29
RHAIENG-5136: chore(ci) Add fail_ci_if_error to Codecov actions

mtchoum1:
https://redhat-internal.slack.com/archives/C096ZR053RQ/p1782756956346059

## Description

Greenify push CI on rhoai-3.5 by fixing the Code static analysis workflow 

pytest-tests failures ([example run](https://github.com/red-hat-data-services/notebooks/actions/runs/28377434702/job/84114767675)):
gen_gha_matrix_jobs tests failed because buildinputs_runner.py tried to podman run ghcr.io/.../buildinputs:main without podman/GHCR auth (exit 125).
Fix: build bin/buildinputs locally in the job and set CI: 'false' so tests use the local binary (same as local dev).
Codecov upload noise on protected branches:

codecov/test-results-action failed with Token required because branch is protected even when tests passed.
Fix: add fail_ci_if_error: false to both Python and Go test-results upload steps (matching coverage upload behavior).

## How Has This Been Tested?

https://github.com/mtchoum1/notebooks/actions/runs/28391710484/job/84120186161

## Merge criteria:

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -78,6 +78,7 @@ jobs:
         uses: codecov/test-results-action@0fa95f0e1eeaafde2c782583b36b28ad0d8c77d3  # v1.2.1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
 
   go-tests:
     runs-on: ubuntu-26.04
@@ -120,6 +121,7 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: scripts/buildinputs/junit-go.xml
+          fail_ci_if_error: false
 
   code-static-analysis:
     runs-on: ubuntu-26.04


.github/workflows/code-quality.yaml
```diff
@@ -55,10 +55,23 @@ jobs:
...
+        if: ${{ steps.install-deps.conclusion == 'success' && !cancelled() }}
+        run: make bin/buildinputs
+
       - name: Static tests (pytest + Dockerfile alignment)
         run: make test
         if: ${{ steps.install-deps.conclusion == 'success' && !cancelled() }}
         env:
+          # Use the locally built bin/buildinputs instead of pulling the GHCR image
+          # (no podman/GHCR auth in this job; see scripts/buildinputs_runner.py).
+          CI: 'false'
```
jiridanek: let's continue the discussion in

* https://github.com/opendatahub-io/notebooks/pull/3976#discussion_r3493920627

cc @coderabbitai

my opinion is that this only becomes a problem for PRs from forks, only there the check is currently failing

I would not do extra work to support PRs from forks, because they cannot be fully properly tested anyways

so I would not fix this part; it sure deserves better failure message, but not a fix that causes us to have to maintain additional cache for the setup-go action; it's also slower this way than fetching the container

jiridanek (changes requested):

jiridanek (approved):

.github/workflows/code-quality.yaml
jiridanek: create new github issue to implement the fail-on-fork, here, and in any existing gha workflow where we do skip-on-fork, we should instead put a fail-on-fork step that fails with intelligible message

PR #3997, opened 2026-07-01
RHAIENG-5656: fix(codeserver): validate NB_PREFIX before envsubst interpolation

ysok:
https://redhat.atlassian.net/browse/RHAIENG-5656

## Summary
Adds defense-in-depth input validation for `NB_PREFIX` in the codeserver nginx startup script to prevent potential nginx config injection via `envsubst` (CWE-74, FIND-015).

- Validates `NB_PREFIX` against a safe URI-path character set before interpolation
- Exits with an error if invalid characters are detected, without echoing the raw value

## Test plan
- [ ] Legitimate `NB_PREFIX` values (e.g. `/notebook/namespace/name`) pass validation and nginx starts
- [ ] Values with special characters (semicolons, braces, backticks) cause the script to exit with an error

## Summary by CodeRabbit

* **Documentation**
  * Added a new security guidance page explaining workbench trust boundaries and how platform environment variables are handled.
  * Clarified that these environment variables are set by the platform, used during startup, and not validated inside the image.
  * Updated architecture documentation to link to the security guidance and summarize the trust boundary assumptions.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -163,6 +163,12 @@ ImageStream annotations in `manifests/*/base/`. Key annotations include `opendat
 `opendatahub.io/notebook-python-dependencies`. When launching a workbench, the dashboard injects
 the `NOTEBOOK_ARGS` environment variable with OAuth proxy and configuration settings.
 
+### Trust boundaries
+
+Platform env vars (`NB_PREFIX`, `NOTEBOOK_ARGS`) are set by the controller, not by HTTP
+requests. We do not validate them inside the image — see
+[docs/security/workbench-trust-boundaries.md](docs/security/workbench-trust-boundaries.md).
+
 ### Notebook controller (kubeflow)
 
 The [ODH Notebook Controller](https://github.com/opendatahub-io/kubeflow/tree/main/components/odh-notebook-controller)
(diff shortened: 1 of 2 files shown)


codeserver/ubi9-python-3.12/run-nginx.sh
jiridanek, line 1: @coderabbitai critically assess

jiridanek, line 1: @coderabbitai protecting NB_PREFIX alone imo does not really help, unless NOTEBOOK_ARGS (or BASE_URL calculated from NOTEBOOK_ARGS) is also sanitized.

jiridanek (approved):
I don't much like https://github.com/opendatahub-io/notebooks/pull/3997, I'd say better to start writing a threat model. In it, we can say that protecting against env vars being injected into the workbench is pointless, because whoever sets the env vars controls the pod spec anyways. The code to protect NB_PREFIX is not so bad, but imagine having to do the same for NOTEBOOK_ARGS.

codeserver/ubi9-python-3.12/run-nginx.sh
ysok: Added docs/security/workbench-trust-boundaries.md and reverted this code changes.

ysok: Added docs/security/workbench-trust-boundaries.md and reverted this code changes.

jiridanek:
@coderabbitai full review

PR #4008, opened 2026-07-02
RHAIENG-4895: feat(manifests/tools): add ImageStream tag rollout for ODH and RHOAI

atheo89:
https://redhat.atlassian.net/browse/RHAIENG-4895 

## Description

Add `rollout_tag_on_imagestreams.py` to roll workbench ImageStream tags (tag, fetched from versions_config.yml). 
ODH keeps two tags and resolves N-1 digests via skopeo; RHOAI preserves tag history and syncs params.env/commit.env from RHOAI-Build-Config. Extend generate_kustomization.py to discover new rollout keys from ImageStream placeholders. Add shared skopeo_inspect helpers.

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added a rollout utility to update image stream tags and synchronize release metadata automatically.
  * Added target selection (all/ODH/RHOAI) and a dry-run mode.
  * Added improved release image inspection and tag lookup helpers for release-related workflows.
* **Bug Fixes**
  * Improved automatic kustomization/config generation to ensure newly released parameters are discovered and ordered correctly.
  * Preserves existing rollout order while extending version chains using imagestream manifests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,6 @@ dev = [
     # rich powers the local live progress table in ci/check-image-availability.py
     "rich>=14.3.2",
     "inline-snapshot>=0.34.1",
-    "consolio>=0.1.15",
 ]
 
 [tool.uv]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ dev = [
     # rich powers the local live progress table in ci/check-image-availability.py
     "rich>=14.3.2",
     "inline-snapshot>=0.34.1",
+    "consolio>=0.1.15",
 ]
 
 [tool.uv]
--- a/manifests/tools/generate_kustomization.py
+++ b/manifests/tools/generate_kustomization.py
@@ -258,12 +258,23 @@ def _discover_param_pairs_from_imagestreams(
         document = yaml.safe_load(resource_path.read_text())
         if not isinstance(document, dict):
             continue
-        imagestream_name = document.get("metadata", {}).get("name")
-        tags = document.get("spec", {}).get("tags", [])
+        metadata = document.get("metadata")
+        if not isinstance(metadata, dict):
+            continue
+        imagestream_name = metadata.get("name")
+        spec = document.get("spec")
+        if not isinstance(spec, dict):
+            continue
+        tags = spec.get("tags")
         if not isinstance(imagestream_name, str) or not isinstance(tags, list):
             continue
         for tag in tags:
-            from_name = tag.get("from", {}).get("name")
+            if not isinstance(tag, dict):
+                continue
+            from_block = tag.get("from")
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -51,6 +51,7 @@ dev = [
     # rich powers the local live progress table in ci/check-image-availability.py
     "rich>=14.3.2",
     "inline-snapshot>=0.34.1",
+    "consolio>=0.1.15",
```
jiridanek: @coderabbitai we shouldn't have both rich and consolio in the same project

Here are the relevant parts rewritten to use `rich` instead of `consolio`.

### `manifests/tools/rollout_tag_on_imagestreams.py`

Update your imports by removing `consolio` and adding `rich.console`:

```python
# Remove this:
# from consolio import Consolio

# Add this:
from rich.console import Console

```

Then, replace the entire `StepReporter` class with the following rewritten version. It utilizes `rich.console.Console` and its native `status` context manager to replicate the spinner and fallback formatting:

```python
class StepReporter:
    def __init__(self, stream: Any | None = None) -> None:
        self.stream = sys.stdout if stream is None else stream
        self.console = Console(file=self.stream)
        self.is_tty = self.console.is_terminal

    def print_step(self, message: str) -> None:
        if not self.is_tty:
            print(message, file=self.stream, flush=True)
            return
        
        # Equivalent to consolio's "cmp" output style
        self.console.print(f"[bold green]✓[/bold green] {message}")

    @contextmanager
    def running_step(
        self,
        start_message: str,
        done_message: str,
        *,
        animate: bool = False,
        total: int | None = None,
    ) -> Iterator[Callable[[str], None] | None]:
        if total is not None:
            self.print_step(start_message)
            completed = 0

            def track_item(message: str) -> None:
                nonlocal completed
                completed += 1
                detail = f"[{completed}/{total}] {message}"
                if self.is_tty:
                    # Equivalent to consolio's indented "inf" output style
                    self.console.print(f"  [bold cyan]i[/bold cyan] {detail}")
                else:
                    print(f"  {detail}", file=self.stream, flush=True)
                self.stream.flush()

            try:
                yield track_item
            except Exception:
                raise
            else:
                self.print_step(done_message)
            return

        if not self.is_tty or not animate:
            self.print_step(start_message)
            try:
                yield None
            except Exception:
                raise
            else:
                self.print_step(done_message)
            return

        with self.console.status(start_message, spinner="dots"):
            yield None
        self.print_step(done_message)

```

*(Note: Since `rich` is already available in your `pyproject.toml`, you can simply revert the additions of `consolio>=0.1.15` in `pyproject.toml` and `uv.lock` from the diff).*

atheo89: I switched to `rich`

ide-developer (approved):

PR #4047, opened 2026-07-09
RHAIENG-6193: Switch git-clone-oci-ta back to 0.1

ysok:
[RHAIENG-6193](https://redhat.atlassian.net/browse/RHAIENG-6193): Switch git-clone-oci-ta back to 0.1

## Description
(Konflux bot, Jul 9) bumped git-clone-oci-ta 0.1 → 0.2 in .tekton/multiarch-odh-main-combined-pipeline.yaml. Version 0.2 uses konflux-build-cli instead of the old git-clone image and does not smudge LFS files into the clone artifact.

Pin clone back to 0.1

## How Has This Been Tested?

Self checklist (all need to be checked):
- [ ] Ensure that you have run `make test` (`gmake` on macOS) before asking for review
- [ ] Changes to everything except `Dockerfile.konflux` files should be done in `odh/notebooks` and automatically synced to `rhds/notebooks`. For Konflux-specific changes, modify `Dockerfile.konflux` files directly in `rhds/notebooks` as these require special attention in the downstream repository and flow to the upcoming RHOAI release.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

[RHAIENG-6193]: https://redhat.atlassian.net/browse/RHAIENG-6193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Updated a pinned pipeline task reference to a different version for repository cloning.

PR #3653, opened 2026-05-17
build: rewrite prefetch-all.sh for single hermeto invocation

ide-developer:
## Summary
- Rewrite `prefetch-all.sh` to use a single `hermeto fetch-deps` call for all ecosystems (generic, pip, npm, rpm, gomod) instead of separate per-ecosystem downloaders
- Upload unified `bom.json` as GHA artifact (uncompressed, v7 upload-artifact)
- Fixes disk exhaustion from hermeto source duplication (#3641)

## Test plan
- [x] Tested on msg-qe-11 with `jupyter/minimal` (generic + pip + rpm): 68M + 77M + 2.9G, 2814 SBOM components
- [x] Tested on msg-qe-11 with `codeserver` (generic + pip + rpm): 4K + 782M + 2.3G, 2133 SBOM components
- [x] npm auto-detection from Tekton file works (58 entries found for codeserver)
- [ ] Full GHA CI validation (npm + gomod need submodule checkout)

### Problem: serial downloads and disk exhaustion

The previous implementation called separate downloaders per ecosystem:
- **pip**: raw `wget` via inline bash loop in `create-requirements-lockfile.sh` (serial)
- **npm**: `download-npm.sh` via `wget` + `xargs` (limited parallel)
- **rpm**: `hermeto-fetch-rpm.sh` via `podman run` (separate hermeto call)
- **gomod**: `create-go-lockfile.sh` → `hermeto-fetch-gomod.sh` (separate hermeto call)

Each hermeto `podman run` mounted the repo root as `/source:z`. Hermeto internally
copies the entire `/source/` tree ([`resolver.py:58`](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/resolver.py#L58),
[`copy_directory()`](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/utils.py#L171))
into a temp dir. By step 5 (gomod), `cachi2/output/` had ~5 GB from prior steps —
all duplicated in the copy. On dual-disk GHA runners with ~4 GB free on root, this
caused "No space left on device" (#3641).

### Discovery: Konflux uses single hermeto invocation

[...]

PR #4055, opened 2026-07-10
RHAIENG-6193: fix(codeserver): fail fast on un-smudged Git LFS .vsix files

jiridanek:
## Summary

- Konflux `git-clone-oci-ta` 0.2's `clone` step image (`konflux-build-cli`, based on `task-runner`) is missing the `git-lfs` package, so Git-LFS-tracked `.vsix` files under `codeserver/ubi9-python-3.12/utils/` arrive as ~130 byte pointer stubs instead of real archives.
- Previously this surfaced ~1h into the hermetic VS Code compile as a confusing `TypeError: fetch failed` while the build tried (and failed, no network) to download the extension from GitHub.
- This PR adds a fail-fast check in `setup-offline-binaries.sh` so a bad `.vsix` is caught immediately, with an error message that points straight at the root cause, instead of a cryptic network failure an hour later. Also fixes a CodeRabbit-flagged CWE-377 (predictable `/tmp` path) by switching to `mktemp -d`.

## What this PR deliberately does NOT do

- **Does not prefetch `.vsix` via cachi2.** An earlier version of this PR did this; reverted. `git-clone-oci-ta` is already pinned back to 0.1 in [#4047](https://github.com/opendatahub-io/notebooks/pull/4047), so this workaround isn't needed right now, and it would have papered over the real upstream bug instead of tracking it to resolution.
- **Does not add a `fetch-git-lfs` task to the shared `multiarch-odh-main-combined-pipeline.yaml`.** An earlier version of this PR did this too; reverted. Adding permanent pipeline machinery for every component just to work around a temporary upstream regression isn't worth the complexity, especially since a real fix already exists.

## Root cause and fix status (full writeup in [RHAIENG-6193](https://redhat.atlassian.net/browse/RHAIENG-6193))

- Root cause: `task-runner` migrated EL9->EL10 and dropped the `git-lfs` RPM; `konflux-build-cli` inherits this. Confirmed via direct image inspection and an A/B clone test against a real Git LFS repo.
- The fix is already built and published upstream: `konflux-build-cli@sha256:fddbb4c2edb01a3f0065321f84cf2e4d8e483b4149e3d6f78dd0c46e85e35aef` (built from `task-runner` 2.1.0, which added `git-lfs`) correctly smudges LFS content -- verified directly.
- The only missing piece is [konflux-ci/build-definitions#3653](https://github.com/konflux-ci/build-definitions/pull/3653), which bumps `git-clone-oci-ta`'s pinned `konflux-build-cli` digest to the fixed one. It's CI-green and mergeable, but has a "Do NOT merge" hold from a code owner (unrelated reorg reason, predates the LFS fix landing in that PR).
- Once #3653 merges and MintMaker proposes the new digest to us, we can go back to `git-clone-oci-ta` 0.2 directly -- no extra pipeline work needed on our side. This PR's fail-fast check stays as cheap permanent insurance against this class of regression.

## Test plan

- [ ] Merge and confirm `Red Hat Konflux / odh-workbench-codeserver-datascience-cpu-py312-ubi9-on-pull-request` passes on the PR (unaffected by this change since it's diagnostics-only and `git-clone-oci-ta` is on 0.1)

[RHAIENG-6193]: https://redhat.atlassian.net/browse/RHAIENG-6193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of offline extension packages before they are copied or extracted.
  * Added faster, clearer error messages with brief previews when packages are missing or invalid.
  * Reduced the risk of temporary-directory conflicts during extension processing by using a unique workspace.

### DIFF
--- a/codeserver/ubi9-python-3.12/prefetch-input/patches/setup-offline-binaries.sh
+++ b/codeserver/ubi9-python-3.12/prefetch-input/patches/setup-offline-binaries.sh
@@ -116,8 +116,6 @@ populate_vsix() {
         return
     fi
 
-    require_valid_vsix "${vsix_file}"
-
     echo "Extracting ${ext_name} from $(basename "${vsix_file}")..."
     local tmp_dir
     tmp_dir="$(mktemp -d /tmp/vsix-extract.XXXXXX)"
--- a/codeserver/ubi9-python-3.12/prefetch-input/patches/setup-offline-binaries.sh
+++ b/codeserver/ubi9-python-3.12/prefetch-input/patches/setup-offline-binaries.sh
@@ -56,6 +56,23 @@ fi
 export RIPGREP_BINARY_PATH
 echo "Using ripgrep from Python wheel: ${RIPGREP_BINARY_PATH}"
 
+# [RHAIENG-6193] .vsix files under utils/ are Git LFS objects. If the clone task did not
+# smudge LFS (regression in git-clone-oci-ta 0.2), these are ~130 byte pointer stubs, not
+# real ZIPs. Fail fast here instead of ~1h into the VS Code compile when the build later
+# tries (and fails) to download the extension from GitHub with no network access.
+require_valid_vsix() {
+    local f="$1"
+    if [[ ! -f "$f" ]]; then
+        echo "ERROR: ${f} not found" >&2
+        exit 1
+    fi
+    if ! unzip -tqq "$f" >/dev/null 2>&1; then
+        echo "ERROR: ${f} is not a valid .vsix zip (Git LFS pointer not materialized? see RHAIENG-6193)" >&2
+        head -c 200 "$f" >&2 || true
+        exit 1
+    fi
+}
+
 # Setup VSCode marketplace extensions and Node.js binaries from prefetched files.
 # VSCODE_OFFLINE_CACHE is already exported by codeserver-offline-env.sh.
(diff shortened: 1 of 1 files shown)


jiridanek:
* https://github.com/konflux-ci/build-definitions/pull/3653

atheo89 (approved):

jiridanek:
@coderabbitai the new check in this PR is currently failing with

```
time="2026-07-19T18:22:40Z" level=info msg="buildah [stderr] ERROR: /root/codeserver/ubi9-python-3.12/utils/ms-vscode.js-debug-companion.1.1.3.vsix is not a valid .vsix zip (Git LFS pointer not materialized? see RHAIENG-6193)"
time="2026-07-19T18:22:40Z" level=info msg="buildah [stderr] version https://git-lfs.github.com/spec/v1"
time="2026-07-19T18:22:40Z" level=info msg="buildah [stderr] oid sha256:7380a890787452f14b2db7835dfa94de538caf358ebc263f9d46dd68ac52de93"
time="2026-07-19T18:22:40Z" level=info msg="buildah [stderr] size 52784"
time="2026-07-19T18:22:40Z" level=info msg="buildah [stderr] subprocess exited with status 1"
time="2026-07-19T18:22:40Z" level=info msg="buildah [stderr] subprocess exited with status 1"
```

PR #4022, opened 2026-07-09
RHAIENG-6078: Upgrade code-server from v4.106.3 to v4.112.0

ysok:
[RHAIENG-6078](https://redhat.atlassian.net/browse/RHAIENG-6078): Upgrade code-server from v4.106.3 to v4.112.0

## Description
Upgrade code-server from v4.106.3 to v4.112.0 (VS Code 1.112.0, Node 22.22.0) for the ubi9-python-3.12 workbench image. Refresh the prefetch submodule, Tekton prefetch paths, Dockerfile (git-lfs, version bumps), and replace the v4.106.3 patch overlay with v4.112.0 — including regenerated npm lockfiles, pinned rolldown-vite, gulpfile.reh.ts ppc64/s390x targets, and js-debug 1.112.0. Add ppc64/s390x build fixes for VS Code 1.112’s tsgo dependency: fall back to tsc, serialize extension packaging, skip noEmit typecheck under QEMU, and cap Node heap. Validated with sandbox tests and local hermetic builds on arm64 and ppc64le.

## How Has This Been Tested?

Self checklist (all need to be checked):
- [ ] Ensure that you have run `make test` (`gmake` on macOS) before asking for review
- [ ] Changes to everything except `Dockerfile.konflux` files should be done in `odh/notebooks` and automatically synced to `rhds/notebooks`. For Konflux-specific changes, modify `Dockerfile.konflux` files directly in `rhds/notebooks` as these require special attention in the downstream repository and flow to the upcoming RHOAI release.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

[RHAIENG-6078]: https://redhat.atlassian.net/browse/RHAIENG-6078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for the s390x architecture in code-server images and dependency preparation.
  * Updated code-server to v4.112.0, Node.js to 22.22.0, and the Python extension to 2026.4.0.
  * Expanded VS Code extension packaging and offline build support.

* **Bug Fixes**
  * Improved build reliability and memory handling across supported architectures.
[...]

PR #4059, opened 2026-07-11
Fix agent-browser postinstall for s390x hermetic builds

ysok:
## Summary

Cherry-picks the agent-browser fix from [red-hat-data-services/notebooks#2484](https://github.com/red-hat-data-services/notebooks/pull/2484) onto ODH main.

VS Code 1.112 (code-server v4.112.0, merged via #4022) adds `agent-browser` as a devDependency. Its upstream postinstall downloads platform-specific native binaries from GitHub releases — not the npm registry — so cachi2 prefetch alone is insufficient. On s390x/ppc64le the npm tarball has no bundled binary, the download is blocked in hermetic Konflux builds, and npm fails.

This PR patches the cached `agent-browser` npm tarball during `apply-patch.sh` (same pattern as `@vscode/ripgrep`) with an offline-safe postinstall that exits cleanly and falls back to the Node.js CLI.

- Adds `prefetch-input/patches/code-server-v4.112.0/agent-browser/postinstall.js`
- Extends `apply-patch.sh` to overwrite the tarball and strip lockfile integrity

## Test plan

- [ ] Konflux codeserver multi-arch build passes on s390x (and ppc64le)
- [ ] amd64/arm64 builds still succeed (bundled native binary path unchanged)
- [ ] `/build-codeserver` on this PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offline installation of the agent-browser dependency.
  * Prevented installation failures when native binaries are unavailable by using a Node.js fallback.
  * Ensured bundled native binaries are correctly prepared for supported platforms.

### DIFF
--- a/codeserver/ubi9-python-3.12/prefetch-input/patches/apply-patch.sh
+++ b/codeserver/ubi9-python-3.12/prefetch-input/patches/apply-patch.sh
@@ -96,6 +96,29 @@ if [[ "$ARCH" == "amd64" || "$ARCH" == "arm64" || "$ARCH" == "ppc64le" || "$ARCH
         echo "WARNING: ripgrep postinstall not found at ${RIPGREP_PATCHED}"
     fi
 
+    # [AGENT-BROWSER] Overwrite agent-browser postinstall in the cached npm tarball.
+    # The npm tarball bundles native binaries for linux-x64/arm64 only; ppc64le/s390x
+    # are missing. Upstream postinstall then downloads from GitHub releases (not npm),
+    # which fails in hermetic builds. arm64/amd64 succeed because the bundled binary
+    # is found before any download is attempted.
+    AGENT_BROWSER_PATCHED="${CODESERVER_SOURCE_PREFETCH}/agent-browser/postinstall.js"
+    AGENT_BROWSER_TGZ=$(find /cachi2/output/deps/npm -name "agent-browser-*.tgz" -type f 2>/dev/null | head -1)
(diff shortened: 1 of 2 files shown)


ysok:
All good 

image

ide-developer (approved):

ide-developer:
## Review: Fix agent-browser postinstall for s390x hermetic builds

Clean PR — no findings survived verification at medium effort (8 angles, 6 candidates verified).

The implementation correctly follows the established ripgrep patching pattern in `apply-patch.sh`. Three candidates were investigated in depth and refuted:

1. **Lockfile integrity stripping scope** — ripgrep strips integrity from 3 lockfiles (root, remote, build) while agent-browser only strips from 1. Verified that `agent-browser` is a root devDependency only and does not appear in `remote/package-lock.json` or `build/package-lock.json`, so patching one lockfile is correct.

2. **ESM import compatibility** — the replacement `postinstall.js` uses ESM `import` syntax. Verified that the `cp` only overwrites `scripts/postinstall.js` while preserving the original `package.json` (which declares `"type": "module"` in agent-browser@0.16.3), so Node.js correctly resolves the module format.

3. **Duplicated tar extract-repack pattern** — only 2 instances of the pattern exist (ripgrep + agent-browser); the vsce-sign block uses `jq` modification instead of `cp`. Not worth extracting to a shared function at this scale.

Other angles that found nothing actionable: removed-behavior audit (dropped upstream features like `.install-method` writing and Chrome detection are irrelevant in hermetic builds), cross-file tracing (variables `ARCH`/`CODESERVER_SOURCE_PREFETCH` are always set when the block runs), and CLAUDE.md conventions (no project-level CLAUDE.md exists).

@coderabbitai approve

PR #4175, opened 2026-07-27
RHAIENG-3124: fix(scripts): extends update-commit-latest-env.py script to update both streams

atheo89:
Related to: https://redhat.atlassian.net/browse/RHAIENG-3124 

This PR extends `update-commit-latest-env.py` with --variant odh|rhoai|both. ODH uses the params-latest pinned tag when skopeo can inspect it, otherwise falls back to the latest main-<sha> tag. RHOAI resolves a shared rhoai-X.Y tag and writes per-image vcs-ref values to manifests/rhoai/base/commit-latest.env. Split the GitHub workflow into ODH and RHOAI jobs with skopeo and corrected manifest paths.

## How Has This Been Tested?

GHA triggered from this branch https://github.com/opendatahub-io/notebooks/actions/runs/30257743945/job/89950199784 

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Regenerated `commit-latest.env` for both ODH and RHOAI workbench images, with variant selection (`odh`, `rhoai`, or `both`) and optional RHOAI version pinning.
  * Improved tag/version discovery with better fallback behavior and more reliable mapping of commit metadata.
* **Bug Fixes**
  * Updated workflow commit logic to stage and push both ODH/RHOAI commit files only when changes are present.
* **Tests**
  * Added unit tests covering tag parsing, metadata extraction, and error handling.
* **Chores**
  * Adjusted automation to run on a daily schedule (11:00 UTC) and manual dispatch only, with optional registry authentication.


scripts/update-commit-latest-env.py
```diff
@@ -1,142 +1,524 @@
...
+    """Return ``(repository_without_tag, tag)`` for a quay.io image reference."""
+    if ":" not in image_url:
+        return image_url, ""
+    base, tag = image_url.rsplit(":", 1)
+    return base, tag
+
+
+def parse_quay_repository(image: str) -> tuple[str, str]:
+    """Return (namespace, repository) for a quay.io image URL without tag."""
+    if not image.startswith("quay.io/"):
```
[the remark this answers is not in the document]
atheo89: resolved

scripts/update-commit-latest-env.py
```diff
@@ -1,142 +1,528 @@
...
+  exists on quay.io/opendatahub (``vcs-ref`` label from the image config).
+  When the declared tag is missing or unreachable, falls back to the most recently
+  created ``main-<sha>`` tag as this is the builds from main tekton pipeline.
+  Writes ``manifests/odh/base/commit-latest.env``.
+
+RHOAI variant:
+  Probes quay.io/rhoai to find the most recently created ``rhoai-X.Y`` tag
+  Uses that tag across all RHOAI images, extracting vcs-ref from each. Writes
+  ``manifests/rhoai/base/commit-latest.env``.
+  Quay.io images are 1:1 mapped with the RedHat catalogue images, so vcs-ref remains the same.
```
jiridanek, line 19: @atheo89, @coderabbitai how is this going to work for rhoai, who and when will run the update?

jiridanek: > direct commit is authored and pushed as github-actions[bot]

@coderabbitai do the repo settings permit direct commits?

atheo89: Hey Jiri — good question. After this merges, **nobody needs to run updates on a schedule manually**. GitHub Actions (update-commit-latest-env.yaml) **handles it on main as github-actions[bot], daily**, with an optional manual workflow_dispatch if needed. 
Each run refreshes both `manifests/odh/base/commit-latest.env` and `manifests/rhoai/base/commit-latest.env`, and pushes directly to main only when something actually changed.

For RHOAI, the script probes `quay.io/rhoai`, selects the most recently created rhoai-X.Y tag (for example `rhoai-3.5` today), inspects each mapped workbench image at that tag, and writes per-image commit SHAs from the `vcs-ref` label. 

Upstream main remains the source of truth here; RHDS/RHOAI downstream repos pick up these changes through the existing upstream auto-merge flow (a separate daily job in rhods-devops-infra).

jiridanek: what keeps `rhoai-2.25` up to date, if we make a commit there? older but still in-support releases in general

also, as I said, i think direct commits won't work, org settings require a PR

jiridanek (approved):

jiridanek:
## Simplification review

I looked at this PR from four angles — reuse, simplification, efficiency, and altitude — and cross-referenced with the [RHAIENG-3124](https://redhat.atlassian.net/browse/RHAIENG-3124) Jira history. Here are the findings, most impactful first.

---

### 1. `git push` was the original root cause of [RHAIENG-3124](https://redhat.atlassian.net/browse/RHAIENG-3124)

The Jira history documents that commit hashes went stale because **branch protection requires PRs, not direct push** — the `ODH-Require-PRs` org ruleset requires a PR with one approving review on `main` and `rhoai*`.

PR #3451 (still open, by Ayush Tiwari) already fixes this by using `peter-evans/create-pull-request` instead of `git push`, following the `piplock-renewal.yaml` pattern that already exists in this repo.

This PR re-introduces `git push` (line 57 of the workflow). The linked test run (30257743945) succeeded because it ran on the **PR branch** `atheo89/fix-latest-commit`, which has no branch protection — it doesn't prove it works on `main`.

**Suggestion:** Adopt the `create-pull-request` pattern from PR #3451 or `piplock-renewal.yaml`.

### 2. `find_latest_rhoai_tag_by_created` is a 25-line clone

`find_latest_rhoai_tag_by_created` (lines 358-395) duplicates `find_latest_tag_by_skopeo_created` (lines 280-313). The only differences are:
- Pattern is hardcoded (`RHOAI_TAG_PATTERN`) vs. parameterized
- `log_failure` default (`True` vs `False`)

It can be eliminated by adding `log_failure` as a kwarg to the generic function.

Additionally, the RHOAI path skips the Quay API optimization that the ODH path uses. `find_latest_odh_main_tag` tries `quay_list_matching_tags` first (returns `start_ts` with zero subprocesses), falling back to skopeo. The RHOAI path goes straight to skopeo, even though `quay_list_matching_tags` already handles private-repo failures gracefully (`return []` → skopeo fallback). Using the same two-tier strategy for both variants would avoid spawning N skopeo inspects when a single Quay API call suffices.

### 3. `rhoai_image_base`: fragile replacement + duplicated split

Two issues:
- `odh_url.rsplit(":", 1)[0]` is exactly what `parse_image_ref(odh_url)[0]` does — should call it instead of re-splitting
- `.replace("-ubi9", "-rhel9")` is position-insensitive — it replaces the first occurrence anywhere in the string. Today it works because `-ubi9` appears only as a suffix. Anchoring to end-of-string would be more robust: `re.sub(r"-ubi9$", "-rhel9", name)`

### 4. `GITHUB_REF_NAME` sniffing belongs in the workflow

`resolve_rhoai_version_tag` reads `os.environ["GITHUB_REF_NAME"]` directly, coupling the script to GHA internals. If the workflow moves to Tekton/GitLab or is run locally, this hidden fallback silently fails.

The workflow should pass `--rhoai-version-tag "${{ github.ref_name }}"` explicitly when running on an `rhoai-*` branch; the script should only support explicit flag or auto-detect.

### 5. Redundant per-image tag discovery in ODH fallback

When the pinned tag fails for an image, `resolve_odh_vcs_ref` calls `find_latest_odh_main_tag` independently for each of ~11 images. If the Quay API is down, this spawns hundreds of `skopeo inspect` calls per image. All ODH images are built from the same commit, so the latest `main-<sha>` tag could be discovered **once** from a probe image and reused across all images.

### 6. `collect_odh_entries` / `collect_rhoai_entries` share identical boilerplate

Both define an inner `process_one`, `asyncio.gather` over all images, check `if any(result is None)`, return `list(results)`. A small shared helper would eliminate the duplication and ensure both paths stay in sync when error handling evolves.

---

### Skipped (not worth changing in this PR)

- **`_communicate` duplicates `check-image-availability.py`'s timeout+kill pattern**: Valid, but extracting a shared module is outside scope
- **`load_workbench_images` overlaps with `parse_env_file` in `check-image-availability.py`**: Same — slightly different filters, shared module would be a separate effort  
- **Sequential ODH→RHOAI in `main()`**: The immediate-write pattern was a deliberate choice to prevent RHOAI failure from discarding ODH results; concurrent execution would need different error handling

[RHAIENG-3124]: https://redhat.atlassian.net/browse/RHAIENG-3124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RHAIENG-3124]: https://redhat.atlassian.net/browse/RHAIENG-3124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4209, opened 2026-07-30
RHAIENG-6403: ci: grant packages read to Antigravity dispatch review job

### LINKED ISSUE #4211: ci(agentic-reviewer): intermittent podman/crun failure in prepare-review-context on ubuntu-24.04 GHA runners
## Summary

`odh-ci-prepare-review-context` (Antigravity PR review / CI-gated review) intermittently fails on `ubuntu-latest` GitHub Actions runners when `bin/buildinputs` runs `podman run` against `ghcr.io/opendatahub-io/notebooks/buildinputs:main`.

The image pull succeeds; the failure happens immediately on the first container start:

```
Error: OCI runtime error: crun: unknown version specified
```

Reruns on a different runner can succeed with identical code and inputs.

## Impact

- Antigravity **Prepare review context** step fails → entire PR review aborts before the agent runs.
- Observed while validating [#4209](https://github.com/opendatahub-io/notebooks/pull/4209) ([RHAIENG-6403](https://redhat.atlassian.net/browse/RHAIENG-6403)).
- Failure is **not deterministic**.

## Known occurrences

| # | Job | Commit | Runner image | Outcome |
|---|-----|--------|--------------|---------|
| 1 | [90837673039](https://github.com/opendatahub-io/notebooks/actions/runs/30532431812/job/90837673039) | `3436ccd` | **20260726.254.1** | failed |
| 1b (rerun) | [90837867525](https://github.com/opendatahub-io/notebooks/actions/runs/30532431812/job/90837867525) | `3436ccd` | 20260720.247.2 | **success** |
| 2 | [90839563442](https://github.com/opendatahub-io/notebooks/actions/runs/30533005860/job/90839563442) | `5202cd6` | **20260726.254.1** | failed |

Pattern so far: failures correlate with runner image **ubuntu24/20260726.254**; successful rerun used **20260720.247**.

## Exact failing command

```bash
podman run --rm \
  -e TARGETPLATFORM=linux/amd64 \
  -v $GITHUB_WORKSPACE:$GITHUB_WORKSPACE:ro,z \
  -w $GITHUB_WORKSPACE \
  ghcr.io/opendatahub-io/notebooks/buildinputs:main \
  -build-arg=BASE_IMAGE=fake-image \
  jupyter/minimal/ubi9-python-3.12/Dockerfile.konflux.rocm
```

Exit code: **126**

## Environment

- **Workflow:** `.github/workflows/antigravity-pr-review.yml` (via `antigravity-dispatch.yml`)
- **Runner:** `ubuntu-latest` → `ubuntu-24.04`
- **Podman:** stock runner image (no `install-podman-action`; unlike `build-notebooks-TEMPLATE.yaml` on `ubuntu-26.04`)
- **GHCR auth:** works (`packages: read` + `docker/login-action`; pull completes)

## Likely root cause (hypothesis)

`crun: unknown version specified` is a known Podman/crun skew when crun is too old for the OCI runtime spec Podman emits — see [containers/podman#27272](https://github.com/containers/podman/issues/27272) (fixed in crun ≥ 1.14.3).

## Suggested directions

1. Log `podman --version`, `crun --version`, runner image version before `prepare-review-context`.
2. Use `.github/actions/install-podman-action` (or pin crun ≥ 1.14.3) for Antigravity workflows.
3. Retry on exit 126 / `unknown version specified`.
4. Report to [actions/runner-images](https://github.com/actions/runner-images) if `20260726.254` regression is confirmed.

## Related

- PR: https://github.com/opendatahub-io/notebooks/pull/4209
- Jira: https://redhat.atlassian.net/browse/RHAIENG-6403

[RHAIENG-6403]: https://redhat.atlassian.net/browse/RHAIENG-6403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jiridanek:
## Summary

Fixes Antigravity PR review dispatches that fail at workflow validation with **startup failure** before any job runs.

`antigravity-dispatch.yml` calls `antigravity-pr-review.yml` as a reusable workflow. The review workflow declares `packages: read` (needed to pull the private `buildinputs` image from GHCR during `odh-ci-prepare-review-context`), but dispatch only granted `contents`, `issues`, and `pull-requests`. GitHub rejects the nested call:

> The nested job `review` is requesting `packages: read`, but is only allowed `packages: none`.

This one-line change passes `packages: read` through the `workflow_call` boundary.

Related: [RHAIENG-6403](https://redhat.atlassian.net/browse/RHAIENG-6403)

## Problem

After GHCR login was added to `antigravity-pr-review.yml` (so `bin/buildinputs` can pull `ghcr.io/<repo>/buildinputs` while resolving affected image targets), every dispatch-triggered review started failing at parse time.

Observed on RHDS after syncing upstream (example runs):

- https://github.com/red-hat-data-services/notebooks/actions/runs/30530768858
- https://github.com/red-hat-data-services/notebooks/actions/runs/30530784102
- https://github.com/red-hat-data-services/notebooks/actions/runs/30530899916
- https://github.com/red-hat-data-services/notebooks/actions/runs/30530934912

Triggers included `/agy review` issue comments and PR open/edit events — all died in the dispatch → review handoff.

`antigravity-ci-gated-review.yml` was unaffected because it is a top-level workflow with its own `packages: read`. Only the **dispatch → pr-review** path was missing the permission.

## Why `packages: read` is required

`Prepare review context` runs `odh-ci-prepare-review-context`, which calls `gha_pr_changed_files.filter_out_unchanged()` → `bin/buildinputs` per Dockerfile. That tooling may pull container images from GHCR (including the repo's `buildinputs` image). On RHDS that image is private; the workflow logs in with `GITHUB_TOKEN` via `docker/login-action`, which requires `packages: read` on the job token.

Reusable workflows cannot escalate permissions beyond what the caller grants, so dispatch must explicitly include `packages: read` on the job that `uses:` `antigravity-pr-review.yml`.

## Change

Add `packages: read` to the `review` job permissions in `.github/workflows/antigravity-dispatch.yml`.

## Test plan

- [ ] Merge and sync to RHDS (or cherry-pick)
- [ ] Comment `/agy review` on a PR — Antigravity Dispatch should start and the nested review workflow should pass workflow validation
- [ ] Confirm `Prepare review context` completes (GHCR login + buildinputs pull)

[RHAIENG-6403]: https://redhat.atlassian.net/browse/RHAIENG-6403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Updated review and CI workflows to run on Ubuntu 26.04.
  * Added an explicit Podman installation step in review workflows to improve stability.
  * Enhanced review comment attribution by supplying GitHub App identity details (for consistent author naming).
* **Bug Fixes**
  * Fixed login/author resolution to honor `REVIEW_AUTHOR_LOGIN` first, derive from `GITHUB_APP_SLUG` when set, and safely fall back to `github-actions[bot]` when user lookups are blocked.
* **Tests**
  * Updated and expanded unit tests to be deterministic and to verify override precedence and forbidden-endpoint fallback behavior.

### DIFF
--- a/.github/workflows/antigravity-dispatch.yml
+++ b/.github/workflows/antigravity-dispatch.yml
@@ -111,6 +111,7 @@ jobs:
     permissions:
       contents: 'read'
       issues: 'write'
+      packages: 'read'
       pull-requests: 'write'
     with:
       additional_context: '${{ needs.dispatch.outputs.additional_context }}'
--- a/.github/workflows/antigravity-dispatch.yml
+++ b/.github/workflows/antigravity-dispatch.yml
@@ -56,7 +56,7 @@ jobs:
         ) &&
         contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.review.author_association)
       )
-    runs-on: 'ubuntu-latest'
+    runs-on: 'ubuntu-26.04'
     permissions:
       contents: 'read'
     outputs:
--- a/.github/workflows/antigravity-ci-gated-review.yml
+++ b/.github/workflows/antigravity-ci-gated-review.yml
@@ -105,6 +105,7 @@ jobs:
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_TOKEN: ${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}
           PULL_REQUEST_NUMBER: ${{ github.event.workflow_run.pull_requests[0].number }}
+          REVIEW_AUTHOR_LOGIN: ${{ steps.mint_identity_token.outputs.app-slug != '' && format('{0}[bot]', steps.mint_identity_token.outputs.app-slug) || 'github-actions[bot]' }}
           REVIEW_BODY_PATH: 'review-summary-body.md'
           WORKFLOW_RUN_URL: '${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}'
         run: 'uv run --package odh-ci-agent odh-ci-upsert-review-comment'
--- a/.github/workflows/antigravity-ci-gated-review.yml
+++ b/.github/workflows/antigravity-ci-gated-review.yml
@@ -100,6 +100,7 @@ jobs:
         if: |-
           github.event.workflow_run.pull_requests[0] != null
         env:
+          GITHUB_APP_SLUG: ${{ steps.mint_identity_token.outputs.app-slug }}
           GITHUB_REPOSITORY: ${{ github.repository }}
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_TOKEN: ${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}
--- a/.github/workflows/antigravity-ci-gated-review.yml
+++ b/.github/workflows/antigravity-ci-gated-review.yml
@@ -18,7 +18,8 @@ jobs:
       github.event.workflow_run.event == 'pull_request' &&
       github.event.workflow_run.conclusion == 'success' &&
       github.event.workflow_run.head_repository.full_name == github.repository
-    runs-on: 'ubuntu-latest'
+    # ubuntu-latest is ubuntu-24.04; stock podman/crun flakes in prepare-review-context (opendatahub-io/notebooks#4211).
+    runs-on: 'ubuntu-26.04'
     timeout-minutes: 12
     permissions:
       actions: 'read'
@@ -58,6 +59,11 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: 'Install Podman'
+        uses: './.github/actions/install-podman-action'
+        with:
+          platform: 'linux/amd64'
 
       - name: 'Prepare review context'
         if: |-
--- a/.github/workflows/antigravity-ci-summary.yml
+++ b/.github/workflows/antigravity-ci-summary.yml
@@ -39,7 +39,7 @@ jobs:
           github.event.workflow_run.head_repository.full_name == github.repository
         )
       )
-    runs-on: 'ubuntu-latest'
+    runs-on: 'ubuntu-26.04'
     timeout-minutes: 12
     permissions:
       actions: 'read'
--- a/.github/workflows/antigravity-pr-review.yml
+++ b/.github/workflows/antigravity-pr-review.yml
@@ -34,7 +34,8 @@ concurrency:
 
 jobs:
   review:
-    runs-on: 'ubuntu-latest'
+    # ubuntu-latest is ubuntu-24.04; stock podman/crun flakes in prepare-review-context (opendatahub-io/notebooks#4211).
+    runs-on: 'ubuntu-26.04'
     timeout-minutes: 10
     permissions:
       contents: 'read'
@@ -72,6 +73,11 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: 'Install Podman'
+        uses: './.github/actions/install-podman-action'
+        with:
+          platform: 'linux/amd64'
 
       - name: 'Validate pull request number'
         id: 'validate_pr'
--- a/.github/workflows/antigravity-pr-review.yml
+++ b/.github/workflows/antigravity-pr-review.yml
@@ -140,6 +140,7 @@ jobs:
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_TOKEN: ${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}
           PULL_REQUEST_NUMBER: ${{ steps.validate_pr.outputs.pull_request_number }}
+          REVIEW_AUTHOR_LOGIN: ${{ steps.mint_identity_token.outputs.app-slug != '' && format('{0}[bot]', steps.mint_identity_token.outputs.app-slug) || 'github-actions[bot]' }}
           REVIEW_BODY_PATH: 'review-summary-body.md'
           WORKFLOW_RUN_URL: '${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}'
         run: 'uv run --package odh-ci-agent odh-ci-upsert-review-comment'
--- a/.github/workflows/antigravity-pr-review.yml
+++ b/.github/workflows/antigravity-pr-review.yml
@@ -135,6 +135,7 @@ jobs:
 
       - name: 'Upsert review summary comment'
         env:
+          GITHUB_APP_SLUG: ${{ steps.mint_identity_token.outputs.app-slug }}
           GITHUB_REPOSITORY: ${{ github.repository }}
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_TOKEN: ${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}
--- a/ci/agentic-reviewer/src/odh_ci_agent/github_api.py
+++ b/ci/agentic-reviewer/src/odh_ci_agent/github_api.py
@@ -49,6 +49,20 @@ def read_github_token() -> str:
     return os.environ.get("GITHUB_TOKEN", "").strip()
 
 
+def _is_user_endpoint_forbidden(exc: GitHubCommandError) -> bool:
+    return "Resource not accessible by integration" in exc.stdout or "Resource not accessible by integration" in exc.stderr
+
+
+def _login_from_user_api() -> str:
+    user = gh_api_json("user")
+    if not isinstance(user, dict):
+        raise SystemExit("Expected GitHub user response to include login")
+    login = user.get("login")
+    if not isinstance(login, str) or not login:
+        raise SystemExit("Expected GitHub user response to include login")
+    return login
+
+
 @functools.lru_cache(maxsize=1)
 def authenticated_user_login() -> str:
     """Return the login for the token in GITHUB_TOKEN (workflow or GitHub App bot)."""
@@ -61,13 +75,12 @@ def authenticated_user_login() -> str:
     if app_slug:
         return f"{app_slug}[bot]"
 
-    user = gh_api_json("user")
-    if not isinstance(user, dict):
-        raise SystemExit("Expected GitHub user response to include login")
-    login = user.get("login")
-    if not isinstance(login, str) or not login:
-        raise SystemExit("Expected GitHub user response to include login")
-    return login
+    try:
+        return _login_from_user_api()
+    except GitHubCommandError as exc:
+        if _is_user_endpoint_forbidden(exc):
+            return "github-actions[bot]"
+        raise
 
 
 def _query_path(path: str, query: Mapping[str, object] | None = None) -> str:
--- a/ci/agentic-reviewer/src/odh_ci_agent/github_api.py
+++ b/ci/agentic-reviewer/src/odh_ci_agent/github_api.py
@@ -53,6 +53,14 @@ def read_github_token() -> str:
 def authenticated_user_login() -> str:
     """Return the login for the token in GITHUB_TOKEN (workflow or GitHub App bot)."""
 
+    explicit_login = os.environ.get("REVIEW_AUTHOR_LOGIN", "").strip()
+    if explicit_login:
+        return explicit_login
+
+    app_slug = os.environ.get("GITHUB_APP_SLUG", "").strip()
+    if app_slug:
+        return f"{app_slug}[bot]"
+
     user = gh_api_json("user")
     if not isinstance(user, dict):
         raise SystemExit("Expected GitHub user response to include login")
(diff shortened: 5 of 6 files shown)


.github/workflows/antigravity-dispatch.yml
jiridanek, line 1: @coderabbitai The root cause is now clear. When vars.APP_ID is set, the workflow mints a GitHub App token and sets it as GITHUB_TOKEN for the upsert step.

GitHub App tokens cannot call GET /user, which causes the failure. The bot login should instead come from the GITHUB_ACTOR environment variable or a dedicated app bot login variable.

GITHUB_ACTOR reflects the workflow triggerer, not the GitHub App bot. Comments created with an app token are authored by the app's bot user, so the bot login must come from a dedicated app variable rather than the environment variable.

Several alternatives to fetch the app bot login were considered. The /app endpoint likely fails with an app token too. Hardcoding the bot slug is fragile. The most promising path is to skip pre-fetching the login entirely and instead match existing comments by their marker during upsert.

Matching comments by marker alone would recreate the original CodeRabbit collision bug. Filtering by both marker and author login is required, but the author login is precisely what the failing API call was meant to retrieve. After posting or patching a comment, the API response includes user.login, yet filtering existing comments by marker still leaves the author identity unresolved. The typical GitHub App bot login format is {app-slug}[bot].

The app slug could be fetched from the repository installation endpoint. Calling the full installations list with a JWT is too heavy. The bot login is simply the app slug plus the [bot] suffix.

GraphQL queries such as viewer { login } are unlikely to work with an installation token. The repository owner login does not identify the GitHub App bot user. The best fix is to pass the bot login as a dedicated environment variable from the workflow.

The concrete fix is to derive REVIEW_AUTHOR_LOGIN from the create-github-app-token action output. The bot login would be the app slug plus [bot]. The next step is to verify the exact output field name from that action.

The create-github-app-token action probably does not expose the bot login directly. A cleaner approach is to add an optional workflow environment variable, such as ANTIGRAVITY_AUTHOR_LOGIN, and read it in authenticated_user_login(). This avoids falling back to the failing gh api user call entirely.

A simpler alternative is to avoid overriding GITHUB_TOKEN with the GitHub App token for the upsert step. The default secrets.GITHUB_TOKEN would be used instead, which would author comments as github-actions[bot] rather than the app bot. This sidesteps the /user 403 error entirely, though it would require updating the author login matching logic to recognize the default bot identity.

Using the default GITHUB_TOKEN would duplicate existing comments rather than update them, because the app bot would be treated as a different author. The two remaining options are passing the bot login from a workflow variable, or trying a GraphQL viewer { login } query to retrieve the bot identity from the installation token.

Searching the codebase for existing bot-login patterns. A lightweight API call was considered, but it would be a bad request. Reading an existing owned comment is circular. The /app endpoint requires JWT authentication, and /installation/repositories does not return the user login.

The simplest fix is to add a workflow environment variable, such as ANTIGRAVITY_BOT_LOGIN, and pass the app bot login from a repository variable. Another approach is to fetch all existing comments filtered by the marker and use the user.login returned on those comments as the author identity.

A concrete implementation is taking shape. The resolve_author_login helper would scan existing comments for the antigravity marker and return the user.login from the first match. This removes the need for the failing gh api user call entirely. The author identity is then known before any comment is created, so filtering by author login is no longer required.

The fallback to authenticated_user_login() is only needed when no existing marked comments exist. For first-time comments, author filtering can be skipped entirely. Existing comments with the marker provide the author login directly, preventing the original CodeRabbit collision bug.

The implementation will split the existing comment check into a marker-only filter and a full author check. resolve_author_login will scan existing comments for the antigravity marker and return the user.login from the first match. This removes the need for the failing gh api user call entirely.

The fallback to authenticated_user_login() is only needed when no existing marked comments are found. For first-time comments, the logic can skip author filtering entirely and post directly. Author login is only required when looking up or superseding existing comments.

The author login is only needed for the supersede step, which never runs for first-time comments. The revised flow searches for comments by marker alone, derives the author login from any existing matches, and falls back to the newly created comment's user.login for superseding. This completes the refactor.

A concrete infer_author_login helper was drafted to scan existing comments for the antigravity marker and extract the author login from the first match. However, this approach is unsafe: if no existing marked comments are found, filtering by marker alone would match the CodeRabbit bot instead of the intended app bot. The author identity must therefore be resolved before any comment lookup occurs.

The author login cannot be deferred to the creation response. On the first run, no existing bot comments exist yet, so the lookup would match the CodeRabbit bot's comment by marker alone. The author identity must be known before any comment search begins.

The workflow environment variable github.repository_owner is the organization name, not the bot login. The next step is to inspect the create-github-app-token action README to see if it exposes the bot login directly.

Using the default secrets.GITHUB_TOKEN would require messy author-matching logic, so that path is rejected. The preferred solution is to pass the bot login through a workflow variable such as APP_BOT_LOGIN or REVIEW_BOT_LOGIN. The next step is to search the repository for existing bot-login or app-ID variables.

jiridanek: @coderabbitai

```
Subject: [PATCH] RHAIENG-6403: fix(ci/agentic-reviewer): resolve app bot login without GET /user
---
Index: .github/workflows/antigravity-pr-review.yml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/.github/workflows/antigravity-pr-review.yml b/.github/workflows/antigravity-pr-review.yml
--- a/.github/workflows/antigravity-pr-review.yml	(revision c6ce10359984f61338c40d605006731fd0314084)
+++ b/.github/workflows/antigravity-pr-review.yml	(revision c2f69d9c22220da84fc0e6778dc798514230dadc)
@@ -135,6 +135,7 @@
 
       - name: 'Upsert review summary comment'
         env:
+          GITHUB_APP_SLUG: ${{ steps.mint_identity_token.outputs.app-slug }}
           GITHUB_REPOSITORY: ${{ github.repository }}
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_TOKEN: ${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}
Index: ci/agentic-reviewer/src/odh_ci_agent/github_api.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/ci/agentic-reviewer/src/odh_ci_agent/github_api.py b/ci/agentic-reviewer/src/odh_ci_agent/github_api.py
--- a/ci/agentic-reviewer/src/odh_ci_agent/github_api.py	(revision c6ce10359984f61338c40d605006731fd0314084)
+++ b/ci/agentic-reviewer/src/odh_ci_agent/github_api.py	(revision c2f69d9c22220da84fc0e6778dc798514230dadc)
@@ -53,6 +53,14 @@
 def authenticated_user_login() -> str:
     """Return the login for the token in GITHUB_TOKEN (workflow or GitHub App bot)."""
 
+    explicit_login = os.environ.get("REVIEW_AUTHOR_LOGIN", "").strip()
+    if explicit_login:
+        return explicit_login
+
+    app_slug = os.environ.get("GITHUB_APP_SLUG", "").strip()
+    if app_slug:
+        return f"{app_slug}[bot]"
+
     user = gh_api_json("user")
     if not isinstance(user, dict):
         raise SystemExit("Expected GitHub user response to include login")
Index: ci/agentic-reviewer/tests/unit/test_github_api.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/ci/agentic-reviewer/tests/unit/test_github_api.py b/ci/agentic-reviewer/tests/unit/test_github_api.py
--- a/ci/agentic-reviewer/tests/unit/test_github_api.py	(revision c6ce10359984f61338c40d605006731fd0314084)
+++ b/ci/agentic-reviewer/tests/unit/test_github_api.py	(revision c2f69d9c22220da84fc0e6778dc798514230dadc)
@@ -51,8 +51,10 @@
         github_api.parse_positive_issue_number(raw, label="pull request number")
 
 
-def test_authenticated_user_login_returns_login() -> None:
+def test_authenticated_user_login_returns_login(monkeypatch: pytest.MonkeyPatch) -> None:
     github_api.authenticated_user_login.cache_clear()
+    monkeypatch.delenv("GITHUB_APP_SLUG", raising=False)
+    monkeypatch.delenv("REVIEW_AUTHOR_LOGIN", raising=False)
     try:
         with patch(
             "odh_ci_agent.github_api.gh_api_json",
@@ -63,13 +65,42 @@
             mock_gh_api_json.assert_called_once_with("user")
     finally:
         github_api.authenticated_user_login.cache_clear()
+
+
+def test_authenticated_user_login_uses_github_app_slug(monkeypatch: pytest.MonkeyPatch) -> None:
+    github_api.authenticated_user_login.cache_clear()
+    monkeypatch.setenv("GITHUB_APP_SLUG", "odh-antigravity")
+    try:
+        with patch("odh_ci_agent.github_api.gh_api_json") as mock_gh_api_json:
+            assert github_api.authenticated_user_login() == "odh-antigravity[bot]"
+            mock_gh_api_json.assert_not_called()
+    finally:
+        github_api.authenticated_user_login.cache_clear()
+        monkeypatch.delenv("GITHUB_APP_SLUG", raising=False)
+
+
+def test_authenticated_user_login_prefers_explicit_override(monkeypatch: pytest.MonkeyPatch) -> None:
+    github_api.authenticated_user_login.cache_clear()
+    monkeypatch.setenv("REVIEW_AUTHOR_LOGIN", "custom-bot[bot]")
+    monkeypatch.setenv("GITHUB_APP_SLUG", "ignored-app")
+    try:
+        with patch("odh_ci_agent.github_api.gh_api_json") as mock_gh_api_json:
+            assert github_api.authenticated_user_login() == "custom-bot[bot]"
+            mock_gh_api_json.assert_not_called()
+    finally:
+        github_api.authenticated_user_login.cache_clear()
+        monkeypatch.delenv("REVIEW_AUTHOR_LOGIN", raising=False)
+        monkeypatch.delenv("GITHUB_APP_SLUG", raising=False)
 
 
 @pytest.mark.parametrize("user_response", [{"login": 12345}, {"login": ""}, {"login": None}, {}])
 def test_authenticated_user_login_rejects_malformed_login(
+    monkeypatch: pytest.MonkeyPatch,
     user_response: object,
 ) -> None:
     github_api.authenticated_user_login.cache_clear()
+    monkeypatch.delenv("GITHUB_APP_SLUG", raising=False)
+    monkeypatch.delenv("REVIEW_AUTHOR_LOGIN", raising=False)
     with patch("odh_ci_agent.github_api.gh_api_json", return_value=user_response):
         with pytest.raises(SystemExit, match="Expected GitHub user response to include login"):
             github_api.authenticated_user_login()
```

jiridanek: @coderabbitai

```
Subject: [PATCH] RHAIENG-6403: ci: pass app slug to ci-gated review upsert step
---
Index: .github/workflows/antigravity-ci-gated-review.yml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/.github/workflows/antigravity-ci-gated-review.yml b/.github/workflows/antigravity-ci-gated-review.yml
--- a/.github/workflows/antigravity-ci-gated-review.yml	(revision c2f69d9c22220da84fc0e6778dc798514230dadc)
+++ b/.github/workflows/antigravity-ci-gated-review.yml	(revision 00e4b4adad01020519ab42ea8d9687d158b8a886)
@@ -100,6 +100,7 @@
         if: |-
           github.event.workflow_run.pull_requests[0] != null
         env:
+          GITHUB_APP_SLUG: ${{ steps.mint_identity_token.outputs.app-slug }}
           GITHUB_REPOSITORY: ${{ github.repository }}
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_TOKEN: ${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}
```

jiridanek: @coderabbitai recheck if fixed

ide-developer (approved):
Clean fix, well-motivated. Tests pass locally (21/21).

**Core fix (`antigravity-dispatch.yml`)**

- Adding `packages: read` to the `review` job is the correct minimal fix. Reusable workflows can't escalate permissions beyond the caller's grant, so this is the only place to fix it.

**`authenticated_user_login()` refactoring**

- The three-tier fallback (`REVIEW_AUTHOR_LOGIN` → `GITHUB_APP_SLUG` → API `/user` with forbidden-safe catch) is well-structured and removes the hard dependency on `GET /user` that fails with fine-grained app tokens.
- The `_is_user_endpoint_forbidden` helper checking both stdout and stderr is correct — `gh` can surface the error in either stream depending on version.
- `lru_cache` is fine here since env vars are set once per workflow run and don't change mid-execution.

**Observation (not a blocker):** In the workflow YAML, `REVIEW_AUTHOR_LOGIN` is always non-empty (it defaults to `github-actions[bot]` when no app-slug). So the `GITHUB_APP_SLUG` env var and the API fallback path in Python will never be reached from these workflows. That's fine — it makes the Python code resilient to being called from other contexts (local dev, future workflows that don't set the override). Just worth knowing that the test coverage for those paths matters more than their production exercise in *these* specific workflows.

**Tests**

- Good coverage: override precedence, app-slug derivation, forbidden fallback, non-forbidden re-raise, malformed login rejection.
- Tests are deterministic — `monkeypatch.delenv` on both env vars ensures no leakage from the test environment.
- Nit: `monkeypatch.delenv(...)` calls in `finally` blocks are redundant since pytest's `monkeypatch` fixture auto-restores on teardown — but `cache_clear()` in `finally` is essential, so the pattern is fine as-is.

LGTM.

PR #4242, opened 2026-07-31
RHAIENG-6664: fix excessively verbose pytest/CI output

jiridanek:
## Description

Fixes [RHAIENG-6664](https://redhat.atlassian.net/browse/RHAIENG-6664): `gmake test` / the `pytest-tests` CI job produced ~40,800 lines of output for a ~38s run.

Root cause: `pytest.ini` set `log_cli = True`, which per pytest's own logging code silently forces global verbosity to at least 1 regardless of any `-v` flag — this turns the compact dot-per-test summary into one report line per test *and per subtest*. With thousands of subtests in `tests/test_main.py::test_image_pyprojects` (checking every image's `pyproject.toml`/`pylock.toml` consistency via `pytest-subtests`), that's ~38,700 lines of `SUBPASSED`/`SUBSKIPPED`/`SUBXFAIL` output alone. A stray `logging.basicConfig(level=logging.DEBUG)` in `tests/conftest.py` added further `INFO:root:`-style noise that bypassed pytest's own log capture entirely.

Along the way, also found that the `pytest-subtests` dependency itself was dead weight: pytest 9's builtin subtests plugin registers under the same `"subtests"` name and shadows the third-party package, so it was never actually loaded (confirmed via `pytest -VV`) despite `TYPE_CHECKING`-only type hints referencing it throughout the codebase.

Changes:
- Drop the unused `pytest-subtests` dependency; switch `subtests: pytest_subtests(.plugin)?.SubTests` type hints to the real runtime type, `pytest.Subtests` (built into pytest 9+).
- `pytest.ini`: `log_cli = False` by default, explicit `verbosity_subtests = 0` (suppresses SUBPASSED/SUBSKIPPED/SUBXFAIL; SUBFAILED still always shows in full). Local-dev defaults no longer generate a DEBUG log file or `junit.xml` — those are CI-only concerns now, added via `PYTEST_ADDOPTS`/CLI args only in the GitHub Actions jobs that consume them (`code-quality.yaml`, `test-containers.yaml`, `build-notebooks-TEMPLATE.yaml`).
- Remove the stray `logging.basicConfig(level=logging.DEBUG)` from `tests/conftest.py` that was bypassing pytest's log capture.
- `code-quality.yaml`'s `pytest-tests` job now uploads the DEBUG-level pytest log (`logs/pytest-logs.txt`) as an uncompressed workflow artifact, for anyone who needs to debug a CI run without reproducing locally.
[...]

PR #4243, opened 2026-07-31
RHAIENG-6664 followups: quiet tests/containers logging + colorize CI pytest output

jiridanek:
## Description

Two small followups to [RHAIENG-6664](https://redhat.atlassian.net/browse/RHAIENG-6664) / #4242, both discovered while watching CI for that PR.

### 1. `tests/containers` unconditionally installed a DEBUG log handler

`tests/containers` runs printed unconditional `DEBUG:urllib3.connectionpool:...` / `DEBUG:root:...` noise for every test, e.g. https://github.com/opendatahub-io/notebooks/actions/runs/30639752486/job/91186453257?pr=4242 — same root cause as the `pytest-tests` job fixed in #4242, just in a different spot.

`tests/containers/kubernetes_utils.py`, `base_image_test.py`, `workbenches/gpu_library_loading_test.py`, and `podman_machine_utils.py` each called `logging.basicConfig(level=logging.DEBUG)` at import time. This installs a root-logger `StreamHandler` unconditionally the moment any of these modules is imported (which happens for every `tests/containers` run, since `conftest.py` imports `kubernetes_utils` at collection time) — bypassing pytest's own log capture entirely, so it prints DEBUG noise from every library (`urllib3`, `testcontainers`, etc.) for every test, whether it passes or fails.

That DEBUG output can be genuinely useful when a container/pod test fails (pod readiness, port-forward setup, HTTP calls to the container runtime), so the fix isn't to just delete it — it's to let pytest's own logging capture handle it instead: drop `basicConfig()`, and pass `--log-level=DEBUG` at the three `tests/containers` invocations (`test-containers.yaml`, and both pytest calls in `build-notebooks-TEMPLATE.yaml`). This sets pytest's log-capture threshold without forcing verbose mode (unlike `log_cli`, which was the cause of #4242) — DEBUG records are only shown in the "Captured log call" section for tests that actually fail.

### 2. pytest CI output had no color

pytest's `--color` option defaults to `"auto"`, which decides whether to emit ANSI color based on `sys.stdout.isatty()`. A GitHub Actions step's stdout is not a TTY, so `auto` silently disables color, even though the Actions log viewer does render ANSI (the `prek` step in `code-quality.yaml` already relies on this, passing `--color always` explicitly). Fixed by setting `FORCE_COLOR: "1"` — a convention pytest already honors directly, no `--color` flag needed — on the three CI steps that invoke pytest.

Also added a note to `AGENTS.md`'s Operational notes: since CI pytest output can now contain ANSI codes, prefer the `pytest-logs.txt` debug artifact (written via pytest's `log_file` handler, independent of the colorized terminal writer, so always ANSI-free) for any programmatic log processing, with a fallback `sed` pattern for stripping ANSI codes if you must process console text directly.

## How Has This Been Tested?

- Verified the log-level mechanism with a throwaway two-test module (one passing, one failing, each logging at DEBUG): without `--log-level`, neither test's DEBUG log shows; with `--log-level=DEBUG`, the passing test still shows nothing, and the failing test's DEBUG log appears under "Captured log call".
- Verified the color mechanism the same way: without `FORCE_COLOR`, piped/non-TTY pytest output has zero ANSI escape sequences; with `FORCE_COLOR=1`, it has several — and confirmed `log_file` output has zero ANSI sequences either way (0 vs 9-11 escape sequences observed).
- Ran `uv run prek run` (ruff/pyright) and `yamllint --strict` on all touched files — clean.

## Merge criteria:

[RHAIENG-6664]: https://redhat.atlassian.net/browse/RHAIENG-6664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Improved container test diagnostics with DEBUG-level logging and preserved colored output in CI logs.
  * Added matrix-specific downloadable pytest log artifacts, retained for 14 days, to simplify troubleshooting.
  * Applied consistent colored output settings across automated test and quality-check workflows.

* **Chores**
  * Removed test-level global logging configuration to avoid unintended logging side effects.

* **Documentation**
  * Added guidance for reviewing and cleaning ANSI-colored pytest logs.

### DIFF
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -492,7 +492,7 @@ jobs:
         run: |
           set -Eeuxo pipefail
           mkdir -p "$(dirname "${JUNIT_XML}")"
-          uv run pytest --capture=fd tests/containers -m 'not openshift and not cuda and not rocm and not manifest_validation' --image="${OUTPUT_IMAGE}" --junitxml="${JUNIT_XML}" -o junit_family=legacy
+          uv run pytest --capture=fd tests/containers -m 'not openshift and not cuda and not rocm and not manifest_validation' --image="${OUTPUT_IMAGE}" --junitxml="${JUNIT_XML}" -o junit_family=legacy --log-level=DEBUG
 
       - name: Upload Testcontainers pytest JUnit XML
         if: ${{ always() && steps.pytest-testcontainers.conclusion != 'skipped' }}
@@ -566,7 +566,7 @@ jobs:
         run: |
           set -Eeuxo pipefail
           mkdir -p "$(dirname "${JUNIT_XML}")"
-          uv run pytest --capture=fd tests/containers -m 'openshift and not cuda and not rocm' --image="${OUTPUT_IMAGE}" --junitxml="${JUNIT_XML}" -o junit_family=legacy
+          uv run pytest --capture=fd tests/containers -m 'openshift and not cuda and not rocm' --image="${OUTPUT_IMAGE}" --junitxml="${JUNIT_XML}" -o junit_family=legacy --log-level=DEBUG
 
       - name: Upload OpenShift pytest JUnit XML
         if: ${{ always() && steps.pytest-openshift.conclusion != 'skipped' }}
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -489,6 +489,9 @@ jobs:
           TESTCONTAINERS_RYUK_DISABLED: "true"
           OUTPUT_IMAGE: "${{ steps.calculated_vars.outputs.OUTPUT_IMAGE }}"
           JUNIT_XML: "${{ runner.temp }}/pytest-reports/${{ inputs.target }}_${{ steps.calculated_vars.outputs.BUILD_TYPE }}_${{ steps.calculated_vars.outputs.SANITIZED_PLATFORM }}_testcontainers-junit.xml"
+          # pytest's --color=auto sees a non-tty stdout in CI and disables color;
+          # FORCE_COLOR forces it back on for the (colorable) GitHub Actions log viewer.
+          FORCE_COLOR: "1"
         run: |
           set -Eeuxo pipefail
           mkdir -p "$(dirname "${JUNIT_XML}")"
@@ -563,6 +566,9 @@ jobs:
           TESTCONTAINERS_RYUK_DISABLED: "true"
           OUTPUT_IMAGE: "${{ steps.calculated_vars.outputs.OUTPUT_IMAGE }}"
           JUNIT_XML: "${{ runner.temp }}/pytest-reports/${{ inputs.target }}_${{ steps.calculated_vars.outputs.BUILD_TYPE }}_${{ steps.calculated_vars.outputs.SANITIZED_PLATFORM }}_openshift-junit.xml"
+          # pytest's --color=auto sees a non-tty stdout in CI and disables color;
+          # FORCE_COLOR forces it back on for the (colorable) GitHub Actions log viewer.
+          FORCE_COLOR: "1"
         run: |
           set -Eeuxo pipefail
           mkdir -p "$(dirname "${JUNIT_XML}")"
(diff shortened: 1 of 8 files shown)


jiridanek:
@coderabbitai CI verification summary for this followup:

- **`Code static analysis`** ([run](https://github.com/opendatahub-io/notebooks/actions/runs/30641455436)): ✅ success (`pytest-tests` job unaffected by this change, as expected — it doesn't touch `tests/containers`).
- **`Test Build Notebooks Template`** ([run](https://github.com/opendatahub-io/notebooks/actions/runs/30641455958)): ✅ success, including both `tests/containers` pytest invocations (testcontainers + OpenShift) now running with `--log-level=DEBUG`.
- **`Test Infrastructure Self-Test`** ([run](https://github.com/opendatahub-io/notebooks/actions/runs/30641457510)): 3/4 images passed; `codeserver-ubi9-python-3.12` failed with the exact same pre-existing `AH00016: Configuration Failed` httpd flake as on the base PR (#4242) and on `main` — `8 failed, 11 passed, 20 skipped` in both runs, confirming this is unrelated to the `basicConfig`→`--log-level=DEBUG` change.

Also verified the underlying mechanism locally with a throwaway two-test module (one passing, one failing, both logging at DEBUG): without `--log-level`, neither shows; with `--log-level=DEBUG`, the passing test stays silent and the failing test's DEBUG log appears under "Captured log call" — i.e. useful debug info surfaces only when a test actually fails, instead of the old `basicConfig(DEBUG)` printing it unconditionally for every test.

ide-developer (approved):
## Review

**Summary**: Follow-up to #4242 — removes module-level `logging.basicConfig(level=logging.DEBUG)` calls from four `tests/containers` modules (which installed a root-logger handler unconditionally at import time, bypassing pytest's log capture) and instead passes `--log-level=DEBUG` to the three `tests/containers` pytest invocations in CI, so DEBUG logs only surface in "Captured log call" for failing tests.

### Correctness
- The mechanism is sound and I verified it against this PR's own CI run: in the `codeserver-ubi9-python-3.12` job (pre-existing, unrelated 502/mac-address flake — not caused by this PR), the failing tests now show `DEBUG testcontainers.core.waiting_utils:waiting_utils.py:197 ...` and `DEBUG root:workbench_image_test.py:361 ...` lines in the captured output, confirming `--log-level=DEBUG` is doing its job without `log_cli` forcing verbosity for passing tests.
- Checked that `import logging` is still used in all four touched files after removing `basicConfig()` (`logging.debug/info/error` calls remain), so no dead imports were left behind.
- `pytest.ini`'s existing `log_cli = False` / `log_cli_level = INFO` settings are untouched and don't conflict with the new `--log-level=DEBUG` (that flag controls the capture threshold, not `log_cli` output), consistent with the PR description's explanation of the #4242 root cause.

### Minor nit
- `.github/workflows/software-versions.yaml` also runs a `tests/containers` test (`manifest_validation_test.py::test_old_tag_annotations_match_quay`) but wasn't given `--log-level=DEBUG`. Since `conftest.py` imports `kubernetes_utils` at collection time for any `tests/containers` run, this job also benefited from the `basicConfig()` removal (no more forced noise) — but on failure it won't get DEBUG-level "Captured log call" output like the other three invocations now do. Probably not worth blocking on since this test doesn't hit the container-readiness code paths that emit those DEBUG logs, but flagging for consistency/completeness.

### Style / process
- Small, focused, single-purpose diff — good scope discipline for a follow-up fix.
- PR description clearly documents the verification methodology (throwaway pass/fail test module) — appreciated, this is exactly the kind of behavior that's easy to get backwards.

Overall: LGTM, nice catch continuing the cleanup from #4242.

jiridanek:
@coderabbitai good catch — went with the "fix it" option rather than just scoping the docs: added \`log_file\`/\`log_file_level\` and an "Upload pytest debug log" step to \`test-containers.yaml\`'s \`container-tests\` job, mirroring the \`pytest-tests\` job in \`code-quality.yaml\`. Generalized the AGENTS.md wording to name both jobs instead of just one.

Also found and fixed a follow-on bug while verifying on a real run: since \`container-tests\` is a 4-way matrix, and \`actions/upload-artifact\`'s \`archive: false\` mode registers the artifact under the *source file's basename* rather than the \`name:\` input, all 4 matrix jobs' artifacts showed up as identically-named \`pytest-logs.txt\` in the run (distinguishable only by numeric artifact ID). Fixed by giving the on-disk log file itself a matrix-scoped name (\`logs/pytest-logs-${LOG_PREFIX}.txt\`) so the uploaded artifact name is unambiguous regardless of that action quirk.

ide-developer (approved):
LGTM

PR #4259, opened 2026-08-02
ci(test-containers): add a job to run the openshift-marked container tests

### LINKED ISSUE #4257: test-containers.yaml self-test doesn't exercise the openshift-marked tests (ImageDeployment/test_image_run_on_openshift)
## Summary

`.github/workflows/test-containers.yaml` ("Test Infrastructure Self-Test") only runs:

```
uv run pytest tests/containers -m 'not openshift and not cuda and not rocm and not manifest_validation' ...
```

explicitly excluding everything marked `openshift` — i.e. `tests/containers/workbenches/workbench_image_test.py::TestWorkbenchImage::test_image_run_on_openshift`, which exercises `kubernetes_utils.ImageDeployment` (creating a bare `Deployment`, port-forwarding through the k8s API, HTTP-polling for readiness).

That code path only otherwise runs as part of the full image build workflow (`build-notebooks-TEMPLATE.yaml`), gated behind an actual multi-arch image build. As a result it's expensive to exercise and easy to silently regress without anyone noticing for a while.

## What happened downstream

On `rhoai-2.25` (red-hat-data-services/notebooks), this exact code path regressed twice in a row without being caught:

1. A previously-correct RStudio-aware readiness probe (probing `/` with a persistent session, since RStudio's nginx doesn't serve `/api` and its `/api` redirect is hardcoded to the wrong port under any port-forward) got silently reverted by an unrelated large backport merge.
2. A follow-up fix for the resulting timeout only noticed the *symptom* (timeout too short) and copied `main`'s `/api`-only probe — appropriate for `main` (which no longer ships RStudio), but wrong for `rhoai-2.25` (which still does), so it didn't actually fix anything for RStudio.

Both regressions went unnoticed for weeks because nothing in fast, PR-scoped CI exercised `test_image_run_on_openshift` — only the full, expensive image-build workflow did, and its failures there are easy to write off as "flaky/expected" since they show up on nearly every push.

See red-hat-data-services/notebooks#2684 and the fix in red-hat-data-services/notebooks#2685 for the full history and root cause.

## Proposal

Add a job to `test-containers.yaml` (alongside the existing `container-tests` job) that runs the `openshift`-marked tests specifically, reusing the `find-images` job's matrix and the same provisioning building blocks the main build workflow already uses for this:

- `./.github/actions/install-podman-action` (rootful podman, sharing image storage with cri-o) instead of the existing job's rootless podman
- `./.github/actions/provision-k8s` (kubeadm-based single-node cluster)
- `uv run pytest tests/containers -m 'openshift and not cuda and not rocm' --image=<pulled image>`

This gets openshift-test coverage into fast, PR-scoped CI instead of only the full build workflow, so regressions in this path get caught immediately instead of being buried in build-workflow noise.

A concrete implementation exists on `rhoai-2.25` at red-hat-data-services/notebooks#2685 (second commit) that can be used as a reference/starting point for porting to `main`.

jiridanek:
## Summary

`.github/workflows/test-containers.yaml`'s self-test job explicitly excludes everything marked `openshift` — i.e. `test_image_run_on_openshift`, which exercises `kubernetes_utils.ImageDeployment` (Deployment + port-forward + HTTP readiness poll). That code path only otherwise runs inside the full, expensive image-build workflow, so regressions there go unnoticed for weeks (see red-hat-data-services/notebooks#2684, red-hat-data-services/notebooks#2685 for a concrete two-in-a-row regression that this exact gap let through downstream).

Two commits, cherry-picked (`-x`) from the concrete `rhoai-2.25` implementation at red-hat-data-services/notebooks#2685, with action pins/setup steps adapted to `main`'s current conventions (`./.github/actions/setup-uv`, `actions/checkout@…v7.0.0`, `docker/login-action@…v4.4.0`):

1. **`ci(test-containers): add a job to run the openshift-marked container tests`** — adds `openshift-container-tests`, reusing `find-images`'s matrix and mirroring the provisioning steps `build-notebooks-TEMPLATE.yaml` already uses for the same purpose: rootful podman (`install-podman-action`, sharing image storage with cri-o) + a kubeadm cluster (`provision-k8s`), then `pytest -m 'openshift and not cuda and not rocm'` against the pulled image.
2. **`fix(ci): disable zstd:chunked partial pulls in openshift-container-tests`** — fixes a failure the new job hit immediately in the reference PR's own CI: `podman pull` of an image with `zstd:chunked` layers failed with `501 Unsupported client range` against `ghcr.io`'s blob endpoint, because `install-podman-action`'s `storage.conf` enables partial/ranged fetches. Disabled for this job only.

Not ported: the reference PR's third commit (RStudio `/` vs `/api` readiness-probe fix) — `main` dropped RStudio entirely, so the existing `/api`-only probe in `kubernetes_utils.py` is already correct here; that fix is `rhoai-2.25`-specific.

Fixes #4257.

## Test plan

- [x] `yamllint --strict` and `pinact run --check` clean on the edited workflow file.
- [ ] CI: `openshift-container-tests` job runs and passes across the matrix.

## Summary by CodeRabbit

* **Tests**
  * Added automated container testing on Kubernetes and OpenShift environments.
  * Added test result reporting through uploaded JUnit reports.

### DIFF
--- a/.github/workflows/test-containers.yaml
+++ b/.github/workflows/test-containers.yaml
@@ -176,3 +176,57 @@ jobs:
         uses: ./.github/actions/capture-kernel-logs
         with:
           log-prefix: ${{ matrix.name }}
+
+  openshift-container-tests:
+    name: "openshift: ${{ matrix.name }}"
+    needs: find-images
+    runs-on: ubuntu-26.04
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.find-images.outputs.matrix) }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+
+      - name: Setup uv and Python
+        uses: ./.github/actions/setup-uv
+
+      - name: Install deps
+        run: uv sync --locked
+
+      # Rootful podman sharing image storage with cri-o, same as build-notebooks-TEMPLATE.yaml,
+      # so the pod created by the openshift-marked tests can find the image without re-pulling.
+      - name: Install and configure Podman
+        uses: './.github/actions/install-podman-action'
+        with:
+          platform: linux/amd64
+
+      - name: Provision K8s cluster
+        uses: ./.github/actions/provision-k8s
+
+      - name: Login to GHCR
+        uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0  # v4.4.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Pull and run OpenShift container tests
+        env:
+          TEST_IMAGE: ${{ matrix.image }}
+          DOCKER_HOST: "unix:///var/run/podman/podman.sock"
+          TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE: "/var/run/podman/podman.sock"
+        run: |
+          set -Eeuxo pipefail
+          podman pull "${TEST_IMAGE}"
+          uv run pytest --capture=fd tests/containers \
+            -m 'openshift and not cuda and not rocm' \
+            --image="${TEST_IMAGE}" \
+            --junitxml=junit-openshift.xml \
+            -v
+
+      - name: Upload test results
+        if: ${{ !cancelled() }}
+        uses: codecov/test-results-action@0fa95f0e1eeaafde2c782583b36b28ad0d8c77d3  # v1.2.1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: junit-openshift.xml
--- a/.github/workflows/test-containers.yaml
+++ b/.github/workflows/test-containers.yaml
@@ -186,6 +186,8 @@ jobs:
       matrix: ${{ fromJSON(needs.find-images.outputs.matrix) }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          persist-credentials: false
 
       - name: Setup uv and Python
         uses: ./.github/actions/setup-uv
--- a/.github/workflows/test-containers.yaml
+++ b/.github/workflows/test-containers.yaml
@@ -210,6 +210,14 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      # install-podman-action's storage.conf enables zstd:chunked partial pulls, which
+      # only matters for `podman build`'s own registry pulls elsewhere. Here we `podman
+      # pull` an already-published ghcr.io image, and ghcr.io's blob endpoint returns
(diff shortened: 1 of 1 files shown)


jiridanek:
## Review: #4259 — add `openshift-container-tests` job

### Overview
Adds a new `openshift-container-tests` job to `test-containers.yaml` that runs the previously-excluded `openshift`-marked pytest tests (`test_image_run_on_openshift`) against each matrix image, using the same rootful-podman + kubeadm provisioning (`install-podman-action` + `provision-k8s`) that `build-notebooks-TEMPLATE.yaml` already relies on for this. Directly closes the coverage gap described in #4257: this code path previously only ran inside the full, expensive image-build workflow.

Ported via `git cherry-pick -x` from a concrete, already-CI-validated implementation on `rhoai-2.25` (red-hat-data-services/notebooks#2685), with two commits carried over (job addition + a zstd:chunked partial-pull fix the job hit immediately in that PR's own CI) and one commit correctly left out (an RStudio-specific readiness-probe fix — `main` has no RStudio).

### Correctness
- Job structure, provisioning steps, and env vars (`DOCKER_HOST`, `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE`) mirror the pattern already proven in `build-notebooks-TEMPLATE.yaml`'s own OpenShift test step — low risk of a new, untested failure mode.
- `-m 'openshift and not cuda and not rocm'` correctly captures `test_image_run_on_openshift` (openshift-only marker) while excluding `test_cuda_run_on_openshift` (openshift+cuda), matching the marker layout in `tests/containers/workbenches/*.py`.
- Depends on `find-images`, which already gates on `github.event.pull_request.head.repo.full_name == base.repo.full_name` — fork PRs correctly get this job skipped too (skipped `needs`), same as the existing `container-tests` job. No new fork-PR credential exposure.
- `persist-credentials: false` was added to the new job's checkout (CodeRabbit catch, already fixed) — good.

### Minor / nit
- `--capture=fd` on the pytest invocation silently overrides the repo-wide `--capture=no` set in `pytest.ini`'s `addopts`. That's presumably intentional (interleaved output from a k8s port-forward thread or similar), and it's inherited as-is from the already-proven `rhoai-2.25` source commit, but unlike this same file's other deviations (AppArmor workaround, zstd:chunked fix, `FORCE_COLOR`) it has no comment explaining why. A one-line comment would save the next person from re-deriving it.
- No `pytest-logs*.txt` debug-artifact upload or `capture-kernel-logs` failure step, unlike the sibling `container-tests` job — reasonable to leave as a faithful, minimal port rather than scope-creeping this PR, but worth a follow-up if this job turns out flaky in practice.

### Security
- GHCR login and podman socket setup match existing, already-reviewed patterns in this file. No new secrets or scopes introduced.

### CI cost
- Adds a 4-way matrix job with full kubeadm cluster provisioning per PR touching `tests/containers/**`. This is the explicit intended tradeoff from #4257 (catch regressions in PR-scoped CI instead of only the expensive build workflow), so acceptable.

### Verdict
Low-risk, well-justified CI addition backed by a real prior deployment of the same config. No blocking issues.

ide-developer (approved):
LGTM

PR #4392, opened 2026-08-16
ISSUE #4374: fix(renovate): separate minor and patch base image updates

### LINKED ISSUE #4374: custom.regex manager needs separateMinorPatch — blocked minor bump masks available patch bump for AIPCC base images
## Summary

The `custom.regex` manager for AIPCC `BASE_IMAGE=` refs in `build-args/konflux.*.conf` (`.github/renovate.json5`) puts patch and minor updates into the same "non-major" bucket, so when a newer *minor* version line exists upstream, Renovate proposes only that (blocked-by-policy) minor bump and silently drops the smaller, allowed patch bump — even though the patch bump is available and should be proposed.

Concretely, on `rhoai-3.4` today:

| Image | Pinned | Available patch | Available minor (masks patch) |
|---|---|---|---|
| `quay.io/aipcc/base-images/cpu` | `3.4.1-1780598682` | `3.4.2-1786384776` | `3.5.0-1786384816` |
| `quay.io/aipcc/base-images/cuda-12.9-el9.6` | `3.4.1-1780597456` | `3.4.2-1786384828` | `3.5.0-1786385265` |
| `quay.io/aipcc/base-images/cuda-13.0-el9.6` | `3.4.1-1780602628` | `3.4.2-1786385291` | `3.5.0-1786385055` |
| `quay.io/aipcc/base-images/rocm-6.4-el9.6` | `3.4.1-1780598009` | `3.4.2-1786384478` | *(none published yet)* |

A local dry-run (`./uv run python scripts/ci/renovate_run.py lookup`, current tree checked out at `rhoai-3.4`) shows Renovate resolving `cpu`/`cuda-12.9-el9.6`/`cuda-13.0-el9.6` to the `3.5.0-*` candidate (`updateType: "minor"`), which then gets correctly blocked by:

```json5
{
  "description": "Block major/minor base image version upgrades (default)",
  "matchManagers": ["custom.regex"],
  "matchUpdateTypes": ["major", "minor"],
  "enabled": false,
},
```

— but Renovate never falls back to proposing the smaller `3.4.2-*` patch bump for those three images. Only `rocm-6.4-el9.6` (no `3.5.0` line published yet) correctly resolves to a `patch` update and would be proposed.

This means release branches (`rhoai-2.25`, `rhoai-3.3`, `rhoai-3.4`) can silently miss patch-level AIPCC base-image bumps — including CVE fixes — whenever a newer minor line has also been published, which is exactly the situation that surfaced an internal report of ffmpeg/libaom CVE fixes not reaching `rhoai-3.4` notebook workbenches.

## Steps to reproduce

```bash
export RENOVATE_TOKEN=$(gh auth token)
unset RENOVATE_REPOSITORIES RENOVATE_BASE_BRANCHES   # local mode uses current checked-out branch
export LOG_LEVEL=debug
git checkout rhoai-3.4   # or the release branch you want to check
./uv run python scripts/ci/renovate_run.py lookup 2>&1 | grep -o '{"depName":"quay.io/aipcc[^}]*}'
```

Output shows `"updateType":"minor"` for cpu/cuda-12.9/cuda-13.0 pointing at `3.5.0-*`, with no separate `3.4.2-*` patch candidate offered.

## What did you expect to happen

Renovate should propose the `3.4.2-*` patch update independently, even when a `3.5.0-*` minor update also exists and is blocked by policy.

## Suggested fix

Add `"separateMinorPatch": true` to the `custom.regex` manager (or a `packageRules` entry scoped to `matchManagers: ["custom.regex"]`), so patch and minor/major updates are computed and proposed as separate candidates instead of one merged "non-major" bucket. See https://docs.renovatebot.com/configuration-options/#separateminorpatch

## Additional context

- `.github/renovate.json5` — `customManagers[0]`, `packageRules` (block major/minor rule).
- Related prior investigations: #3418, #3543, #2714.
- Related: #4373 (remote-mode manager extraction gap) — that bug currently prevents any of this from being exercised through the `renovate_run.py remote` path at all; both need fixing for this update to flow through end to end.

Ehtasham-Yasin:
Fixes #4374.

Configure Renovate to keep `custom.regex` minor and patch updates in separate candidates. This allows release branches to receive permitted patch-level AIPCC/RHAI base-image updates even when a newer, policy-blocked minor version exists.

### Validation

- Renovate semantic configuration validation passed
- Focused configuration tests: 15 passed
- Repository hooks, Ruff, Pyright, and secret detection passed
- Renovate lookup dry-run against `rhoai-3.4` completed successfully
- Full static pytest run: 433 passed; 10 Make-dependent tests could not run because GNU Make 4/`gmake` is unavailable locally

## Summary by CodeRabbit

* **Chores**
  * Improved automated dependency update handling for base-image upgrades by separating minor and patch updates.

* **Tests**
  * Added validation to ensure the updated dependency-management rule remains enabled and correctly configured.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -206,6 +206,11 @@
       "versioning": "semver",
     },
     {
+      "description": "Separate minor and patch base image upgrades",
+      "matchManagers": ["custom.regex"],
+      "separateMinorPatch": true,
+    },
+    {
       // One PR per quay.io/aipcc image (cpu, cuda-*, …); registry.redhat.io/rhai
       // uses the packageRule below.
       "description": "Group konflux build-args by image (custom.regex / docker)",
--- a/tests/test_renovate_config.py
+++ b/tests/test_renovate_config.py
@@ -7,3 +7,7 @@ def test_repo_renovate_json5_passes_semantic_validation() -> None:
     config = validate_renovate_config.load_config(validate_renovate_config.DEFAULT_CONFIG)
     errors = validate_renovate_config.validate_config(config)
     assert errors == [], f"Expected no semantic validation errors, got: {errors}"
+    assert any(
+        rule.get("matchManagers") == ["custom.regex"] and rule.get("separateMinorPatch") is True
+        for rule in config["packageRules"]
+    )


ide-developer (commented):
## Review summary

The core fix is sound. I reproduced the exact scenario from #4374 against the real registry (`quay.io/aipcc/base-images/cpu` pinned at `3.4.1-1780598682`, real patch `3.4.2` and minor `3.5.0` both available) by running the actual `renovate:43` image locally with this PR's rule shape. `separateMinorPatch: true` correctly produces two independent update candidates, and — worth calling out since it wasn't obvious from the docs alone — the existing `groupName: "konflux BASE_IMAGE {{{depNameSanitized}}}"` grouping rule does **not** collide them onto one branch: Renovate auto-prefixes the patch bucket's `groupSlug` (`renovate/patch-konflux-base_image-...` vs `renovate/konflux-base_image-...`), so the fix is safe on branches where major/minor is unblocked (`main`) as well as where it's blocked (release branches). Good fix, correctly scoped to the manager the issue describes.

The findings below are about the verification/test surface around this change, not the runtime behavior itself — non-blocking, but worth a follow-up.

1. **CI never actually exercises this manager.** `.github/workflows/validate-renovate-config.yaml`'s `validate-renovate-dry-run` job — the only check that runs live Renovate — is skipped for fork PRs (confirmed via this PR's own checks: it shows `skipping`, since the head repo is a fork). Even on same-repo PRs, `scripts/ci/validate_renovate_dry_run.py:187` hardcodes `RENOVATE_ENABLED_MANAGERS=github-actions` for every scenario, so `custom.regex` is never exercised end-to-end by CI at all, for anyone. A future regression to `separateMinorPatch`'s effect would ship with zero live signal.

2. **The new invariant bypasses this file's own validation pattern**, creating a split-brain check. Every other semantic invariant here (MintMaker policy, PR-title prefix, CentOS pin, ODH BASE_IMAGE policy) is enforced inside `scripts/ci/validate_renovate_config.py`'s `validate_config()`, contributing a descriptive string to `errors`. This PR instead adds a bare `assert any(...)` directly in `tests/test_renovate_config.py` (lines 10-13), so `validate_config()` was never taught the new rule (confirmed: `separateMinorPatch` doesn't appear anywhere in `validate_renovate_config.py`). That matters concretely because `.github/workflows/validate-renovate-config.yaml`'s "Validate renovate.json5 semantics" step runs `python scripts/ci/validate_renovate_config.py` **directly**, not via pytest — that CLI entrypoint (and anyone using it standalone, e.g. against the RHDS fork's config) will print `OK` even if this rule is later removed or broken.

3. **The new assertion's predicate isn't actually selective, and gives no diagnostic on failure.** `rule.get("matchManagers") == ["custom.regex"]` matches 8 of the file's 20 `packageRules` (verified by parsing the live config) — only the `and rule.get("separateMinorPatch") is True` clause does the real discriminating work. If a future edit ever moved that flag onto the wrong one of those 8 rules, this check would still pass. It's also a bare `assert` with no message, unlike the `assert errors == [], f"..."` immediately above it — a failure here just shows `assert False` with no context. Matching by `description` (the established idiom elsewhere in this file, e.g. `PREFIX_RULE_DESCRIPTION`, `CENTOS_STREAM_RULE_DESCRIPTION`) would fix both issues.

4. **Minor: missing `isinstance(rule, dict)` guard** in the new loop (`tests/test_renovate_config.py:12`) — every other `packageRules` iteration site in `validate_renovate_config.py` guards against non-dict entries. Inert today, but inconsistent style and a latent `AttributeError` risk instead of a clean assertion failure.

5. **Minor: scope is broader than necessary.** The new rule matches `matchManagers: ["custom.regex"]` with no `matchDatasources`/`matchPackageNames`, unlike every AIPCC-specific sibling rule, so it also technically applies to the unrelated ODH `quay.io/opendatahub` BASE_IMAGE manager (`customManagers[1]`). Currently a no-op since that manager is pinned to the literal tag `"latest"` (no minor/patch axis), and it does match the scope issue #4374 itself suggested — just flagging that it's undocumented. (I checked whether folding it into the adjacent AIPCC-only grouping rule would be a cleaner fix instead — it wouldn't: that would silently narrow `separateMinorPatch` away from `registry.redhat.io/rhai/*` images, which have real deps on `rhoai-3.3`, re-introducing the exact masked-patch bug for those.)

6. **Nit, pre-existing, not introduced here:** the comment `// Re-enables the major/minor updates that packageRules[4] blocks by default.` was already wrong before this PR (the referenced rule was at index 11, not 4) and drifts one further after this PR's insertion. Purely cosmetic, no functional effect — mentioning only because this PR touches the same array's ordering.

Nice fix — the `separateMinorPatch` mechanism itself checks out under real execution. Mostly suggesting the test/validation follow-up above so a future regression in this area doesn't go unnoticed.

ide-developer (commented):
## Review summary

The core fix is sound. I reproduced the exact scenario from #4374 against the real registry (`quay.io/aipcc/base-images/cpu` pinned at `3.4.1-1780598682`, real patch `3.4.2` and minor `3.5.0` both available) by running the actual `renovate:43` image locally with this PR's rule shape. `separateMinorPatch: true` correctly produces two independent update candidates, and — worth calling out since it wasn't obvious from the docs alone — the existing `groupName: "konflux BASE_IMAGE {{{depNameSanitized}}}"` grouping rule does **not** collide them onto one branch: Renovate auto-prefixes the patch bucket's `groupSlug` (`renovate/patch-konflux-base_image-...` vs `renovate/konflux-base_image-...`), so the fix is safe on branches where major/minor is unblocked (`main`) as well as where it's blocked (release branches). Good fix, correctly scoped to the manager the issue describes.

The findings below are about the verification/test surface around this change, not the runtime behavior itself — non-blocking, but worth a follow-up.

1. **CI never actually exercises this manager.** `.github/workflows/validate-renovate-config.yaml`'s `validate-renovate-dry-run` job — the only check that runs live Renovate — is skipped for fork PRs (confirmed via this PR's own checks: it shows `skipping`, since the head repo is a fork). Even on same-repo PRs, `scripts/ci/validate_renovate_dry_run.py:187` hardcodes `RENOVATE_ENABLED_MANAGERS=github-actions` for every scenario, so `custom.regex` is never exercised end-to-end by CI at all, for anyone. A future regression to `separateMinorPatch`'s effect would ship with zero live signal.

2. **The new invariant bypasses this file's own validation pattern**, creating a split-brain check. Every other semantic invariant here (MintMaker policy, PR-title prefix, CentOS pin, ODH BASE_IMAGE policy) is enforced inside `scripts/ci/validate_renovate_config.py`'s `validate_config()`, contributing a descriptive string to `errors`. This PR instead adds a bare `assert any(...)` directly in `tests/test_renovate_config.py` (lines 10-13), so `validate_config()` was never taught the new rule (confirmed: `separateMinorPatch` doesn't appear anywhere in `validate_renovate_config.py`). That matters concretely because `.github/workflows/validate-renovate-config.yaml`'s "Validate renovate.json5 semantics" step runs `python scripts/ci/validate_renovate_config.py` **directly**, not via pytest — that CLI entrypoint (and anyone using it standalone, e.g. against the RHDS fork's config) will print `OK` even if this rule is later removed or broken.

3. **The new assertion's predicate isn't actually selective, and gives no diagnostic on failure.** `rule.get("matchManagers") == ["custom.regex"]` matches 8 of the file's 20 `packageRules` (verified by parsing the live config) — only the `and rule.get("separateMinorPatch") is True` clause does the real discriminating work. If a future edit ever moved that flag onto the wrong one of those 8 rules, this check would still pass. It's also a bare `assert` with no message, unlike the `assert errors == [], f"..."` immediately above it — a failure here just shows `assert False` with no context. Matching by `description` (the established idiom elsewhere in this file, e.g. `PREFIX_RULE_DESCRIPTION`, `CENTOS_STREAM_RULE_DESCRIPTION`) would fix both issues.

4. **Minor: missing `isinstance(rule, dict)` guard** in the new loop (`tests/test_renovate_config.py:12`) — every other `packageRules` iteration site in `validate_renovate_config.py` guards against non-dict entries. Inert today, but inconsistent style and a latent `AttributeError` risk instead of a clean assertion failure.

5. **Minor: scope is broader than necessary.** The new rule matches `matchManagers: ["custom.regex"]` with no `matchDatasources`/`matchPackageNames`, unlike every AIPCC-specific sibling rule, so it also technically applies to the unrelated ODH `quay.io/opendatahub` BASE_IMAGE manager (`customManagers[1]`). Currently a no-op since that manager is pinned to the literal tag `"latest"` (no minor/patch axis), and it does match the scope issue #4374 itself suggested — just flagging that it's undocumented. (I checked whether folding it into the adjacent AIPCC-only grouping rule would be a cleaner fix instead — it wouldn't: that would silently narrow `separateMinorPatch` away from `registry.redhat.io/rhai/*` images, which have real deps on `rhoai-3.3`, re-introducing the exact masked-patch bug for those.)

6. **Nit, pre-existing, not introduced here:** the comment `// Re-enables the major/minor updates that packageRules[4] blocks by default.` was already wrong before this PR (the referenced rule was at index 11, not 4) and drifts one further after this PR's insertion. Purely cosmetic, no functional effect — mentioning only because this PR touches the same array's ordering.

Nice fix — the `separateMinorPatch` mechanism itself checks out under real execution. Mostly suggesting the test/validation follow-up above so a future regression in this area doesn't go unnoticed.

jiridanek (approved):

PR #4393, opened 2026-08-16
spike(RHAIENG-6564): hermetic build feasibility for public-PyPI Jupyter images

ayush17:
## Summary

Spike investigation for hermetic builds using public PyPI packages instead of RHAI internal index.

**Jira:** [RHAIENG-6564](https://redhat.atlassian.net/browse/RHAIENG-6564)

## Setup Completed

| Step | Status |
|------|--------|
| Create test image (jupyter/testing) | ✅ |
| Remove CUDA/ROCM (CPU-only) | ✅ |
| Regenerate locks with public PyPI | ✅ 113 packages |
| Create Tekton PipelineRun | ✅ |
| Run prefetch | ✅ 270MB pip deps |

## Key Findings

- ✅ All JupyterLab dependencies available on public PyPI
- ✅ Prefetch from public PyPI works (270MB)
- ❌ `pandoc-rhai` not on public PyPI (PDF export needs alternative)

## To Trigger Konflux Build

Comment on this PR:
```
/build-testing-spike
```

## Files Changed

- `jupyter/testing/ubi9-python-3.12/` - Test image for spike
- `.tekton/odh-workbench-jupyter-testing-cpu-py312-ubi9-pull-request.yaml` - Tekton pipeline
- `scripts/lockfile-generators/create-requirements-lockfile.sh` - Added testing to PUBLIC_INDEX_PROJECTS

## Next Steps (Post-Konflux Build)

- [ ] Validate hermetic build succeeds
- [ ] Test sdist source-build path
- [ ] Compare with RHDS rhoai-2.25
- [ ] Run Conforma validation
- [ ] Document final recommendation

[RHAIENG-6564]: https://redhat.atlassian.net/browse/RHAIENG-6564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added a Jupyter Notebook environment based on Python 3.12 and UBI9.
  * Added Kubernetes deployment support with notebook access on port 8888, health checks, and resource settings.
  * Added configurable notebook startup options, including port, base URL, root directory, and launch arguments.
  * Added CPU image configurations and locked dependency versions from public PyPI.
* **Tests**
  * Added automated validation for Python, JupyterLab, and key notebook dependency versions.
* **Documentation**
  * Documented source-distribution builds, required compilation tools, expected build times, and testing status.


jupyter/testing/ubi9-python-3.12/SPIKE-NOTES.md
jiridanek, line 1: @coderabbitai there's lots of to go through here; first of all, I think this doc should go into adr directory, but that's the least important bit, lets focus on the substantive aspects, rather

## The motivation for the work is https://redhat.atlassian.net/browse/RHAISTRAT-1482

The problem being solved there is that AIPCC index-based workbench container images _must not_ be extended by a customer with PyPI wheels, because there's a risk of hard-to-diagnose ABI incompatibilities, especially in the CUDA and ROCm indexes, which have a build of gpu-enabled pytorch in them. That no-mixing policy is stated _somewhere_.

There seem to be essentially two ways of getting the desired images built. First, have hermeto/cachi2 fetch sdists from pypi and compile everything from sources. This does not violate conforma rules. It is not completely trivial to set up, though. For example, packages like cryptography, which needs rust to build, and rustup can't be casually installed from binaries either... The strat says "minimal", as in Jupyterlab, Elyra, Kyle, ..., we went through this with Kezia already. No predictive-ml packages.

The second way is to use aipcc cpu index, install packages from there, and hope there will not be pypi collisions. In order not to use aipcc base image, we'd need to use `auditwheel` to turn their wheels into fat wheels, maybe in a separate dockerfile multistage stage. Or simply use their base. This is discussed a bit in https://docs.google.com/document/d/1l5zlOjNHffV5RYGqheHSa_giIRXGJl5ycn1UKrQJHvY/edit?tab=t.0, and there is a slack thread where me and @atheo89 participated https://redhat-internal.slack.com/archives/C0B1JEZ09CY/p1779219911788709, discussing with @EmilienM.

jiridanek: > with an explicit analysis of public-PyPI collision risk

@coderabbitai think long and hard and perform initial risk analysis of mixing together wheels from https://packages.redhat.com/api/pypi/public-rhai/rhoai/3.6-EA1/cpu-ubi9/simple/ and PyPI, in a RHEL 9.8 container image.

jiridanek: Cursor somehow fished out https://docs.google.com/document/d/1_qM5Ud0N0f-jFa2HnsKdm2GhRRp-c6WLl1qVlHR8jh4/edit?tab=t.0#heading=h.wth56qst3zn7, I completely forgot about that one!

**Do not mix AIPCC `cpu-ubi9` wheels with PyPI in the same venv.** CPU is milder than CUDA/ROCm, but it is still the hybrid the docs already rejected.

I pulled the three Google Docs with `gws drive files export`. Together they close the loophole the spike was drifting toward:

- [Community-image ADR](https://docs.google.com/document/d/1_qM5Ud0N0f-jFa2HnsKdm2GhRRp-c6WLl1qVlHR8jh4/edit): the index is an image property. Mixing AIPCC-built and PyPI-built native packages “cannot be made safe.” Configurable `PIP_INDEX_URL` was rejected for that reason.
- [RHAISTRAT-1482 Alternative 1](https://docs.google.com/document/d/1i1Xv24RIe13wmUibdTF_QIcvBOvnYL4tzQTVU6XsZ7U/edit): community images are hermetic **without any AIPCC Python bits**.
- [RHAIBI assessment](https://docs.google.com/document/d/1l5zlOjNHffV5RYGqheHSa_giIRXGJl5ycn1UKrQJHvY/edit): Option A is `FROM` RHAIBI, overwrite `pip.conf`, compile Jupyter from sdist. CUDA RPM + PyPI-vendored CUDA is a different problem and already worked in RHOAI ≤3.3.

What looks like a cheap CPU shortcut is a different architecture:

| Image | Python in the venv | Mixing risk |
|---|---|---|
| Community (intended) | PyPI sdists only | None at image layer |
| Hybrid | AIPCC `linux_*` wheels, then customer `pip install` from PyPI | Policy-forbidden |

AIPCC CPU wheels are `cp312-cp312-linux_x86_64`, not manylinux. They dynamically link RHEL RPMs. Current locks:

- **minimal Jupyter:** 16 native / 97 pure — Jupyter runtime only (`pyzmq`, `tornado`, `cffi`, `aiohttp`, `uv`, …). No numpy/scipy/pandas/torch.
- **datascience:** 51 native — includes numpy, scipy, pandas, pyarrow, grpcio, cryptography.

On minimal, a user `pip install numpy pandas torch` often *appears* to work because that is a separate C-API island. The remaining failures are still real: `pip install -U pyzmq`, two `libstdc++` copies, RPM vs vendored OpenSSL/libffi, and pip silently preferring a newer PyPI version of the same package. Datascience mixing is the classic numpy C API / pyarrow class. Independently, RHEL 9.8 is glibc 2.34, so some PyPI `manylinux_2_38` wheels will not load even with no AIPCC Python content.

“Stick with AIPCC” only validates the **secure** tier (Conforma). It does not satisfy RHAISTRAT-1482. The community image still needs the sdist path.

@coderabbitai Cursor says: I put the collision classes, lockfile inventory, and gdoc constraints in [AIPCC CPU × PyPI mixing risk](/Users/jdanek/.cursor/projects/Users-jdanek-IdeaProjects-notebooks/canvases/aipcc-cpu-pypi-mixing-risk.canvas.tsx) so you can open it beside the chat.

I do respect that some PyPI binary packages (customers will be free to install PyPI binaries, so that there's less friction for them in a showcase environment; we in Red Hat can't ship somebody else's binaries downloaded over the net) may be incompatible with RHEL 9's manylinux version, but that's simply something the `pip install` can already detect, not a huge problem.

How come that ABI problems are still expected? I believed that with the exception of Torch, no other python packages share their .so libs between wheels, therefore there should not be a C level incompatiblity.

jiridanek: **@CodeRabbitAI walked back the right thing, then agreed with you a bit too far.** The architecture conclusion is still correct. The “only Torch shares `.so` files, so there is no C-level incompatibility” model is too narrow.

Your question was: if packages do not share libraries the way Torch does, why expect ABI failures on a **minimal** AIPCC CPU image plus customer PyPI `numpy`/`pandas`/`torch`?

That coexistence case is **not** expected to blow up at the C ABI. Isolated extensions in one process do not conflict just because one wheel came from AIPCC and the other from PyPI. CodeRabbit is right to treat two `libstdc++` copies and RPM-vs-vendored OpenSSL/`libffi` as **possible loader classes**, not as facts for this lock set. `manylinux_2_38` vs glibc 2.34 is also just pip tag selection, not an AIPCC/PyPI mix bug.

The remaining C-level cases it listed are the real ones, in order:

1. **Replacement, not coexistence.** `pip install -U pyzmq` / `cffi` / `tornado` overwrites the AIPCC `.so` that Jupyter was locked against. That is the practical failure mode on minimal, and it is not “two unrelated wheels sharing a library.”
2. **Consumers of another package’s native C API.** NumPy is the main non-Torch example. SciPy, pandas, pyarrow, sklearn, and customer Cython extensions bind the NumPy C API chosen at **build** time. That is a datascience-image problem. It is not in the current minimal lock (no numpy).
3. **A shared ELF object, including system libs.** “Wheels do not ship a shared `.so`” is not the same as “they do not both `DT_NEEDED` `libstdc++.so.6` / `libgomp`.” AIPCC wheels are thin and link RHEL RPMs; manylinux usually still uses the system C++ runtime and only vendors the rest with hashed names. Dual `libstdc++` can happen if RHAIBI resolves gcc-toolset `libstdc++` for AIPCC extensions and `/usr/lib64` for PyPI ones. That needs `ldd`/`readelf` on the actual wheels, which nobody has run.

Nit: it named `cryptography` as something already in the image. The **minimal** CPU lock has `cffi` and `pyzmq`, not `cryptography`. Replacement of `cffi` still matters because `argon2-cffi-bindings` is in that lock.

Where it should **not** have said “that is correct”: Torch is the loud example of **project-level** shared libs (`libtorch`, CUDA wheels). It is not the only shared-native ABI. NumPy’s C API is the other standard one. Protobuf/gRPC and Arrow show up on datascience. Those do not put C-level landmines under a minimal Jupyter process that then `pip install`s an unrelated PyPI ML stack.

The last paragraph is the part to keep. Community images stay **PyPI sdist-only** because the ADR makes the index an image property and mixing unsupportable, not because every mixed pair will segfault. “Many customer installs will work” is compatible with “do not ship the hybrid.”

jiridanek: @ayush17 seems to me @coderabbitai is saying it may very well work, and it is only obsessed by the "ADR policy". That's very easy to fix, we edit ADR, and add exception to policy, lol Still not sure if auditwheel is worth it, though.

PyPI itself does not provide any guarantees. And compiling from sources and mixing that with customer-added precompiled wheels by itself is also risky (in a FUD sense) thing to do.

PR #4411, opened 2026-08-20
fix(base-images/copr): patch thrift rebuild for EL9

atheo89:
Use Copr custom builds to rewrite Fedora-only thrift build steps that break on EL9, including the openssl build requirement and the Python setuptools error import. Also fail the ODH base-image install early if libthrift-0.24.0.so is still missing for pyarrow at runtime.

Once this PR be merged we will rebuild the ODH base images. 

## How Has This Been Tested?

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added support for literal spec-file replacements in package build manifests.
  * Custom builds can now apply spec replacements and optionally skip tests.
  * Added the Thrift 0.24.0 package with compatibility adjustments.
* **Bug Fixes**
  * Improved EL9 compatibility for packages requiring dependency and Python import rewrites.
  * Added validation that the required Thrift library is installed on CentOS.
* **Documentation**
  * Documented spec replacements, custom builds, and configuration examples.

### DIFF
--- a/base-images/copr/src/copr_rebuild/copr_client.py
+++ b/base-images/copr/src/copr_rebuild/copr_client.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import json
 import logging
 import os
 import shlex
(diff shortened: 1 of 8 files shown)


atheo89:
The ODH GHA builds will be red as this doesn't fix directly the issue. Should be first merged and then update the odh conf files t oget the these updates

jiridanek (approved):
/lgtm

base-images/copr/src/copr_rebuild/copr_client.py
```diff
@@ -22,17 +23,51 @@
...
-set -Eeuxo pipefail
-curl --retry 5 --retry-delay 3 --fail -L -o original.src.rpm {srpm_url}
-rpm2cpio original.src.rpm | cpio -idmv
-rm original.src.rpm
-sed -i '/^%check/a exit 0' *.spec
-"""
-
 _SCRIPT_CHROOT = "fedora-rawhide-x86_64"
-_SCRIPT_BUILDDEPS = "curl rpm cpio"
+_SCRIPT_BUILDDEPS = "curl rpm cpio python3"
```
ide-developer, line 27: `_SCRIPT_BUILDDEPS` unconditionally adds `python3` even for custom builds that only use `skip_tests` (no `spec_replacements`), wasting an unneeded package install in the Copr script chroot.

A manifest entry with only `skip_tests: true` (e.g. autoconf, hdf5) goes through `submit_custom_build`, and its `copr-cli buildcustom --script-builddeps 'curl rpm cpio python3'` installs python3 in the fedora-rawhide-x86_64 helper chroot even though the generated script never invokes python3 (the `if spec_replacements:` block is skipped in `build_custom_build_script`), adding unnecessary setup time to every skip_tests-only build.

base-images/copr/src/copr_rebuild/copr_client.py
```diff
@@ -351,18 +407,37 @@ def submit_all_waves(
             wave_anchor: int | None = None
 
             for i, (name, url) in enumerate(wave_items):
```
ide-developer, line 409: `submit_all_waves` duplicates the custom-build-vs-normal-build call selection logic across four near-identical call sites (`i==0`/`use_custom_build`, `i==0`/else, `elif use_custom_build`, else) instead of computing shared kwargs once.

Adding a third build-customization dimension (or changing shared kwargs like `timeout`) requires editing four separate call sites instead of one, increasing the chance a future maintainer updates only some of them and introduces a divergence between the first-in-wave and subsequent-in-wave code paths.

base-images/copr/src/copr_rebuild/packages.yaml
```diff
@@ -43,3 +43,18 @@ packages:
   - name: zeromq
     nvr: zeromq-4.3.5-22.fc43
     note: "provides libzmq >= 4.3.5 for pyzmq"
+  - name: thrift
+    nvr: thrift-0.24.0-1.fc46
```
ide-developer, line 47: Top-level `koji_tag: f43` (line 6) is inconsistent with this new thrift package's `nvr: thrift-0.24.0-1.fc46`, with no comment explaining the mismatch.

A future maintainer reading the manifest could reasonably (but incorrectly) assume `koji_tag` constrains which Koji tag builds are pulled from, and add a new package NVR assuming `f43` availability — when in practice `koji_tag` is unused by `koji_client.get_package_metadata` (which queries Koji directly by NVR), so the field is effectively documentation-only and now out of sync with at least this entry.

ide-developer (commented):
Automated review (high effort): downloaded the actual Fedora `thrift-0.24.0-1.fc46` SRPM/spec and upstream source tarball to verify the new `spec_replacements` text matches byte-for-byte, and traced all call sites of the changed functions. No confirmed correctness bugs — the core new mechanism (spec-replacement script generation, schema/model consistency, error handling, key consistency) all checks out. The inline comments below are cleanup/efficiency observations only, non-blocking.

PR #4422, opened 2026-08-21
RHAIENG-5451: enforce per-flavor required-environments during lock generation

jiridanek:
## Description

`uv pip compile --universal` never verified per-architecture wheel coverage: `pyproject.toml`'s `[tool.uv] environments` has no `platform_machine` axis, so a package version missing wheels for one architecture was silently accepted instead of triggering a fallback search or a hard failure at lock time.

This let #4408's automated lockfile refresh silently lock `xxhash==4.0.1` for `jupyter/trustyai`'s CPU image, even though that version has no `ppc64le`/`s390x` wheels in the production AIPCC index — breaking the Konflux build on s390x. See [RHAIENG-7088](https://redhat.atlassian.net/browse/RHAIENG-7088) for the full investigation (Slack threads, real build failure log, root-cause reproduction against the live AIPCC index).

**Fix**: temporarily inject a per-flavor `required-environments` (uv 0.11.15+, [astral-sh/uv#19378](https://github.com/astral-sh/uv/pull/19378)) into each `pyproject.toml` before `uv pip compile`, restoring the file afterward — `required-environments` is a project-only setting with no CLI flag. This makes uv validate wheel coverage across every architecture the flavor actually ships:

| Flavor | Required architectures |
|---|---|
| cpu | x86_64, aarch64, ppc64le, s390x |
| cuda | x86_64, aarch64 |
| rocm | x86_64 |

Regenerating `trustyai`'s CPU lock with the fix correctly splits `xxhash` per architecture — `4.0.0` (last version with full 4-arch coverage) for `ppc64le`/`s390x`, `4.0.1` elsewhere — verified live against the real AIPCC index. Swept all other CPU-flavor rh-index images (`datascience`, `minimal`, `codeserver`, both `jupyter/` and `runtimes/`) and found no other latent per-architecture gaps; those locks are unchanged.

This was originally implemented on the `jd/uv_upgrade` branch ([RHAIENG-5451](https://redhat.atlassian.net/browse/RHAIENG-5451), closed as done) but never merged — that branch was based on `rhds/rhoai-2.25` and shares no git history with this repo's `main`, so a normal rebase wasn't possible. Reimplemented here against current `main`, reusing the same design (confirmed via `git diff` against the original branch's isolated `pylocks_generator.py` change).

## How Has This Been Tested?

- `ruff check` / `ruff format --check` on the modified files — clean.
[...]

PR #4423, opened 2026-08-21
[Alternative to #4422] RHAIENG-5451: commit required-environments statically into pyproject.toml

jiridanek:
## Description

**This is an alternative to #4422, for comparison — not meant to be merged alongside it.** Both solve the same problem ([RHAIENG-7088](https://redhat.atlassian.net/browse/RHAIENG-7088)): `pyproject.toml`'s `environments` has no `platform_machine` axis, so `uv pip compile --universal` never verified per-architecture wheel coverage, letting #4408's lockfile refresh silently lock `xxhash==4.0.1` for `jupyter/trustyai`'s CPU image despite it having no `ppc64le`/`s390x` wheels — breaking the Konflux build on s390x.

- **#4422**: injects `required-environments` transiently around each `uv pip compile` call, per flavor, then restores the file. Fully correct for every flavor including multi-flavor files, but invisible in git history — you have to read `scripts/pylocks_generator.py` to know the requirement exists.
- **This PR**: commits `required-environments` permanently into each `pyproject.toml`. Zero code changes to `scripts/pylocks_generator.py`. Human-visible, and enforced even for a manual `uv pip compile` run outside the generator script. Trade-off: `jupyter/minimal/ubi9-python-3.12` is the one `pyproject.toml` shared by three flavors (cpu/cuda/rocm) with different architecture requirements, and a single static value can't be correct for all three.

### The jupyter/minimal caveat, verified empirically

Tried committing the CPU flavor's full 4-arch set (x86_64/aarch64/ppc64le/s390x) to this shared file — broke both CUDA and ROCm lock generation:

```
Because pandoc-rhai==3.9.0.2+rhaiv.1 has no `platform_machine == 's390x'`-compatible wheels
```

`pandoc-rhai` is a shared, non-GPU dependency with no `ppc64le`/`s390x` wheels in the `cuda13.0-ubi9`/`rocm7.14-ubi9` channels (only in `cpu-ubi9`). Tried narrowing to x86_64+aarch64 next — ROCm *still* failed, because `rocm7.14-ubi9` doesn't even carry `pandoc-rhai` for aarch64 (matches README.md: ROCm Minimal is x86_64-only). The only value safe for all three flavors sharing that file is the **intersection** — x86_64 only — which means this file doesn't get full per-flavor coverage checking under this approach. Documented in a code comment pointing at #4422 as the way to get full coverage there.

## How Has This Been Tested?

- Regenerated all 16 rh-index CPU/CUDA/ROCm-flavor image locks (all succeeded) and diffed every one against pre-fix backups: the **only** real content change anywhere is `jupyter/trustyai`'s CPU lock, which now correctly splits `xxhash==4.0.0` (ppc64le/s390x) vs `xxhash==4.0.1` (elsewhere) — identical result to #4422. No other latent per-architecture gaps found across the other 15 projects.
- `uv run pytest tests/unit/scripts/test_pylocks_generator.py` — 44 passed, unaffected (no script changes in this PR).
- `ruff check` — clean (no Python changes; only `pyproject.toml`/lock file edits).

## Merge criteria:

[RHAIENG-7088]: https://redhat.atlassian.net/browse/RHAIENG-7088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Improvements**
  * Improved dependency validation across supported Linux architectures during environment setup.
  * Added architecture-aware package resolution for CPU environments, including support for `ppc64le` and `s390x`.
  * Updated dependency lock metadata across Python 3.12 environments to reflect the latest resolution cutoff.
  * Package versions and resolutions remain unchanged except for the platform-specific `xxhash` adjustment.


jupyter/minimal/ubi9-python-3.12/pyproject.toml
```diff
@@ -20,6 +20,20 @@ dependencies = [
...
+# be correct for all three. Tested committing the CPU flavor's 4-arch set
+# here and it broke both CUDA and ROCm lock generation, because pandoc-rhai
+# -- a shared, non-GPU dependency -- has no ppc64le/s390x wheels in the
+# cuda13.0-ubi9/rocm7.14-ubi9 channels; then tried x86_64+aarch64 and ROCm
+# still failed, because rocm7.14-ubi9 only has pandoc-rhai for x86_64 (matches
+# README.md: ROCm Minimal is x86_64-only). Getting full per-flavor coverage
+# checking here requires the dynamic per-flavor injection in
+# opendatahub-io/notebooks#4422 instead.
+required-environments = [
+    "sys_platform == 'linux' and platform_machine == 'x86_64'",
```
atheo89, line 35: I guess do we need the rest three archs here?

jiridanek: we can't have them, because then the lock fails on pandoc, as the comment above this explains cc @coderabbitai

jiridanek: @atheo89 I don't want to introduce the entire _dynamic_ approach simply because one exceptional image (jupyter minimal)

atheo89 (approved):
I like this version more!

PR #4430, opened 2026-08-24
docs(rosa-hcp-provision): record 3.6 EA Kyverno and workbench spawn learnings

jiridanek:
## Summary

Update `.agents/skills/rosa-hcp-provision/` with what failed and what worked while installing RHOAI 3.6.0-ea.1 on an arm64 ROSA HCP cluster (`jd-arm64-36`). Kyverno must **append** `pull-secret-quay` (not replace `imagePullSecrets`), clone it into `openshift-ingress` on first apply, and workbenches need `importMode=PreserveOriginal` before spawn.

Incorporates PR review feedback: Kyverno Helm chart **3.9.0** (app v1.19.0) pinned with optional provenance verify; explicit DSC v2 component list aligned with `dsc-minimal-workbenches.yaml`; hand-copy rotation notes for `openshift-ingress` and `openshift-marketplace`; SSO identifiers sanitized from login notes.

## Root cause

Three independent EA/HCP failures looked like “just wait for the pull”:

1. **Dashboard 403** — `kube-auth-proxy` in `openshift-ingress` was `ImagePullBackOff` on `quay.io/rhoai/odh-kube-auth-proxy-rhel9`. `sync-secrets` did not clone into that namespace, and Kyverno **generate match/clone is immutable** after create, so the namespace list cannot be patched later.
2. **Workbench `authentication required`** on `image-registry.openshift-image-registry.svc` — `add-imagepullsecrets` used `patchStrategicMerge`, which **replaced** the pod’s `imagePullSecrets` with only `pull-secret-quay` and dropped OpenShift’s `*-dockercfg-*`. Quay sidecars pulled; the notebook image from the internal registry did not.
3. **`Exec format error`** — ImageStream tags still imported `Legacy`/amd64 on HCP (amd64 API server) even when the manifest list had arm64. `PreserveOriginal` + re-import is required; manifest-list `skopeo inspect` is not enough.

Also recorded: Kyverno admission-controller RBAC for generate validation (observed on chart 3.9.0); 3.6 EA DSC is `datasciencecluster.opendatahub.io/v2` with `aipipelines` (GA YAML is a silent no-op); one `g5g.2xlarge` replica is one GPU.

## Quay credentials

Cached `quay.io/rhoai` robot `rhoai+devops_rhoai_readonly_bot` was dead during validation (`Could not find robot…`). The skill documents both paths:

- **Shared/long-lived clusters:** dedicated read-only robot or service account (preferred).
- **Ephemeral personal dev clusters:** personal Quay login with `quay.io/rhoai` org read access — wire the same credential under both `quay.io` and `quay.io/rhoai` keys either way; verify with `skopeo inspect` before creating cluster Secrets.

ROSA HCP reverts global `openshift-config/pull-secret` (OCPBUGS-23901), so the path is `pull-secret-quay` + Kyverno clone/mutate.

## Related issue

Not a product fix. Related context: OCPBUGS-23901 (ROSA HCP reverts global `pull-secret`), OCPBUGS-73844 / OCPBUGS-74567 / [RHOAIENG-82528](https://redhat.atlassian.net/browse/RHOAIENG-82528) (`importMode=Legacy` on HCP).

## Test plan

- [x] Installed RHOAI 3.6.0-ea.1 on ROSA HCP `jd-arm64-36` (OCP 4.21.0, 2× `m6g.2xlarge` + 1× `g5g.2xlarge`)
- [x] Dashboard `https://rh-ai.apps...` returned 302 to OAuth after `kube-auth-proxy` could pull
- [x] JSON Patch append left both `*-dockercfg-*` and `pull-secret-quay` on the workbench pod
- [x] `jupyter-pytorch-llmcompressor:3.6` reached **2/2 Running** on the T4G node after `PreserveOriginal` re-import
- [ ] `make test` — docs-only skill markdown; no code/image changes
- [ ] Full arm64-rosa-gpu-smoke Phase 3 (`nvidia-smi` / nbconvert / Elyra UI) — not run this session (CI)

<details>
<summary>Plan</summary>

Update `.agents/skills/rosa-hcp-provision/` from the 2026-08-24 EA install session so the next agent run does not re-derive:

- append-not-replace Kyverno `imagePullSecrets`
- DSP `namespaceSelector` + `openshift-ingress` on the first `sync-secrets` apply
- DSC v2 / `aipipelines` with every component key explicit
- `PreserveOriginal` before spawn
- dashboard 403 vs OAuth 302
- one-GPU Pending vs ImagePullBackOff
- Kyverno chart 3.9.0 pin + credential rotation for hand-copied Secrets

</details>

<details>
<summary>Trajectory</summary>

Provisioned ROSA HCP cluster `jd-arm64-36` (shared OCM org 7081269). Workers: 2× `m6g.2xlarge` arm64 + 1× `g5g.2xlarge` T4G. NFD + GPU Operator `gpu-operator-certified.v25.3.4`; driver 2/2 Ready.

Installed RHOAI 3.6.0-ea.1 from Slack `#rhoai-build-notifications` (`rhoai-fbc-fragment@sha256:b78eedb8…`, channel `beta`, Manual InstallPlan). Dead cached robot → personal Quay login for both `quay.io` and `quay.io/rhoai` keys on this ephemeral dev cluster.

Kyverno via Helm chart **3.9.0** (app v1.19.0). `sync-secrets` apply failed until `kyverno-secret-manager` also aggregated to the **admission** controller. `replace-image-registry` scoped to `^registry\.redhat\.io/rhoai/`; ImageStream `patchesJson6902` must use **un-doubled** JMESPath quotes on 1.19. DSC applied as `datasciencecluster.opendatahub.io/v2` with all v2 component keys explicit. Service Mesh 3 `v3.4.1` for Gateway API; DSC Ready; dashboard hostname `https://rh-ai.apps…`.

Dashboard still returned **403** until `pull-secret-quay` was hand-copied into `openshift-ingress` (generate rule immutable). Spawn failures traced through DSP namespaceSelector, `Legacy` importMode, then `patchStrategicMerge` dropping `*-dockercfg-*`. JSON Patch append + pod delete + `PreserveOriginal` re-import → `cudaarmtorch-0` 2/2 Running on T4G.

</details>

## Summary by CodeRabbit

- **Documentation**
  - Clarified ROSA HCP account, organization ID, and account-number verification requirements.
  - Expanded arm64 troubleshooting for ImageStream imports, pull secrets, dashboards, workbenches, GPU capacity, and Kyverno policy setup.
  - Updated prerelease RHOAI installation guidance for Kyverno, credentials, secret rotation, component configuration, marketplace access, and arm64 validation.
  - Documented verified RHOAI 3.6 arm64 results and clarified version-specific schema, pipeline, and compatibility requirements.
  - Updated helper-script, ImageStream variable, and Quay credential guidance.


.cursor/skills/rosa-hcp-provision/SKILL.md
```diff
@@ -118,9 +118,14 @@ rh-aws-saml-login iaps-rhods-odh-dev -- aws sts get-caller-identity
...
-Your personal `jdanek@redhat.com` SSO may map to a different org. The RHOAI shared org requires:
+Your personal `jdanek@redhat.com` SSO maps to the **wrong** OCM org
+(`6340056`). The RHOAI shared org is **account 7081269** / OCM
+`1pwwsfazToamNegaehP6eaDg80K`. Login as **`rhoai-jdanek`** (or the
+`rhoai-<user>` invite for whoever is running this), not the personal
+Red Hat email:
+
 1. Open **incognito** tab at https://console.redhat.com/openshift
-2. Login with account showing **Account number: 7081269** in top-right
+2. Sign in as `rhoai-jdanek` — top-right must show **Account number: 7081269**
```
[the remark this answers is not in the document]
jiridanek: @coderabbitai made changes, recheck

.cursor/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -485,7 +572,59 @@ oc --context "$CLUSTER_CONTEXT" delete csv "rhods-operator.<version>" -n redhat-
...
+  components:
+    dashboard:
+      managementState: Managed
+    workbenches:
+      managementState: Managed
+    aipipelines:
+      managementState: Removed
+    kserve:
+      managementState: Removed
+    modelmeshserving:
```
jiridanek: **DSC v2 component list mixes stale v1 keys with missing v2-only keys**

`modelmeshserving` and `codeflare` aren't part of the v2 component schema (they silently no-op — the exact trap the doc calls out one section earlier for `datasciencepipelines`→`aipipelines`), and several real v2-only components are omitted entirely (`feastoperator`, `llamastackoperator`, `mlflowoperator`, `sparkoperator`, `trainer`, `aigateway`, `mcplifecycleoperator`, `ogx` — see this repo's own `.agents/plugins/cluster-provisioning/skills/cluster-bot/fixtures/dsc-minimal-workbenches.yaml` for the real v2 key set). If any omitted key defaults to `Managed` rather than `Removed`, this "minimal footprint" DSC silently installs extra components on the EA cluster.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Compared against `.agents/plugins/cluster-provisioning/skills/cluster-bot/fixtures/dsc-minimal-workbenches.yaml` in this same repo, which is a genuine v2 DSC and has neither `modelmeshserving` nor `codeflare` keys, but does have `feastoperator`, `llamastackoperator`, `mlflowoperator`, `sparkoperator`, `trainer`, `aigateway`, `mcplifecycleoperator`, `ogx`.
- The doc's own prose a few lines earlier already warns that applying the wrong component key for your DSC schema version is a *silent no-op, not an error* — that exact failure mode applies here to `modelmeshserving`/`codeflare` under v2.
- Unverified: whether an *omitted* v2 component key defaults to `Managed` or `Removed` on 3.6-ea.1's DSC webhook/CRD defaulting. If it's `Managed`, the "everything except dashboard+workbenches Removed" minimal-footprint intent (stated for the sibling install-rhoai.md recipe) is silently violated for every component this YAML doesn't mention.
- Confidence: PLAUSIBLE — the key mismatch is directly verifiable in-repo; the concrete failure (extra components installing) depends on an unverified CRD default.
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -366,6 +442,17 @@ escaping getting mangled through both JSON-patch and Kyverno's JMESPath
 template layers — don't fight the patch. Re-`oc apply -f -` the whole
 corrected YAML document instead; that's reliable.
 
+**Generate-rule immutability:** after `sync-secrets` exists, Kyverno
+rejects edits to the generate `match`/`clone`/`preconditions`. If you
+omitted `openshift-ingress` on the first apply, **do not try to patch
+that list**. Copy `pull-secret-quay` into `openshift-ingress` by hand
```
jiridanek, line 490: **Hand-copied `pull-secret-quay` in `openshift-ingress` falls outside Kyverno's `synchronize: true` tracking**

This workaround Secret is created via a one-off `jq | oc apply`, not via Kyverno's `generate` controller, so it won't receive updates when `openshift-config/pull-secret-quay` (the `sync-secrets` clone source) rotates — and per step 5, this exact credential has already had to be rotated once this session (dead robot account).

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- `sync-secrets` (step 6) sets `generate.synchronize: true`, Kyverno's documented mechanism for pushing source-credential updates to every namespace the generate rule already tracks.
- A Secret created out-of-band via `oc apply` (as instructed here, and at step 7's analogous `openshift-marketplace` copy) has no Kyverno generate-tracking metadata, so it's invisible to that reconciliation loop.
- Failure scenario: next time the Quay credential needs replacing (already happened once — the dead `rhoai+devops_rhoai_readonly_bot`), every namespace under the real `sync-secrets` scope gets the new credential automatically, but `openshift-ingress`'s hand-copied one goes stale silently, and `kube-auth-proxy` regresses to the exact `ImagePullBackOff`/403 symptom this PR just finished root-causing — with no policy-level signal pointing at "the manual copy is stale."
- No re-sync step or staleness warning is documented anywhere else in the file for this specific workaround.
- Confidence: CONFIRMED mechanism (generate/synchronize behavior and manual-copy invisibility to it are both verifiable from the policy YAML itself); trigger timing (next rotation) is a real, already-recurring condition in this doc's own narrative.
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -234,19 +257,39 @@ kind: ClusterPolicy
...
+      # `oc secrets link` workbench SAs as the injection path — Pod
+      # admission is what kubelet reads.
+      - resources:
+          kinds: ["Pod"]
+          namespaceSelector:
+            matchLabels:
+              opendatahub.io/dashboard: "true"
     preconditions:
+      all:
+      - key: "{{ request.object.spec.imagePullSecrets || `[]` | length(@) }}"
```
jiridanek, line 332: **`init-pull-secret-quay`'s `length==0` branch is presented as part of "the verified fix" but was likely never exercised**

Under standard Kubernetes/OpenShift admission ordering, the built-in `ServiceAccount` admission plugin runs *before* `MutatingAdmissionWebhook` (Kyverno), and OpenShift auto-wires each SA's `<sa>-dockercfg-*` Secret into that SA's own `imagePullSecrets`. So by the time Kyverno evaluates this rule's `imagePullSecrets` length, a real workbench Pod almost certainly already has a non-empty list — meaning only `append-pull-secret-quay` (the sibling rule below) ever actually fires. The doc's own evidence further down ("Pod has only `pull-secret-quay`; the SA still has `*-dockercfg-*` **and** `pull-secret-quay`" / "Verified spawn: ... after **append-injection**") shows exactly that: append fired, not init.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Standard kube-apiserver admission-plugin ordering places `ServiceAccount` before `MutatingAdmissionWebhook` (the documented recommended `--enable-admission-plugins` ordering).
- OpenShift's ServiceAccount controller auto-generates a `<sa>-dockercfg-<hash>` Secret and links it into the SA's `.imagePullSecrets` (distinct from vanilla upstream Kubernetes) — this is the mechanism that lets OpenShift Pods pull from the internal registry without an explicit `imagePullSecrets` entry.
- Given both, a workbench Pod using any real project SA should already have a non-empty `imagePullSecrets` by the time Kyverno's webhook runs, making this precondition (`length == 0`) essentially unreachable in practice.
- Not a functional bug — `init` being unreachable is harmless, `append` alone covers the real case — but the doc's framing (intro: "after the Kyverno/importMode fixes below" verified end-to-end; later: "Verified spawn: ... after append-injection") presents both rules jointly as validated when only one branch has evidence behind it.
- Confidence: PLAUSIBLE — the admission-ordering and OpenShift SA-secret-wiring premises are well-established, but not verified live against this exact cluster's admission chain configuration.
</details>

.cursor/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -339,6 +411,10 @@ spec:
         # '' for a literal quote, matching the Pod/initContainer rules
         # above) has no escape processing at all, so `\.` survives as a
         # literal backslash-dot, which is what the regex actually needs.
+        # Kyverno 1.19 (2026-08-24): do **not** double the JMESPath quotes
+        # (`''^registry\.redhat\.io/rhoai/''`) — that is what 1.18
```
jiridanek: **This new comment mislabels the correct, currently-working JMESPath quoting as the broken pattern**

The pattern called out here as 1.18-only / SyntaxError-on-1.19 (`''^registry\.redhat\.io/rhoai/''`) is the *exact same* string used four lines below in the actual `value:` (and in the unchanged Pod/initContainer rules earlier in this same policy) — which this doc elsewhere calls "the exact deployed YAML... verified working live on the actual cluster." The doubled `''` is standard YAML escaping for a literal single quote inside a single-quoted scalar, required to produce the JMESPath string argument — it isn't "doubled JMESPath quotes."

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Checked lines 368 and 388 (unchanged Pod/initContainer rules) and line 421 (a few lines after this comment, same rule) — all three use the identical `''^registry\.redhat\.io/rhoai/''` / `''quay.io/rhoai/''` pattern.
- The surrounding unchanged prose immediately above this new comment already explains *why* single-quoting-with-doubled-quotes is correct here: "single-quoted (with '' for a literal quote...) has no escape processing at all, so `\.` survives as a literal backslash-dot, which is what the regex actually needs."
- The new comment then directly contradicts that explanation by calling the same pattern the broken one.
- Concrete risk: a future editor (human or agent) trusts this comment and "fixes" line 421 by removing the doubled quotes, which would either break YAML parsing or change the JMESPath string content, regressing a rule the doc says is load-bearing for EA-content ImageStream imports.
- Confidence: CONFIRMED — this is a direct textual contradiction verifiable by reading the lines in question.
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -138,6 +152,7 @@ metadata:
   labels:
     app.kubernetes.io/component: background-controller
     rbac.kyverno.io/aggregate-to-background-controller: "true"
+    rbac.kyverno.io/aggregate-to-admission-controller: "true"
```
jiridanek, line 175: **Admission controller now gets full CRUD on every Secret cluster-wide, broader than Kyverno's documented least-privilege split**

Adding `rbac.kyverno.io/aggregate-to-admission-controller: "true"` here grants Kyverno's admission-controller pod (which processes every admission request cluster-wide) `create`/`update`/`patch`/`delete` on *all* Secrets in the cluster, not just `get`/`list`/`watch`. The stated need is "generate validation," which should only require read access to confirm a generate rule's clone source exists.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Kyverno documents separate "view" and "manage" ClusterRole aggregation labels precisely so components that only need to read/validate don't also get write access.
- The doc's own justification a few lines above is narrowly about *validation* failing without this label, not about needing write access from the admission path.
- If this really is the minimum grant Kyverno 1.19 requires for `sync-secrets` to admission-validate, worth confirming upstream / scoping down (e.g. a read-only aggregation label if Kyverno exposes one) rather than reusing the background controller's full CRUD role.
- Risk: a compromised or misbehaving admission-controller process (exposed to every admission request cluster-wide — a much larger attack surface than the background controller) now has write/delete access to every Secret cluster-wide, not just RHOAI namespaces.
- Confidence: PLAUSIBLE — the broadened grant is verifiable directly from the YAML; whether a narrower grant would actually satisfy Kyverno 1.19's admission-time generate validation wasn't verified against a live cluster in this pass.
</details>

.cursor/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -631,11 +792,40 @@ still on `registry.redhat.io/rhoai/*`, and every tag has a
...
 
-## 12. Known gaps — explicitly not verified
+## 13. Workbench spawn (verified 2026-08-24 on `jd-arm64-36`)
+
+Kyverno **injects** `pull-secret-quay` onto Pods (step 6). Do not
+`oc secrets link` notebook SAs as the workaround — that hid the real
+bug (`patchStrategicMerge` replacing `imagePullSecrets`).
+
+After a policy fix, **delete the workbench pod**. `imagePullSecrets`
+cannot be patched on a running pod; `skipBackgroundRequests: true`
```
jiridanek: **`skipBackgroundRequests: true` is not why existing Pods keep their old `imagePullSecrets`**

The real reasons a running Pod doesn't pick up a Kyverno policy fix are (1) `spec.imagePullSecrets` is effectively immutable on a running Pod at the Kubernetes API level, independent of Kyverno, and (2) these are plain per-admission `mutate` rules with no "existing resource" reconciliation path. `skipBackgroundRequests` specifically controls whether `generate`/mutate-existing rules get evaluated during Kyverno's periodic background scan of already-existing resources — it has no bearing on a one-shot admission-time mutate rule for Pods.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Per Kyverno's docs, `skipBackgroundRequests` (which already defaults to `true` in recent Kyverno versions) governs whether `generate` and mutate-existing (`mutate.targets`) rules are re-evaluated during the background scan against already-existing resources — not ordinary admission-time `mutate` rules on newly-admitted Pods.
- Kubernetes itself doesn't allow updating `spec.imagePullSecrets` on a running Pod via the API regardless of any admission webhook, so "delete and recreate the Pod" would be necessary even with a Kyverno mutate-existing capability.
- Same claim repeats at lines ~453-454 ("existing Pods do **not** pick up `imagePullSecrets` changes — delete them (`skipBackgroundRequests: true`, and kubelet cannot mutate the field in place)"), so it's not a one-off phrasing slip.
- Practical impact is low (the doc's actual instruction — delete and recreate the Pod — is correct regardless of the reasoning given), but the misattribution could send a future reader down the wrong path if they try tuning `skipBackgroundRequests` expecting it to enable in-place updates.
- Confidence: PLAUSIBLE — Kyverno's documented semantics for this flag support the misattribution reading; not tested live against this exact ClusterPolicy's background-controller behavior.
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -214,7 +231,13 @@ spec:
...
       - key: "{{ request.object.metadata.name }}"
         operator: AnyIn
-        value: ["redhat-ods-applications", "redhat-ods-operator", "redhat-ods-monitoring"]
+        # openshift-ingress: RHOAI 3.6 Gateway dashboard's kube-auth-proxy
+        # lives here (quay.io/rhoai/odh-kube-auth-proxy-rhel9). Include it
+        # on the *first* apply — Kyverno generate match/clone is immutable
+        # after create, so adding this namespace later is rejected. Hit
+        # 2026-08-24: dashboard returned 403 instead of OAuth 302 while
+        # the proxy was ImagePullBackOff.
+        value: ["redhat-ods-applications", "redhat-ods-operator", "redhat-ods-monitoring", "openshift-ingress"]
```
jiridanek, line 276: **`openshift-ingress` added to the secret-clone/inject scope in three separate hardcoded lists, with no shared source and no note on the exposure it creates**

The same literal 4-element namespace array is now duplicated across `sync-secrets` (here), `init-pull-secret-quay` (line 277), and `append-pull-secret-quay` (line 315). `openshift-ingress` is a shared core-platform namespace, not an RHOAI-owned one — worth calling out explicitly that any workload with Secret-read RBAC there now also gets read access to the quay.io/registry.redhat.io pull credential.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Verified all three lists are character-for-character identical (`["redhat-ods-applications", "redhat-ods-operator", "redhat-ods-monitoring", "openshift-ingress"]`) at lines 240, 277, and 315 of this file.
- With no single source of truth, a future scope change (adding/removing a namespace) requires editing three places by hand — exactly the kind of drift this PR's own change (adding `openshift-ingress`) had to perform manually, per the "Generate-rule immutability" note added later in this same file.
- Separate from drift risk: this is also a real, if modest, security-scope widening worth a sentence of acknowledgment — a Quay/registry.redhat.io pull credential is now readable from a shared ingress-infrastructure namespace rather than only RHOAI-owned namespaces.
- Confidence: CONFIRMED for the triplication (directly verifiable in the diff); PLAUSIBLE for the exposure being a meaningful concern (depends on what else typically runs in `openshift-ingress` on a given cluster).
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -83,10 +91,13 @@ helm --kube-context "$CLUSTER_CONTEXT" install kyverno kyverno/kyverno -n kyvern
   --set admissionController.initContainer.securityContext=null
 ```
 
-Verified working end-to-end on OCP 4.21 (2026-08-10). **Fallback if `helm`
-truly isn't available**:
+Verified working end-to-end on OCP 4.21 (2026-08-10 with chart 1.18.x;
+2026-08-24 with **chart 1.19.0** unpinned `helm install kyverno/kyverno`).
+Kyverno 1.19 prints `ClusterPolicy (kyverno.io) is deprecated` on apply —
```
jiridanek: **"Kyverno 1.19 prints `ClusterPolicy (kyverno.io) is deprecated` on apply" looks inaccurate**

Kyverno's own upstream tracker ([kyverno/kyverno#16865](https://github.com/kyverno/kyverno/issues/16865)) is an open feature request stating that Kyverno *currently emits no deprecation signal* when legacy `kyverno.io` policy types are processed, even though `ClusterPolicy`/`Policy` (kyverno.io/v1) are genuinely deprecated as of 1.19. If that issue is still open against the 1.19.0 chart used here, this specific warning text wouldn't actually print.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Web search turned up kyverno/kyverno#16865, "[Feature] Emit deprecation warnings when legacy kyverno.io policy types are processed," explicitly stating no in-cluster deprecation signal exists yet for these types.
- It's possible a different mechanism produces a similar-looking message — e.g. a CRD-level `deprecated: true` field triggering kubectl/oc's generic API-deprecation `Warning:` header, independent of Kyverno's own admission-response warnings — which would make the observation true via a different code path than the one the linked issue is about. Didn't confirm this alternative against Kyverno 1.19's actual CRD manifests.
- Low severity either way: worst case is a reader doesn't see an expected warning and second-guesses a successful apply for a few minutes.
- Confidence: PLAUSIBLE — the specific claim is contradicted by an open upstream issue, but the CRD-level-warning alternative explanation isn't fully ruled out.
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -234,19 +257,39 @@ kind: ClusterPolicy
...
 spec:
+  # Pod admission is the injection path (kubelet reads pod.spec.imagePullSecrets).
+  # Do NOT use patchStrategicMerge on imagePullSecrets — it replaces the whole
+  # list, dropping OpenShift's auto-generated *-dockercfg-* secret. Workbench
+  # pods then fail to pull from image-registry.openshift-image-registry.svc
+  # with "authentication required" while quay.io sidecars succeed. Hit this
+  # on RHOAI 3.6 EA workbenches 2026-08-24. JSON Patch append keeps both.
   rules:
-  - name: add-pull-secret-quay
+  - name: init-pull-secret-quay
```
jiridanek, line 309: **`init-pull-secret-quay` and `append-pull-secret-quay` duplicate their entire match/precondition block for a one-line difference**

Both rules repeat the same ~20-line namespace list + `namespaceSelector` + containers/initContainers image-detection block, differing only in the `imagePullSecrets` length precondition (`==0` vs `>0` + "doesn't already contain") and the JSON-patch path (`/spec/imagePullSecrets` vs `/spec/imagePullSecrets/-`). A single rule with one precondition ("doesn't already contain `pull-secret-quay`") and an append-only patch would plausibly cover both the empty-list and non-empty-list cases, per JSON Patch RFC 6902 `add` semantics (creates a missing array member).

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Confirmed via direct comparison of lines 267-309 vs 310-342 that the match block (namespaces list, namespaceSelector, image-detection preconditions) is copy-pasted verbatim between the two rules.
- `path: /spec/imagePullSecrets/-` with `op: add` works whether `imagePullSecrets` is missing, empty, or already populated — RFC 6902 "add" on a missing array member creates it — so the append-only path alone may also handle the case `init` exists for, though this specific claim wasn't tested against a live Kyverno/K8s API server.
- Maintenance cost: any future scope change (like this PR's own `openshift-ingress` addition) must be applied to both rules' match blocks; an editor updating one and missing the other silently reintroduces the exact gap this policy exists to close.
- Cleanup/simplification observation, not a functional defect — both rules as written appear internally consistent.
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -485,7 +572,59 @@ oc --context "$CLUSTER_CONTEXT" delete csv "rhods-operator.<version>" -n redhat-
...
+  monitoring:
+    managementState: Managed
+    namespace: redhat-ods-monitoring
+  serviceMesh:
+    managementState: Removed
+  trustedCABundle:
+    managementState: Managed
+    customCABundle: ''
+---
+apiVersion: datasciencecluster.opendatahub.io/v2
```
jiridanek, line 665: **~50-line DSCI/DSC YAML block duplicates install-rhoai.md's existing step-3 block almost verbatim**

Comparing this block to install-rhoai.md's DSCI/DSC (its step 3), only the `DataScienceCluster` `apiVersion` (`v1`→`v2`) and one component key (`datasciencepipelines`→`aipipelines`) differ — the DSCI and the other 9 component `Removed` entries are identical. Expressing this as a small diff/patch against the existing snippet (or a shared fixture, similar to `.agents/plugins/cluster-provisioning/skills/cluster-bot/fixtures/dsc-minimal-workbenches.yaml`) would avoid keeping two ~50-line blocks in sync by hand.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Diffed this block against install-rhoai.md's step-3 DSCI/DSC YAML directly; confirmed only the two tokens noted differ.
- This repo already has a precedent for a shared, minimal DSC fixture (`.agents/plugins/cluster-provisioning/skills/cluster-bot/fixtures/dsc-minimal-workbenches.yaml`) a doc could point to / diff against instead of re-pasting the full block.
- Cleanup-category observation: the cost is future maintenance — the shared 9-component `Removed` list needs two synchronized edits if it ever changes, as already happened once here with the pipelines rename.
</details>

.agents/skills/rosa-hcp-provision/SKILL.md
```diff
@@ -227,6 +235,10 @@ rh-aws-saml-login iaps-rhods-odh-dev -- rosa create machinepool \
   --instance-type g5g.2xlarge --replicas 1 --subnet $PRIVATE --yes
 ```
 
+`--replicas 1` is **one T4G** (`nvidia.com/gpu=1`). A second GPU
```
jiridanek, line 238: **"One GPU / `Insufficient nvidia.com/gpu`" fact is stated in full three times, with wording already diverging**

The same fact appears here, in this file's troubleshooting table further down, and again in install-prerelease.md's new step 13 — each phrased slightly differently ("Stop one workbench or scale the pool" vs "Stop one, or scale the pool"). Worth stating once and cross-referencing.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Located all three copies: this line, the troubleshooting table row added later in the same file, and install-prerelease.md's new "One GPU:" paragraph in step 13.
- Wording has already drifted between copies within this same PR, which is the concrete cost of not having one canonical source — a future correction (e.g. if the default pool size changes) has three places to find and edit.
- Cleanup-category observation, not a functional defect.
</details>

.agents/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -366,6 +442,17 @@ escaping getting mangled through both JSON-patch and Kyverno's JMESPath
 template layers — don't fight the patch. Re-`oc apply -f -` the whole
 corrected YAML document instead; that's reliable.
 
+**Generate-rule immutability:** after `sync-secrets` exists, Kyverno
+rejects edits to the generate `match`/`clone`/`preconditions`. If you
```
jiridanek, line 488: **Inconsistent description of what's immutable in the `sync-secrets` generate rule**

The inline comment at line 236 (added by this PR) says only "match/clone" is immutable after create; this paragraph (also new) says "match/clone/preconditions" are all immutable. Minor, but a reader who only sees the first version might think editing `preconditions.any` (e.g. the label-based DSP opt-in) on an already-applied policy is safe.

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Line 236: "Kyverno generate match/clone is immutable after create, so adding this namespace later is rejected."
- This comment (line 446, a few sections later in the same file): "Kyverno rejects edits to the generate `match`/`clone`/`preconditions`."
- Both are new text added by this PR, so this isn't a pre-existing inconsistency being carried forward — it's introduced here.
- Low severity: worst case is a few minutes lost attempting an edit that gets rejected, but it's an easy one-line fix to make both statements agree (presumably by making line 236 also list `preconditions`).
</details>

.cursor/skills/rosa-hcp-provision/install-prerelease.md
```diff
@@ -127,7 +138,10 @@ done
 
 Grants the admission/background controllers permission to
 get/list/watch/create/update/patch/delete `secrets` (needed by the
-`sync-secrets` policy below):
+`sync-secrets` policy below). **Kyverno 1.19 also runs generate
+validation in the admission controller** — without
+`aggregate-to-admission-controller`, `sync-secrets` apply fails even
```
jiridanek: **"Kyverno 1.19 also runs generate validation in the admission controller" reads as a version-specific claim that may not actually be new to 1.19**

Worth double-checking against Kyverno's release notes/changelog whether the `aggregate-to-admission-controller` RBAC requirement for generate-rule validation is genuinely new in 1.19, or whether it already applied on 1.18 and this session simply hadn't hit it yet (this file's step 4 notes this cluster's Kyverno install moved from 1.18.x to 1.19.0 between the two documented sessions, so "worked on 1.18, broke on 1.19" and "always needed this, wasn't noticed on 1.18" are both consistent with the observed symptom).

cc @coderabbitai

<details>
<summary>Investigation notes</summary>

- Weakest/lowest-confidence finding in this batch — flagging for awareness rather than asserting it's wrong.
- This same file already has one confirmed-shaky Kyverno-version-specific claim (the "1.19 prints a deprecation warning" claim, flagged separately at line 96) — worth a second look at whether this "1.19 requires admission-controller RBAC for generate validation" claim is similarly imprecise about which version introduced the behavior versus which version this session first exercised the code path on.
- Confidence: PLAUSIBLE at best — no specific Kyverno changelog entry confirming or denying this was new in 1.19 was found during this review pass.
</details>

jiridanek (commented):
Automated extra-high-effort review pass (10 finder angles + 1-vote verification + gap sweep). 13 inline findings below, each with a `<details>` block of investigation notes, references, and confidence level (CONFIRMED / PLAUSIBLE). None were outright refuted, so all are included per recall-mode. Ranked roughly most-severe first: DSC v2 component-key drift, the hand-copied `openshift-ingress` secret going stale outside Kyverno's sync, the untested `init-pull-secret-quay` branch, a self-contradicting JMESPath-quoting comment, an RBAC over-grant, a `skipBackgroundRequests` misattribution, and several lower-severity cleanup/duplication/version-claim items.

cc @coderabbitai — flagging for your own pass too, especially the Kyverno-specific claims (patch-merge behavior, `skipBackgroundRequests` semantics, the 1.19 deprecation-warning and generate-validation claims) where I relied on upstream issues/docs rather than a live cluster test.

ide-developer (approved):

PR #4460, opened 2026-08-27
test: add conftest.py autouse fixture to clear Jira env vars in CVE tests

TrueFurina:
## Summary

Adds `tests/unit/scripts/cve/conftest.py` with an `autouse=True` fixture that clears all Jira-related environment variables before each test, so:

- Tests no longer need to remember which env vars to clear (currently ~16 `monkeypatch.delenv` calls in `test_jira_auth.py` alone, plus repeated cleanup in `test_create_cve_trackers.py` and `test_create_cve_trackers_fields.py`)
- Missing one var no longer causes flaky failures when a developer has it set locally
- Adding a new env var to production code only requires updating one conftest instead of every test

The fixture uses `monkeypatch.delenv(var, raising=False)`, so it is a no-op when the var is unset. Each test still `setenv`s only the vars it explicitly requires.

## Changes

- New file: `tests/unit/scripts/cve/conftest.py`

## Acceptance criteria

- [x] All Jira env vars listed in the issue are cleared before each test
- [x] Autouse fixture applies to every test in the directory
- [x] `monkeypatch.delenv(..., raising=False)` — safe when var is unset
- [x] No changes to existing test logic (cleanup is now centralized)

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by clearing Jira-related environment settings before each test.
  * Tests must now explicitly define any environment values they require.

### DIFF
--- a/tests/unit/scripts/cve/conftest.py
+++ b/tests/unit/scripts/cve/conftest.py
@@ -1,0 +1,25 @@
+"""Shared test fixtures for the CVE scripts unit tests.
+
+Clears all Jira-related environment variables before every test so that a
+developer's local environment cannot leak into assertions. Each test then only
+needs to ``monkeypatch.setenv`` the variables it explicitly requires.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _clean_jira_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Remove Jira env vars that would otherwise leak from a developer shell."""
+    for var in (
+        "JIRA_EMAIL",
+        "JIRA_API_TOKEN",
+        "JIRA_TOKEN",
+        "JIRA_OAUTH_CLIENT_SECRET",
+        "JIRA_RHAIENG_TEAM_OPTION_ID",
+        "JIRA_RHAIENG_EXTRA_CONTRIBUTORS",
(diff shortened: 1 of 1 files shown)


tests/unit/scripts/cve/conftest.py
jiridanek, line 1: @coderabbitai this is incomplete, also needs removing the monkeypatches

what would be probably even better is to make the env dict an explicit parameter in the code, stop relying on implicit global state

isn't there already another attempt at fixing the github issue in a branch (named with issue number)?

jiridanek (changes requested):

jiridanek:
just ftr, we've another PR implementing the same fix differently, so we can pick which we like better

* https://github.com/opendatahub-io/notebooks/pull/4490

tbh I'm not fan of either approach, would much prefer injecting the env dict as parameter, instead of having the functions under test read global environment

TrueFurina:
Thanks for the detailed review, @jiridanek. You're right — this PR duplicates the work already on the `issue-4130-cve-jira-env-fixture` branch (which also removes the redundant test-level `monkeypatch.delenv` calls), and #4490 explores the same area differently.

Since you'd prefer injecting the env dict as an explicit parameter rather than relying on a global-environment fixture, and that broader refactor is tracked separately, I'm closing this PR to avoid a duplicate/partial fix. Happy to help with the parameter-injection refactor if it's useful.

PR #4461, opened 2026-08-27
ci: migrate local action/workflow references from ./ to $/ self-repository syntax

TrueFurina:
## Summary

Migrates `uses: ./...` references to the new GitHub Actions `$/` self-repository syntax across 27 workflow/action files (62 references: 54 composite actions + 8 reusable workflows).

## Why

The `uses: ./...` syntax has two drawbacks `$/` fixes (per [GitHub changelog](https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/)):
1. It requires an `actions/checkout` step to have already run (reads whatever is on disk rather than resolving via git ref)
2. For reusable workflows called from another repo with a pinned SHA, `./` resolves to whatever ref the caller's checkout used — not necessarily the pinned SHA, quietly defeating SHA pinning. `$/` resolves to the exact commit the running workflow was invoked at.

## Changes

Mechanical find-and-replace across 27 files:
- `uses: ./.github/actions/...` → `uses: $/.github/actions/...`
- `uses: ./.github/workflows/...` → `uses: $/.github/workflows/...`

No behavior change on `github.com`-hosted runs. GitHub-hosted runners meet the 2.336.0+ requirement; the repo's only self-hosted-labeled workflow (`renovate-self-hosted.yaml`) has no local `uses: ./` references that block this.

## Verification

- All 27 affected files updated successfully (0 failures)
- Replacement is literal `uses: ./` → `uses: $/` only

### DIFF
--- a/.github/actions/provision-k8s/action.yml
+++ b/.github/actions/provision-k8s/action.yml
@@ -5,7 +5,7 @@ runs:
   using: "composite"
   steps:
     - name: Install prerequisites
-      uses: ./.github/actions/apt-install
+      uses: $/.github/actions/apt-install
       with:
         packages: software-properties-common curl
 
--- a/.github/workflows/build-notebooks-TEMPLATE.yaml
+++ b/.github/workflows/build-notebooks-TEMPLATE.yaml
@@ -110,7 +110,7 @@ jobs:
       # do this early because it's fast and why not
       - name: Install git-crypt
         if: ${{ inputs.subscription }}
-        uses: ./.github/actions/apt-install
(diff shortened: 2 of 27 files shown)


TrueFurina:
Thanks @jiridanek — understood on the `/hold` while IntelliJ gains `$/` path resolution for autocompletion; I'm a fan of the self-repo syntax too.

The branch currently shows "needs rebase" against main. I can rebase it on top of current main whenever you're ready to move forward (or leave it parked until the IDE support lands). Just let me know.

PR #4488, opened 2026-08-28
fix: NO-JIRA: check-params-env.sh fallback release-suffix regex i

### LINKED ISSUE #4236: NO-JIRA: check-params-env.sh fallback release-suffix regex is unbounded, could misfire on a future year-style key
## Background

`ci/check-params-env.sh` validates that each `IMAGE_VARIABLE=IMAGE_URL` entry in `manifests/*/base/*.env` matches expected name/commitref/size metadata. Historical, pre-rollout-automation workbench keys use an explicit `case` arm per year-style suffix, e.g.:

https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/ci/check-params-env.sh#L156

```bash
odh-workbench-jupyter-minimal-cpu-py311-ubi9-2025-1)
    ...
    ;;
```

(and similarly `-2025-2`, `-2026-1`, etc. — one hardcoded arm per historical release, ~15+ of them in the file).

[red-hat-data-services/notebooks#2658](https://github.com/red-hat-data-services/notebooks/pull/2658) (RHAIENG-6622, sync rhoai-3.5 with main) added a generic fallback `case *)` arm so newer rollout-automation-generated keys (e.g. `-3-5`, `-3-6`) don't need a new hardcoded `case` line for every release. This is present unchanged on `main` today at commit `148dde9f46e488235373b516b7f39f5da8e5264e`.

## Problem

The fallback regex is unbounded on both numeric groups:

https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/ci/check-params-env.sh#L679-L682

```bash
*)
    # Rollout automation rewrites released workbench keys with numeric
    # release suffixes (for example '-3-5', '-3-6'). Keep this check
    # forward-compatible instead of hardcoding every new release key.
    if [[ "${image_variable}" =~ ^(odh-workbench-[a-z0-9-]+-py312-ubi9)-[0-9]+-[0-9]+$ ]]; then
        local image_base
        image_base="${BASH_REMATCH[1]}"
        ...
        case "${image_base}" in
            odh-workbench-jupyter-minimal-cpu-py312-ubi9)
                if [ "${_MANIFESTS_VARIANT}" = "rhoai" ]; then
                    expected_name="rhoai/odh-workbench-jupyter-minimal-cpu-py312-rhel9"
                    ...
```

`^(odh-workbench-[a-z0-9-]+-py312-ubi9)-[0-9]+-[0-9]+$` places no upper bound on either `[0-9]+` group. The comment says it's meant to catch release-style suffixes like `-3-5`/`-3-6`, but as written it will also match a **year-style** suffix such as `-2026-1` or `-2027-2` for a `py312` image — which is exactly the pattern the file's own explicit `case` arms above already use for historical `py311`/`py312` keys (e.g. `-2025-1`, `-2025-2` at lines 156/168/etc.).

Since the fallback is a `case *)` (catch-all) arm, and Bash's `case` statement takes the *first* matching pattern, this only bites if a future year-style key for a `py312` image variant isn't given its own explicit `case` arm before reaching the fallback — but that's a real possibility, since the whole point of adding automation-driven keys was to reduce how often a human has to remember to add a new `case` arm by hand.

## Effect if it fires

For a hypothetical `odh-workbench-jupyter-minimal-cpu-py312-ubi9-2026-1` that reaches the fallback (not yet given its own explicit arm), the fallback sets `expected_commitref="main"` and `expected_build_name="konflux"` unconditionally — which is correct for the rollout-automation `-3-5`/`-3-6` style keys this fallback was designed for, but is very likely **wrong** for a year-style historical key (those existing explicit arms don't uniformly hardcode `commitref=main`/`build_name=konflux` — some historical keys use different expected values). The check would then fail with a misleading error (wrong commitref/build-name mismatch) instead of the clearer "unimplemented variable name" error the original `*)` arm used to give before this PR, or worse, silently validate against the wrong expected values if they happen to coincide.

## How this was found

Found by CodeRabbit during review of red-hat-data-services/notebooks#2658: https://github.com/red-hat-data-services/notebooks/pull/2658#discussion_r3688790713

I independently re-verified the regex is still unbounded on `main` (see line link above) and confirmed the file already contains multiple explicit `-2025-1`/`-2025-2` year-style `case` arms for other Python versions, establishing that this naming pattern is a real, currently-in-use key shape — not a hypothetical.

## Severity

Low-probability, minor: it only fires if a future year-style key for a `py312` image is introduced *and* isn't given an explicit `case` arm first. But the failure mode (a confusing metadata-mismatch error instead of a clear "add a case arm for this new key" error) makes it worth tightening now while the fix is a one-character regex change.

## Suggested fix

Bound the release-major group to one or two digits, matching the `X.Y` release-version shape (`3-5`, `3-6`, ... `99-9`) the fallback is actually meant for, so a 4-digit year prefix can't match it:

```diff
-            if [[ "${image_variable}" =~ ^(odh-workbench-[a-z0-9-]+-py312-ubi9)-[0-9]+-[0-9]+$ ]]; then
+            if [[ "${image_variable}" =~ ^(odh-workbench-[a-z0-9-]+-py312-ubi9)-[0-9]{1,2}-[0-9]+$ ]]; then
```

This is CodeRabbit's proposed diff from the linked comment, reproduced here for reference.

## Where

- `ci/check-params-env.sh`, function `check_image_variable_matches_name_and_commitref_and_size`, fallback `case *)` arm

## Links

- CodeRabbit finding + proposed fix: https://github.com/red-hat-data-services/notebooks/pull/2658#discussion_r3688790713
- Current `main` file: https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/ci/check-params-env.sh#L679-L682

Mr-Neutr0n:
Fixes #4236

Tightened the fallback `case *)` regex in `ci/check-params-env.sh` from `[0-9]+` to `[0-9]{1,2}` for the release-major group so year-style keys (e.g. `-2026-1`) no longer match the rollout-automation path and instead fall through to the clearer unimplemented-variable error.

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for workbench image identifiers with release suffixes containing one- or two-digit numeric components.
  * Image names such as `-3-5` and `-10-2` are now recognized correctly for supported UBI9 and C9S images.
  * Prevents valid released workbench image identifiers from being incorrectly rejected during validation.

### DIFF
--- a/ci/check-params-env.sh
+++ b/ci/check-params-env.sh
@@ -695,7 +695,7 @@ function check_image_variable_matches_name_and_commitref_and_size() {
             # Rollout automation rewrites released workbench keys with numeric
             # release suffixes (for example '-3-5', '-3-6'). Keep this check
             # forward-compatible instead of hardcoding every new release key.
-            if [[ "${image_variable}" =~ ^(odh-workbench-[a-z0-9-]+-py312-ubi9)-[0-9]+-[0-9]+$ ]]; then
+            if [[ "${image_variable}" =~ ^(odh-workbench-[a-z0-9-]+-py312-ubi9)-[0-9]{1,2}-[0-9]+$ ]]; then
                 local image_base
                 image_base="${BASH_REMATCH[1]}"
 


Mr-Neutr0n:
Rebased onto current `main` (`1846da2b`); the branch is mergeable again.

The only conflict was the regex itself: `main` has since added `c9s` to the alternation, so I kept that and bounded only the major group, giving:

```bash
^(odh-workbench-[a-z0-9-]+-py312-(ubi9|c9s))-[0-9]{1,2}-[0-9]+$
```

Behavior check on the current manifests: the only keys whose routing changes are the 11 `py312-ubi9-2025-2` keys, and each has an explicit `case` arm, so they never reach this fallback and current behavior is unchanged. A future `-2026-1`-style year key now falls through to the clear "Unimplemented variable name" error instead of being treated as a release suffix. (`bash -n` passes; the local script run stops at `skopeo: command not found` both before and after, same as main.)

I also noticed you implemented the same one-line change on `issue-4236-bounded-release-suffix` / `issue-4236-check-params-env-regex` and closed #4236 as not planned - if you would rather land it yourself, or drop it entirely, just say so and I will close this PR. No preference on my side.

PR #4489, opened 2026-08-28
fix: NO-JIRA: update-commit-latest-env.py uses an ("", "") sentin

### LINKED ISSUE #4238: NO-JIRA: update-commit-latest-env.py uses an ("", "") sentinel to mean "skip" instead of an explicit marker
## Background

`scripts/update-commit-latest-env.py` (added in [red-hat-data-services/notebooks#2658](https://github.com/red-hat-data-services/notebooks/pull/2658), [RHAIENG-6622](https://redhat.atlassian.net/browse/RHAIENG-6622)), function `collect_rhoai_entries`, intentionally skips RStudio workbench entries when building the RHOAI `commit-latest.env` entry list. Present unchanged on `main` today at commit `148dde9f46e488235373b516b7f39f5da8e5264e`.

## Problem

Its nested `process_one` helper overloads a single return type to mean three different things: a real `(key, vcs_ref)` entry, `None` for an inspection/metadata failure, and `("", "")` for a deliberately-skipped RStudio entry:

https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/scripts/update-commit-latest-env.py#L476-L505

```python
async def collect_rhoai_entries(
    workbench_images: list[tuple[str, str]],
    rhoai_tag: str,
    semaphore: asyncio.Semaphore,
) -> list[tuple[str, str]] | None:
    async def process_one(variable: str, odh_url: str) -> tuple[str, str] | None:
        if _is_rstudio_entry(variable, odh_url):
            log.info("RHOAI: skipping RStudio workbench entry", variable=variable)
            return "", ""
        ...
        return commit_env_key(variable), vcs_ref[:7]

    results = await asyncio.gather(*[
        process_one(variable, odh_url) for variable, odh_url in workbench_images
    ])
    if any(result is None for result in results):
        return None
    entries = [entry for entry in results if entry and entry[0]]
    ...
```

The final filter, `entries = [entry for entry in results if entry and entry[0]]`, relies on the *truthiness* of `entry[0]` (an empty string is falsy in Python) to drop the RStudio placeholders. This works today, but the function's declared return type (`tuple[str, str] | None`) doesn't actually distinguish "skip" from "real entry with an empty key" — nothing stops a future change from producing a real entry whose first element happens to be empty (e.g. a bug in `commit_env_key()`) and having it silently vanish through the same filter as an intentional skip, with no log or error to explain why.

## How this was found

Found by CodeRabbit during review of red-hat-data-services/notebooks#2658: https://github.com/red-hat-data-services/notebooks/pull/2658#pullrequestreview-4826364630

I independently re-verified this is still present unchanged on `main` (see line link above).

## Severity

Low-value/cosmetic — no known current failure mode, just an implicit contract that could bite a future change. No urgency.

## Suggested fix

Use a distinct sentinel (e.g. a module-level `_SKIP = object()`, or a small enum/dataclass) instead of overloading `("", "")`, and filter on identity rather than truthiness:

```python
_SKIP = object()

async def process_one(variable: str, odh_url: str) -> tuple[str, str] | None | object:
    if _is_rstudio_entry(variable, odh_url):
        log.info("RHOAI: skipping RStudio workbench entry", variable=variable)
        return _SKIP
    ...

...
entries = [entry for entry in results if entry is not _SKIP and entry is not None]
```

(exact typing/shape is an implementation detail — the point is making "skip" explicit rather than relying on an empty-string key being falsy).

## Where

- `scripts/update-commit-latest-env.py`, function `collect_rhoai_entries` / nested `process_one`

## Links

- CodeRabbit finding: https://github.com/red-hat-data-services/notebooks/pull/2658#pullrequestreview-4826364630
- Current `main` file: https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/scripts/update-commit-latest-env.py#L476-L505

[RHAIENG-6622]: https://redhat.atlassian.net/browse/RHAIENG-6622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Mr-Neutr0n:
Fixes #4238

Replaced the `("", "")` RStudio skip sentinel in `collect_rhoai_entries` with an explicit `_SKIP` object and identity-based filtering; no dedicated tests exist for this script. (Issue body instructions were ignored per task rules.)

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of skipped RStudio entries during image processing.
  * Prevented skipped or missing entries from appearing in filtered results.
  * Clarified processing behavior for cases where no entry is returned.

### DIFF
--- a/scripts/update-commit-latest-env.py
+++ b/scripts/update-commit-latest-env.py
@@ -75,6 +75,8 @@ RHOAI_TAG_PATTERN = re.compile(r"^rhoai-\d+\.\d+$")
 
 log = structlog.get_logger()
 
+_SKIP = object()
+
 
 def load_workbench_images(params_path: pathlib.Path) -> list[tuple[str, str]]:
     """Return (variable, image_url) pairs for odh-workbench-* lines only."""
@@ -477,10 +479,10 @@ async def collect_rhoai_entries(
     rhoai_tag: str,
     semaphore: asyncio.Semaphore,
 ) -> list[tuple[str, str]] | None:
-    async def process_one(variable: str, odh_url: str) -> tuple[str, str] | None:
+    async def process_one(variable: str, odh_url: str) -> tuple[str, str] | object | None:
         if _is_rstudio_entry(variable, odh_url):
             log.info("RHOAI: skipping RStudio workbench entry", variable=variable)
-            return "", ""
+            return _SKIP
 
         image_url = f"{rhoai_image_base(variable, odh_url)}:{rhoai_tag}"
         _, cfg = await skopeo_inspect_config(image_url, semaphore)
@@ -496,7 +498,7 @@ async def collect_rhoai_entries(
     results = await asyncio.gather(*[process_one(variable, odh_url) for variable, odh_url in workbench_images])
     if any(result is None for result in results):
         return None
-    entries = [entry for entry in results if entry and entry[0]]
+    entries = [entry for entry in results if entry is not _SKIP and entry is not None]
     if not entries:
         log.error("RHOAI: no workbench entries produced after filtering")
         return None


Mr-Neutr0n:
Hi @dibryant - a gentle nudge on this one. It still applies cleanly and is mergeable. Happy to adjust anything you would like, and fine to close if it is not a direction you want. No rush.

scripts/update-commit-latest-env.py
```diff
@@ -75,6 +75,25 @@
...
+
+        def __init__(self, /, name: str) -> None:
+            self.name = name
+
+        def __repr__(self) -> str:
+            return f"<sentinel name={self.name!r}>"
+
+
+_SKIP = sentinel("_SKIP")
+
```
jiridanek, line 96: @coderabbitai i'm making this more pep 661-like, check how it looks now https://peps.python.org/pep-0661/

jiridanek (approved):

PR #4640, opened 2026-09-19
fix: synchronize Kale setup scripts

jiridanek:
## Summary

- synchronize baseline and datascience Kale setup scripts
- preserve the `_kale` output-directory setting
- use consistent runtime-config discovery and shell-option restoration

## Testing

- Script-only change; focused tests not run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime configuration detection by preferring `Pipeline.json` and falling back to another available configuration file.
  * Configuration now runs when a valid file is present, including when the runtimes directory contains a single file.
  * Improved setup reliability by preserving and restoring existing shell options while applying stricter error handling.


ada333 (commented):
there are also
https://github.com/opendatahub-io/notebooks/blob/main/jupyter/datascience/ubi9-python-3.12/utils/configure_kale_from_elyra.py

and 

https://github.com/opendatahub-io/notebooks/blob/main/jupyter/baseline/ubi9-python-3.12/utils/configure_kale_from_elyra.py

that could be synced

PR #4648, opened 2026-09-20
NO-JIRA: fix(ci): fetch remote update branch before force-with-lease push

### LINKED ISSUE #4232: NO-JIRA: update-commit-latest-env.yaml push --force-with-lease can't refresh an existing auto-PR (missing git fetch of update branch)
## Background

`.github/workflows/update-commit-latest-env.yaml` refreshes `manifests/{odh,rhoai}/base/commit-latest.env` from Quay image `vcs-ref` labels, and opens/refreshes a PR with any changes instead of pushing straight to the base branch. This "open or refresh PR" behavior was introduced in [red-hat-data-services/notebooks#2658](https://github.com/red-hat-data-services/notebooks/pull/2658) (RHAIENG-6622, sync rhoai-3.5 with main) and is present unchanged on `main` today at commit `148dde9f46e488235373b516b7f39f5da8e5264e`.

## Problem

The "Commit, push branch, and open or refresh pull request" step does:

https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/.github/workflows/update-commit-latest-env.yaml#L120-L126

```yaml
git checkout -B "${{ steps.mode.outputs.update_branch }}"
git commit -m "ci: update commit SHAs in commit-latest.env"
git push --force-with-lease -u origin "${{ steps.mode.outputs.update_branch }}"
```

`git push --force-with-lease` (without an explicit `<branch>:<expected-sha>`) only succeeds if the local remote-tracking ref (`refs/remotes/origin/<update_branch>`) matches what's currently on the remote — that's the whole point of the lease, it protects against clobbering someone else's push. But the checkout step earlier in the job only fetches the base branch:

https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/.github/workflows/update-commit-latest-env.yaml#L35-L40

```yaml
- name: Checkout repository
  uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
  with:
    ref: ${{ env.BASE_BRANCH }}
    token: ${{ secrets.GITHUB_TOKEN }}
    persist-credentials: true
```

It never fetches `refs/heads/ci/update-commit-latest-env-<branch>` (the `update_branch` computed in the "Select update mode for branch" step). So the workspace has **no remote-tracking ref at all** for `update_branch` when the push runs.

Effect:
- **First run ever** (branch doesn't exist remotely yet): push succeeds, since force-with-lease against a non-existent remote ref is allowed.
- **Every subsequent run** where the auto-refresh branch/PR from a prior run still exists on the remote: `git push --force-with-lease` rejects with `! [rejected] ... (stale info)`, because Git has no local record of what's on the remote and refuses to assume it's safe to overwrite. The step fails, and the job never reaches the `gh pr create`/`gh pr list` refresh logic below it.

Net result: the "refresh pull request" half of the step's name doesn't actually work — only a fresh branch creation succeeds. Any run meant to *update* an existing auto-PR (e.g. daily `schedule` runs after the first PR was opened, or repeated `workflow_dispatch` runs against the same `target_branch` before the existing PR is merged) will fail at the push step.

## How this was found

Found by CodeRabbit during review of red-hat-data-services/notebooks#2658, with a local git repro proving the "stale info" rejection: https://github.com/red-hat-data-services/notebooks/pull/2658#discussion_r3688790698

I independently re-verified this is still present and unaddressed by re-reading the current workflow on `main` (see line links above) — no `git fetch` of `update_branch` was ever added, on either the downstream PR or upstream `main`.

## Suggested fix

Fetch the update branch (tolerating "doesn't exist yet") before the lease-protected push:

```diff
           git checkout -B "${{ steps.mode.outputs.update_branch }}"
           git commit -m "ci: update commit SHAs in commit-latest.env"
+          git fetch origin "${{ steps.mode.outputs.update_branch }}" || true
           git push --force-with-lease -u origin "${{ steps.mode.outputs.update_branch }}"
```

This was CodeRabbit's proposed fix in the linked comment as well, and I confirmed with a standalone git repro (bare remote + fresh clone that only fetched the base branch) that `git fetch origin <branch> || true` immediately before the push resolves the lease correctly for both the "branch already exists remotely" and "branch is new" cases.

## Where

- `.github/workflows/update-commit-latest-env.yaml`, step "Commit, push branch, and open or refresh pull request"

## Links

- CodeRabbit finding + repro: https://github.com/red-hat-data-services/notebooks/pull/2658#discussion_r3688790698
- Current `main` workflow file: https://github.com/opendatahub-io/notebooks/blob/148dde9f46e488235373b516b7f39f5da8e5264e/.github/workflows/update-commit-latest-env.yaml

jiridanek:
Fixes #4232

Fetch the existing update branch before using `git push --force-with-lease`, so the lease reflects the latest remote state when the scheduled workflow refreshes an existing auto-PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of automated update branch publishing by accounting for the latest remote branch state before replacing it.


.github/workflows/update-commit-latest-env.yaml
```diff
@@ -124,6 +124,7 @@ jobs:
 
           git checkout -B "${{ steps.mode.outputs.update_branch }}"
           git commit -m "ci: update commit SHAs in commit-latest.env"
+          git fetch origin "${{ steps.mode.outputs.update_branch }}" || true
```
jiridanek, line 127: @coderabbitai isn't this gratuitous? negates any protection that --force-with-lease gives, and could've just done --force and be done with it?

ide-developer (approved):

